Issue #244, opened 2012-06-26
SdkVersionHelper.getSdkInt() retruns wrong sdk int?


pleerock:
06-26 12:36:50.909: E/dalvikvm(567): Could not find method tj.android.projects.kalima.version1.activity.DownloadPackageListActivity_.onBackPressed, referenced from method tj.android.projects.kalima.version1.activity.DownloadPackageListActivity_.onKeyDown
06-26 12:36:50.909: W/dalvikvm(567): VFY: unable to resolve virtual method 2685: Ltj/android/projects/kalima/version1/activity/DownloadPackageListActivity_;.onBackPressed ()V
06-26 12:36:50.909: W/dalvikvm(567): VFY:  rejecting opcode 0x6e at 0x0010
06-26 12:36:50.919: W/dalvikvm(567): VFY:  rejected Ltj/android/projects/kalima/version1/activity/DownloadPackageListActivity_;.onKeyDown (ILandroid/view/KeyEvent;)Z
06-26 12:36:50.919: W/dalvikvm(567): Verifier rejected class Ltj/android/projects/kalima/version1/activity/DownloadPackageListActivity_;
06-26 12:36:50.919: W/dalvikvm(567): Class init failed in newInstance call (Ltj/android/projects/kalima/version1/activity/DownloadPackageListActivity_;)
06-26 12:36:50.919: D/AndroidRuntime(567): Shutting down VM
06-26 12:36:50.919: W/dalvikvm(567): threadid=3: thread exiting with uncaught exception (group=0x4001aa28)
06-26 12:36:50.919: E/AndroidRuntime(567): Uncaught handler: thread main exiting due to uncaught exception
06-26 12:36:50.959: E/AndroidRuntime(567): java.lang.VerifyError: tj.android.projects.kalima.version1.activity.DownloadPackageListActivity_
06-26 12:36:50.959: E/AndroidRuntime(567):  at java.lang.Class.newInstanceImpl(Native Method)
06-26 12:36:50.959: E/AndroidRuntime(567):  at java.lang.Class.newInstance(Class.java:1472)
06-26 12:36:50.959: E/AndroidRuntime(567):  at android.app.Instrumentation.newActivity(Instrumentation.java:1097)
06-26 12:36:50.959: E/AndroidRuntime(567):  at android.app.ActivityThread.performLaunchActivity(ActivityThread.java:2316)
[...]

PR #254, opened 2012-06-28
Back is now handled only if onBackPressed() exists

pyricau:
Back is now handled only if onBackPressed() exists, not taking into account the Activity method. 

See #244


stroodlepup:
This does not seem to work on 4.0.0

dodgex:
@stroodlepup this has been [removed](https://github.com/excilys/androidannotations/pull/1170) with android annotations 4.0.0

stroodlepup:
how should I handle backpressed events? I am using appcompat and I plan on implementing AA as much as I can

dodgex:
[This SO Post](http://stackoverflow.com/questions/5312334/how-to-handle-back-button-in-activity) should help.

Issue #560, opened 2013-04-16
MenuItem reference


davidjorgensen:
Is there a place to discuss things like this?

Anyways, is it possible to get a menuitem injected?

I am using the @OptionsMenu and @OptionsItem and they work fine.

More specifically I am working with the action bar with a refresh button. When the refresh button is pressed the Bluetooth will go into discover mode and search for nearby BT devices. So I want to change the refresh button to an indeterminate progress bar when i receive the intent "discovery started", and show the refresh button when the discovery finished intent is received.. In order to do this i need a reference to the menuitem to change the view.

Is it possible to do this already or would it make any sense to implement a @MenuItemById or @MenuById annotation to be used like the @ViewById annotation?

Regards,
David

Issue #549, opened 2013-04-03
Injection of MenuItems


EurigJones:
I quite often need to be able to hide ActionBar menu items programmatically. It would be nice to have the possibility of injecting these into my Activity/Fragment.

```
@MenuItem myActionBarMenuItem;
```

... for example.

PR #567, opened 2013-04-22
Add a @OptionMenuItem annotation to inject MenuItem

DayS:
Referred to #560 and #549

This PR add a `@OptionMenuItem` which allows the developer to inject a MenuItem (android or ActionBarSherlock) into an activity.

The developer should just be aware that injection is done in `onCreateOptionsMenu()` so the fields will be set only when the menu is displayed (immediately if an ActionBar is provided, otherwise when the user press the Menu button)

### DIFF
--- a/AndroidAnnotations/androidannotations/src/test/java/com/actionbarsherlock/app/SherlockActivity.java
+++ b/AndroidAnnotations/androidannotations/src/test/java/com/actionbarsherlock/app/SherlockActivity.java
@@ -1,0 +1,22 @@
+/**
+ * Copyright (C) 2010-2013 eBusiness Information, Excilys Group
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may not
+ * use this file except in compliance with the License. You may obtain a copy of
+ * the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed To in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+ * License for the specific language governing permissions and limitations under
+ * the License.
+ */
+package com.actionbarsherlock.app;
+
+import android.app.Activity;
+
+public abstract class SherlockActivity extends Activity {
+
+}
--- a/AndroidAnnotations/androidannotations/src/test/java/com/actionbarsherlock/view/MenuItem.java
+++ b/AndroidAnnotations/androidannotations/src/test/java/com/actionbarsherlock/view/MenuItem.java
@@ -1,0 +1,20 @@
+/**
+ * Copyright (C) 2010-2013 eBusiness Information, Excilys Group
+ *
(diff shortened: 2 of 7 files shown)


Joqus:
Thanks for the explanation guys, I had some trouble with it.

TomIsYourName:
So there is still no solution for this question ? Which means we can not use annotation with menu options ?

WonderCsabo:
@TomIsYourName what is your problem?

laszlourszuly:
If I'm guessing your question correctly, @TomIsYourName, I would say "no, you can't". Not the way I think you mean at least.

Your menu items aren't available in `@AfterInject` or `@AfterViews` since they are inflated by Android (and thereby also injected by AndroidAnnotations) well after those two states.

I would suggest you calculate your desired menu item state in either `@AfterInject` or `@AfterViews` and store it in a member field. When the menu then, at some point in the future, is inflated and injected in your activity, you already have your state prepared and can apply it to the menu item of your choice then.

The below code shows an example where you fetch a couple of items from the network and based on whether that result is empty or not, you set your menu item in question to either enabled or disabled:

A different, maybe more down-to-earth example that would describe the same situation could be if you want to give your grand child a T-shirt. You can't dress him in it now, because he's not born yet. You can still *buy* the T-shirt now and put it in your wardrobe, but you have to await his birth before you can give it to him.

```java
public class MyActivity extends AppCompatActivity {

    @Bean
    protected MyNetworkClient myNetworkClient;

    private MenuItem myMenuItem = null;
    private boolean shouldBeEnabled = false;

    @AfterInject
    protected void afterInject() {
        myNetworkClient.getProductNames(new SuccessListener() {
            public void onSuccess(List<String> productNames) {
                shouldBeEnabled = !productNames.isEmpty();
                updateMyMenuItem();
            }
        });
    }

    @Override
    public void onCreateOptionsMenu(Menu menu, MenuInflater inflater) {
        super.onCreateOptionsMenu(menu, inflater);
        myMenuItem = menu.findItem(R.id.action_continue);
        updateMyMenuItem();
    }

    private void updateMyMenuItem() {
        if (myMenuItem != null) {
            myMenuItem.setEnabled(shouldBeEnabled);
        }
    }

}
```

If you strictly want to, you could maybe even replace the `onCreateOptionsMenu(...)` method with a corresponding `@InjectMenu` annotated method, but this requires AndroidAnnotations 4.0.0 or later:

```java
@InjectMenu
protected void injectMenu(Menu menu) {
    myMenuItem = menu.findItem(R.id.action_continue);
    updateMyMenuItem();
}
```

dodgex:
you can use `@OptionsMenuItem` on a method too.

```java
@OptionsMenuItem(R.id.action_continue)
void singleInjection(MenuItem item) {
	// do stuff with `item`
}
```

this method will be called once your menu has been inflated.

Rafasystec:
> you can use `@OptionsMenuItem` on a method too.
> 
> ```java
> @OptionsMenuItem(R.id.action_continue)
> void singleInjection(MenuItem item) {
>     // do stuff with `item`
> }
> ```
> 
> this method will be called once your menu has been inflated.

It worked for me. Thank you.

PR #967, opened 2014-04-17
323 custom rest headers

maltzj:
It seemed like some of the stuff under the contributions page didn't apply to this PR (I don't think I should be deploying new snapshots willy nilly for example), so I may have done this wrong.

I also manually checked the code that the tests output, but I would feel better if there were automated testing in place to be sure that headers were properly set.

Lastly, I think this code will break if there is both a @RequiresHeader and a @Header annotation on the same method.  Users would have to be pretty dumb to do this, but it is definitely a thing that is conceivable.

Let me know what you all need in the way of support for this.


WonderCsabo:
@yDelouis can you refactor this PR?

yDelouis:
After 4.0.

WonderCsabo:
Refactored to #1470.

Issue #301, opened 2012-08-29
@Parcelable 


eric-taix:
Implement the Parcelable interface is a waste of time and it's always the same code : 
for each attributes to be parceled 
  get the value
  write it (according to its type)
end for

Same kind of code when reading a parcelable

And some glue !

A @Parcelable annotation would be really appreciated and will save a lot of time to all of us

yDelouis:
Implemented

PR #977, opened 2014-04-26
Integrated Parceler into Extras

johncarl81:
Basic integration of Parcler including transparent serialization/deserialization when used as Extras or InstanceState.  I was unable to test this PR fully, so please let me know if anything is out of order or if any changes need to be made.
#301

### DIFF
--- a/AndroidAnnotations/androidannotations/src/main/java/org/androidannotations/handler/ExtraHandler.java
+++ b/AndroidAnnotations/androidannotations/src/main/java/org/androidannotations/handler/ExtraHandler.java
@@ -119,6 +119,9 @@ public class ExtraHandler extends BaseAnnotationHandler<HasExtras> {
 		 * always returns null;
 		 */
 		restoreMethodCall = handleByteArrayExtraBug(element, extraKeyStaticField, restoreMethodCall);
+        if(bundleHelper.isParcelerBean()){
+            restoreMethodCall = codeModel().ref("org.parceler.Parcels").staticInvoke("unwrap").arg(restoreMethodCall);
+        }
 
 		ifContainsKey.assign(extraField, restoreMethodCall);
 	}
--- a/AndroidAnnotations/androidannotations/src/main/java/org/androidannotations/handler/FragmentArgHandler.java
+++ b/AndroidAnnotations/androidannotations/src/main/java/org/androidannotations/handler/FragmentArgHandler.java
@@ -107,6 +107,9 @@ public class FragmentArgHandler extends BaseAnnotationHandler<EFragmentHolder> {
 			}
 
 		}
+        if(bundleHelper.isParcelerBean()){
+            restoreMethodCall = codeModel().ref("org.parceler.Parcels").staticInvoke("unwrap").arg(restoreMethodCall);
+        }
 		ifContainsKey.assign(extraField, restoreMethodCall);
 	}
 
--- a/AndroidAnnotations/androidannotations/src/main/java/org/androidannotations/handler/InstanceStateHandler.java
+++ b/AndroidAnnotations/androidannotations/src/main/java/org/androidannotations/handler/InstanceStateHandler.java
@@ -15,11 +15,7 @@
  */
 package org.androidannotations.handler;
 
-import static com.sun.codemodel.JExpr.ref;
-
-import javax.annotation.processing.ProcessingEnvironment;
-import javax.lang.model.element.Element;
-
+import com.sun.codemodel.*;
 import org.androidannotations.annotations.InstanceState;
 import org.androidannotations.helper.APTCodeModelHelper;
 import org.androidannotations.helper.AnnotationHelper;
@@ -28,14 +24,10 @@ import org.androidannotations.holder.HasInstanceState;
 import org.androidannotations.model.AnnotationElements;
 import org.androidannotations.process.IsValid;
 
-import com.sun.codemodel.JBlock;
-import com.sun.codemodel.JClass;
-import com.sun.codemodel.JExpr;
-import com.sun.codemodel.JExpression;
-import com.sun.codemodel.JFieldRef;
-import com.sun.codemodel.JInvocation;
-import com.sun.codemodel.JMethod;
-import com.sun.codemodel.JVar;
+import javax.annotation.processing.ProcessingEnvironment;
+import javax.lang.model.element.Element;
+
+import static com.sun.codemodel.JExpr.ref;
 
 public class InstanceStateHandler extends BaseAnnotationHandler<HasInstanceState> {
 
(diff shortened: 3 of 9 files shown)


WonderCsabo:
@johncarl81 i think we can revisit this one. It would be nice to have our implementation of `Parcelable` generators, but for the sake of AA + `Parceler` users, we could add this integration.

I think this PR was rejected, because the it seemed that AA will always need the `Parceler` processor. However as you explained it in the last comment, it is only needed if the user actually adds it, and will not make an difference at all for non `Parceler` users.

johncarl81:
@WonderCsabo, ok, sounds good.

Since we ended this effort a few other libraries picked up Parceler support.  I'd like to highlight FragmentArgs' implementation as it offers further decoupling which may be of interested for AA.  FragmentArgs offers an interface to handle any extra serialization concerns.  This plays well into how the Parcels utility class wraps / unwraps `@Parcel` annotated beans (and ones that may not i.e. `@ParcelClass`).  Here's the relevant classes:
https://github.com/sockeqwe/fragmentargs/blob/master/annotation/src/main/java/com/hannesdorfmann/fragmentargs/bundler/ArgsBundler.java
https://github.com/sockeqwe/fragmentargs/blob/master/bundler-parceler/src/main/java/com/hannesdorfmann/fragmentargs/bundler/ParcelerArgsBundler.java
https://github.com/sockeqwe/fragmentargs/#argsbundler

I imagine we could introduce a similar technique into the `@Extra` annotation:
```java
 @Extra(value = "myParcelerExtra", converter = ParcelerConverter.class) ExampleParceler ep;
```

If you like this technique I'd be happy to work on a PR.

Also, I took a lesson from this conversation.  From this I ended up shading the dependencies the Parceler processor had on Transfuse and Guava.  This cleans up the dependency tree associated with Parceler, eliminates any library incompatibilities (remember we had to upgrade Guava), and removes the concern that Transfuse would be included along with Parceler.  Parceler is now simply 2 jars, the Parceler processor and Parceler API.

WonderCsabo:
Well, i am not a real fan of shading, as it has lots of drawbacks, and kicks the dependency management concept in the ass. But it is needed sometimes, and who am i to tell when it is, so i accept your decision. :)

I like the idea of decoupling, but i think adding the converter parameter always is inconvenient. A static method should be available to set the default one. I also i think we should still use Parceler for your annotated classes even if the client do not explicitly set the converter.

johncarl81:
In that case we could simply more forward with the previous approach that will decorate the get/set calls to the extra with `Parcels.wrap/unwrap` if `@Parcel` exists on the target extra.  Honestly, this is the most straight forward approach.

I'd just like to emphasize that the target extra may have a generated `Parcelable` but may not be annotated by `@Parcel`.  Here's the case: http://parceler.org#classes_without_java_source .
We could have a scenario like the following:
```java
@InjectExtra("parcelerExtra") Example e;

public class Example{}

@ParcelClass(Extra.class)
public class SomeOtherClass{}
```
I'd imagine we would error in this case.  Not sure if this is a concern for you, but it would be a gap in the integration.

WDTY?

WonderCsabo:
> In that case we could simply more forward with the previous approach that will decorate the get/set calls to the extra with Parcels.wrap/unwrap if @Parcel exists on the target extra. 

Hmm, i a little bit confused here. Previous is your old implementation in this PR? And decorate is the converter?

About `@ParcelClass`: i read all the comments, i do not forgot it. :) My question is: why we cannot handle that like we handle `@Parcel`?

johncarl81:
Yes, the "previous" implementation I am referring to is the old impl in this PR.  By decorate Im referring to what Im doing with `Parcels.wrap()` / `unwrap()` in this PR.

`@ParcelClass` allows you to generate a `Parcelable` wrapper without annotating the target class.  Notice in the example I gave, a Parcelable is generated for `Example`, but `Example` is not annotated with `@Parcel`, instead it is referenced via the `value` parameter of `@ParcelClass`.  This is handy for classes that you can't edit the Java source (such as in a library).

I see the problem being that it's hard to detect that a given class is a candidate for wrapping if the target class isn't annotated directly with the `@Parcel` annotation.

Here's the possible implementations I see (and I've seen in the community):

1.  Don't handle `@ParcelClass` referenced `Parcelable` wrapping.  I can't say this is the best scenario, but this is a corner case.  This is the case with this PR at the moment.
2.  Add a parameter to `@Extra` (ie: `@Extra(value="forced", forceParceler = true)`) to force writing the `Parcels.wrap/unwrap` call.  This would be a fall-back case in the event that this corner case is reached.  This is what we do on Transfuse at the moment: http://androidtransfuse.org/documentation.html#parcel (although Im considering moving to something like FragmentArgs implemented).
3.  Add a parameter to `@Extra` to handle conversions like FragmentArgs did (mentioned earlier in this tread.
4.  Perform an initial scan of the classes being compiled for the `@ParcelClass` annotation and store the the referenced class `value` in a repository.  Then, when generating `@Extra` reads / writes and the given type matches a class in the repository generate the wrapping/unwrapping code.  Honestly, I've considered doing this in Transfuse, but this approach is pretty involved for such a small feature.  Also, I'm not sure how this would fit into the AA architecture in general.

I'm sure there are other approaches.  My recommendation is to go with point 3 and secondarily 2.

WonderCsabo:
@johncarl81 yeah, i understood `@ParcelClass` after i wrote the comment. :)

I had an idea: can the same annotation be claimed by two annotation processors? If so, we could process `@ParcelClass` and add it in the a map easily, then use it when processing `@Extra` etc.

Also please note we copy all non-AA annotations to our generated subclasses, so if two annotation processors cannot claim the same annotation in a round, maybe we could claim it only the first round, and `Parceler` can claim it in the second.

But i think we an handle `@ParcelClass` later. I am okay with your approach here, can you update it onto develop and push another PR? Also please note i can review the PR soon, but i can only merge it later, because we are currently doing a very big refactor #1480, and we the develop branch is currently frozen.

WonderCsabo:
@johncarl81 any news on this one?

johncarl81:
Sorry, I haven't given this any attention since we last talked.  I'll try to get something going soon.

johncarl81:
@WonderCsabo, just reread this thread, yes, we could have AA recognize `@ParcelClass`, but maybe that would be better to address this corner case if it becomes an issue.  Also, as you mentioned, we can always add the static calls by hand.   To confirm, for now we will be looking for classes annotated with `@Parcel` and if AA finds that the extra is annotated it will wrap the get/set calls with the `Parcels.wrap`/`unwrap` calls.  Is this correct?  Should we also wrap get/set calls around the `@FragmentArg`s?

WonderCsabo:
@johncarl81 we have a bunch of annotations which extract from / add objects to `Intent`s. This is all handled in our `BundleHelper` class. I think it would be enough to just modify that class. I just checked the old changes of yours, and i think it is possible.

About `@ParcelClass`: i agree, let's focus on `@Parcel` for now.

PR #1162, opened 2014-09-24
Annotation helpers for PreferenceActivity and PreferenceFragment

WonderCsabo:
This PR is a WIP, and related to #7.

I decided to use resource IDs instead of plain string preference keys in the annotations, since this is the suggested way (if you use a string resource as a key, you can use it both  in java and xml). Also the implementation is easier, since we can rely on existing concepts.

The implementation is basically the same as View injection/listener. Currently there is some code duplication, but i am not sure we should remove that, because it would really obfuscate the intent of the code. But i gladly accept suggestions on that.

@yDelouis, @DayS, @dodgex i would appreciate your feedback.

### DIFF
--- a/AndroidAnnotations/androidannotations/src/main/java/org/androidannotations/holder/FoundHolder.java
+++ b/AndroidAnnotations/androidannotations/src/main/java/org/androidannotations/holder/FoundHolder.java
@@ -1,0 +1,60 @@
+/**
+ * Copyright (C) 2010-2015 eBusiness Information, Excilys Group
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may not
+ * use this file except in compliance with the License. You may obtain a copy of
+ * the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed To in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+ * License for the specific language governing permissions and limitations under
+ * the License.
+ */
+package org.androidannotations.holder;
+
+import static com.sun.codemodel.JExpr._null;
+import static com.sun.codemodel.JExpr.cast;
+
+import com.sun.codemodel.JBlock;
+import com.sun.codemodel.JClass;
+import com.sun.codemodel.JExpression;
+
+public abstract class FoundHolder extends GeneratedClassHolderDecorator<GeneratedClassHolder> {
+
+	private JClass type;
+	private JExpression ref;
+	private JBlock ifNotNullBlock;
+	private boolean ifNotNullCreated = false;
+
+	public FoundHolder(GeneratedClassHolder holder, JClass type, JExpression ref, JBlock block) {
+		super(holder);
+		this.type = type;
+		this.ref = ref;
+		ifNotNullBlock = block;
+	}
+
+	public JExpression getRef() {
+		return ref;
+	}
+
+	public JExpression getOrCastRef(JClass type) {
+		if (this.type.equals(type) || getBaseType().equals(type)) {
+			return ref;
+		} else {
+			return cast(type, ref);
+		}
+	}
+
+	protected abstract JClass getBaseType();
+
+	public JBlock getIfNotNullBlock() {
+		if (!ifNotNullCreated) {
+			ifNotNullBlock = ifNotNullBlock._if(ref.ne(_null()))._then();
+			ifNotNullCreated = true;
+		}
+		return ifNotNullBlock;
+	}
+}
--- a/AndroidAnnotations/androidannotations/src/main/java/org/androidannotations/holder/HasPreferenceHeaders.java
+++ b/AndroidAnnotations/androidannotations/src/main/java/org/androidannotations/holder/HasPreferenceHeaders.java
@@ -1,0 +1,26 @@
+/**
+ * Copyright (C) 2010-2015 eBusiness Information, Excilys Group
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may not
+ * use this file except in compliance with the License. You may obtain a copy of
+ * the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed To in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+ * License for the specific language governing permissions and limitations under
+ * the License.
+ */
+package org.androidannotations.holder;
+
+import com.sun.codemodel.JBlock;
+import com.sun.codemodel.JVar;
+
+public interface HasPreferenceHeaders extends HasPreferences {
+
+	JBlock getOnBuildHeadersBlock();
+
+	JVar getOnBuildHeadersTargetParam();
+}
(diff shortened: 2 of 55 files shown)


dodgex:
i need this. now. :D

WonderCsabo:
Still waiting for @yDelouis for a code review. In the meantime, you can build your own binaries from this branch and use them in your app. :)

dodgex:
nah. its bad enough that i have to use snapshot builds. i (partialy) solved my case without annotations but would have been nicer with. ;)

yDelouis:
I haven't used the `Preference` API yet. 
So I don't think I'm the right person to review the features. @dodgex, could you validate it fits your needs ?
I still can review the code style and architecture later.

WonderCsabo:
@dodgex already reviewed it by a functional point of view. So if you can review the code style etc, we can merge this i think.

AndroidAnnotations/androidannotations/src/main/java/org/androidannotations/holder/FoundHolder.java
```diff
@@ -0,0 +1,61 @@
...
+			return ref;
+		} else {
+			return cast(type, ref);
+		}
+	}
+
+	protected abstract JClass getBaseType();
+
+	public JBlock getIfNotNullBlock() {
+		if (!ifNotNullCreated) {
```
yDelouis: Why don't you check if `ifNotNullBlock` is not null ?

yDelouis:
I'm not okay with one thing : 
Why `PreferenceScreenHandler` deals with `EComponentWithViewSupportHolder` ? I think you should create an interface `HasPreference` like we did for `HasInstanceState`. It would prevent you from casting to `EFragmentHolder` in the same class.

AndroidAnnotations/androidannotations/src/main/java/org/androidannotations/holder/FoundHolder.java
WonderCsabo: I think this code is completely copied from the original `FoundViewHolder` class.

WonderCsabo:
You are right about `HasPreference`, i'll change it.

WonderCsabo:
@yDelouis updated. I decided to use the already existing `EFragmentWithPreferenceSupportHolder` class instead.

WonderCsabo:
@yDelouis do you have further comments, or can we merge this?

yDelouis:
I wasn't satisfied when I read quickly your last changes. So please wait until I review this again and give you more details.

WonderCsabo:
OK, thanks!

WonderCsabo:
@yDelouis i rebased to develop.

AndroidAnnotations/androidannotations/src/main/java/org/androidannotations/holder/HasPreferenceHeaders.java
```diff
@@ -0,0 +1,26 @@
...
+ * License for the specific language governing permissions and limitations under
+ * the License.
+ */
+package org.androidannotations.holder;
+
+import com.sun.codemodel.JBlock;
+import com.sun.codemodel.JVar;
+
+public interface HasPreferenceHeaders extends HasPreferences {
+
```
WonderCsabo: @yDelouis since headers are only used in `PreferenceActivity`, i decided to create a subinterface, so `EFragmentHolder` does not have to implement these methods with empty method blocks.

AndroidAnnotations/androidannotations/src/main/java/org/androidannotations/holder/PreferencesHolder.java
```diff
@@ -0,0 +1,163 @@
...
+			findPreferenceExpression = cast(preferenceClass, findPreferenceExpression);
+		}
+
+		JVar preference = block.decl(preferenceClass, "preference", findPreferenceExpression);
+		return new FoundPreferenceHolder(this, preferenceClass, preference, block);
+	}
+
+	@Override
+	public JBlock getPreferenceScreenInitializationBlock() {
+		// not used
```
WonderCsabo: @yDelouis the returned block is different in `EFragmentHolder` and `EActivityHolder`. It is easier to just return the appropriate block in those classes and not delegating to this. If you think it is not clean, we can create a subclass like `PreferenceFragmentHolder` (like `PreferenceActivityHolder`), and return the init blocks there.

AndroidAnnotations/androidannotations/src/main/java/org/androidannotations/holder/HasPreferenceHeaders.java
yDelouis: Good.

AndroidAnnotations/androidannotations/src/main/java/org/androidannotations/holder/PreferencesHolder.java
yDelouis: I prefer this way. Maybe we could throw an 'UnsupportedOperationException'.

WonderCsabo: I add that right away.

WonderCsabo: @yDelouis Done.

yDelouis:
It's okay.
Thanks for having applied the requested modifications.

dodgex:
duh. i just found that this PR got merged and wanted to update the pref branch in my app and while testing i found a little problem. while this might be working as intended, we may want to reconsider this.

case: i have some `@PreferenceChange` annotated methods. here i update e.g. the summary based on the new value. but now i realized that this annotation uses `preference.setOnPreferenceChangeListener()` to trigger the method what brings me to my problem, this callback is called on change, but BEFORE the change is persisted

> Sets the callback to be invoked when this Preference is changed by the user (but before the internal state has been updated).

i think it would be nice to have both. a callback before and one after. before using these annotations i used `getPreferenceScreen().getSharedPreferences().registerOnSharedPreferenceChangeListener(this);` and implemented `OnSharedPreferenceChangeListener`.

i have another feature branch where i need the post change callback even more as i need to do some logic and fire some events.

WonderCsabo:
@dodgex i used this listener intentionally. This gives a chance to act before the value is persisted. The value will be persisted after it anyway. So why would you need `OnSharedPreferenceChangeListener`? If the preference is not registered after `@PreferenceChange`, you have bigger problems, and that should never happen actually.

dodgex:
in one of my listener i update the summary based on the new value and in another one i even update another preference and fire an event with the new pref value.

WonderCsabo:
Can you give more detail? I really do not understand why you cannot use current `@PreferenceChange` for that.

dodgex:
currently i do something like this
```
ListPreference pref = (ListPreference) findPreference(key);
if (pref != null) {
    pref.setSummary(pref.getEntry());
}
```

where `findPreference(key)` could be replaced by the `Preference` param of the annotated method. the problem is, that `getEntry()` returns the previous value not the new one. so if i e.g. have "world" selected and change the selection to "hello" i would still get "world" as summary. and then if i change it again (to something else) i would get "hello" as summary and not the new one.

WonderCsabo:
`@PreferenceChange` is [passing](https://github.com/excilys/androidannotations/blob/develop/AndroidAnnotations/androidannotations-api/src/main/java/org/androidannotations/annotations/PreferenceChange.java#L64) you the new value.

dodgex:
Argh! Damn. :/ @dodgex RTFM! 

@WonderCsabo Thank you. Nice Work! :+1:

PR #1165, opened 2014-09-26
Clean injected View fields in onDestroyView()

nbelikov:
If setRetainInstance(true) has been set on the fragment or
addToBackStack() was applied to FragmentTransaction, the fragment won't be
destroyed and will keep the reference to its parent view in contentView_
and could lead to memory leak.

This change implements onDestroyView() which cleans the reference.
However, the developer is still responsible for cleaning up the injected
views, because otherwise a (badly implemented) background task may get a
NullPointerException.


WonderCsabo:
Rebased and merged as of faaaae5.

Issue #1235, opened 2014-11-11
@ReceiverAction should not use method name for actions


dodgex:
as discussed here: https://github.com/excilys/androidannotations/pull/1234#discussion_r20171189

the name of `@ReceiverAction` annotated methods should not be used as action name.

WonderCsabo:
This is already implemented.

PR #1236, opened 2014-11-11
1235 do not use methodname as action

dodgex:
implements issue https://github.com/excilys/androidannotations/issues/1235
based on PR https://github.com/excilys/androidannotations/pull/1234 

do not merge before 4.0


yDelouis:
Thanks

PR #1251, opened 2014-11-19
remove model parameter from @OrmLiteDao

dodgex:
according to the [wiki](https://github.com/excilys/androidannotations/wiki/Ormlite) this should have been gone already. ;)

### DIFF
--- a/AndroidAnnotations/androidannotations-api/src/main/java/org/androidannotations/annotations/OrmLiteDao.java
+++ b/AndroidAnnotations/androidannotations-api/src/main/java/org/androidannotations/annotations/OrmLiteDao.java
@@ -26,14 +26,10 @@ import java.lang.annotation.Target;
  * with the provided mode and helper classes.
  * </p>
  * <p>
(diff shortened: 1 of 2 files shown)


yDelouis:
Thanks.

WonderCsabo:
@dodgex can you update the wiki?

dodgex:
https://github.com/dodgex/androidannotations/wiki/Ormlite

i wonder if we want to deprecate `helper` and just use `value` instead (no need to write the attribute name in code).

WonderCsabo:
I think it is a good idea. Actually we can just rename since 4.0 is a
breaking release already. @yDelouis WDYT?

yDelouis:
I'm not sure we should rename it. It tells clearly what parameter is expected.
But if you think it's ok, go for it.

WonderCsabo:
@yDelouis i think you are right.

WonderCsabo:
@dodgex i merged your wiki changes, thanks.

Issue #1209, opened 2014-11-05
Introduce possibility to change GENERATION_SUFFIX


chpasha:
I would be nice to be able to use a generation suffix other then "underscore" - for classes only of course, class members can have the current one. The quick and dirty solution could be to configure it via compilerArgs and OptionsHelper. The better (but more complicated) would be to configure it on class level with AA annotations - e.g @EBean(suffix="Impl") which will produce MyBeanImpl instead of MyBean_ . I know it sounds stupid but it would be great to have more canonical class names inside some reusable lib. What do you think?

WonderCsabo:
Implemented.

PR #1280, opened 2014-12-22
add option to change GENERATION_SUFFIX

dodgex:
see https://github.com/excilys/androidannotations/issues/1209

this allows to set the `generationSuffix` via options. at the moment the same valule is used for all generated stuff (classes, fields, methods) where it before was `_`. But for a first look i already opened the PR.

### DIFF
--- a/AndroidAnnotations/androidannotations/src/main/java/org/androidannotations/helper/ModelConstants.java
+++ b/AndroidAnnotations/androidannotations/src/main/java/org/androidannotations/helper/ModelConstants.java
@@ -20,6 +20,9 @@ import static java.util.Arrays.asList;
 import java.lang.annotation.Annotation;
 import java.util.List;
 
+import javax.annotation.processing.ProcessingEnvironment;
+import javax.lang.model.SourceVersion;
+
 import org.androidannotations.annotations.EActivity;
 import org.androidannotations.annotations.EApplication;
 import org.androidannotations.annotations.EBean;
@@ -35,7 +38,8 @@ import org.androidannotations.annotations.sharedpreferences.SharedPref;
 
 public abstract class ModelConstants {
 
-	public static final String GENERATION_SUFFIX = "_";
+	private static String generationSuffix = "_";
+	private static String classSuffix;
 
 	@SuppressWarnings("unchecked")
 	public static final List<Class<? extends Annotation>> VALID_ENHANCED_VIEW_SUPPORT_ANNOTATIONS = asList(EActivity.class, EViewGroup.class, EView.class, EBean.class, EFragment.class);
@@ -49,4 +53,24 @@ public abstract class ModelConstants {
 	private ModelConstants() {
 	}
 
+	public static void init(ProcessingEnvironment processingEnv) {
+		OptionsHelper optionsHelper = new OptionsHelper(processingEnv);
+		classSuffix = optionsHelper.getClassSuffix().trim();
+
+		if (classSuffix.isEmpty()) {
+			throw new IllegalArgumentException("'" + classSuffix + "' may not be an emtpy string.");
+		}
+
+		if (!SourceVersion.isName(classSuffix) || classSuffix.contains(".")) {
+			throw new IllegalArgumentException("'" + classSuffix + "' is not a valid Java identifier.");
+		}
+	}
+
+	public static String classSuffix() {
+		return classSuffix;
+	}
+
+	public static String generationSuffix() {
+		return generationSuffix;
+	}
 }
--- a/AndroidAnnotations/androidannotations/src/main/java/org/androidannotations/helper/ValidatorHelper.java
+++ b/AndroidAnnotations/androidannotations/src/main/java/org/androidannotations/helper/ValidatorHelper.java
@@ -26,10 +26,10 @@ import static org.androidannotations.helper.CanonicalNameConstants.CLIENT_HTTP_R
 import static org.androidannotations.helper.CanonicalNameConstants.HTTP_MESSAGE_CONVERTER;
 import static org.androidannotations.helper.CanonicalNameConstants.INTERNET_PERMISSION;
 import static org.androidannotations.helper.CanonicalNameConstants.WAKELOCK_PERMISSION;
-import static org.androidannotations.helper.ModelConstants.GENERATION_SUFFIX;
 import static org.androidannotations.helper.ModelConstants.VALID_ANDROID_ANNOTATIONS;
 import static org.androidannotations.helper.ModelConstants.VALID_ENHANCED_COMPONENT_ANNOTATIONS;
 import static org.androidannotations.helper.ModelConstants.VALID_ENHANCED_VIEW_SUPPORT_ANNOTATIONS;
+import static org.androidannotations.helper.ModelConstants.classSuffix;
 
 import java.lang.annotation.Annotation;
 import java.util.ArrayList;
@@ -619,7 +619,7 @@ public class ValidatorHelper {
 			TypeElement typeElement = (TypeElement) element;
 
 			String componentQualifiedName = typeElement.getQualifiedName().toString();
-			String generatedComponentQualifiedName = componentQualifiedName + ModelConstants.GENERATION_SUFFIX;
+			String generatedComponentQualifiedName = componentQualifiedName + classSuffix();
 
 			if (!typeElement.getModifiers().contains(Modifier.ABSTRACT) && !applicationClassName.equals(generatedComponentQualifiedName)) {
 				if (applicationClassName.equals(componentQualifiedName)) {
@@ -647,9 +647,9 @@ public class ValidatorHelper {
 			String elementTypeName = type.toString();
 
 			boolean sharedPrefValidatedInRound = false;
-			if (elementTypeName.endsWith(GENERATION_SUFFIX)) {
-				String prefTypeName = elementTypeName.substring(0, elementTypeName.length() - GENERATION_SUFFIX.length());
-				prefTypeName = prefTypeName.replace("_.", ".");
+			if (elementTypeName.endsWith(classSuffix())) {
+				String prefTypeName = elementTypeName.substring(0, elementTypeName.length() - classSuffix().length());
+				prefTypeName = prefTypeName.replace(classSuffix() + ".", ".");
 
 				Set<? extends Element> sharedPrefElements = validatedElements.getRootAnnotatedElements(SharedPref.class.getName());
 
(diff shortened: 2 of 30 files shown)


AndroidAnnotations/androidannotations/src/main/java/org/androidannotations/helper/ModelConstants.java
```diff
@@ -49,4 +51,12 @@
 	private ModelConstants() {
 	}
 
+	public static void init(ProcessingEnvironment processingEnv) {
+		OptionsHelper optionsHelper = new OptionsHelper(processingEnv);
```
WonderCsabo: @dodgex you can use `SourceVersion.isName` method to the the validation.

dodgex: and what SourceVersion value to use? `SourceVersion.RELEASE_6` or `SourceVersion.RELEASE_7`?

WonderCsabo: https://github.com/excilys/androidannotations/blob/326fe9666df37284245c4e2706f4c042f0b396c1/AndroidAnnotations/androidannotations/src/main/java/org/androidannotations/AndroidAnnotationProcessor.java#L285

WonderCsabo: BTW, isName is static, isn't it? The enum value does not matter.

WonderCsabo: So `if (SourceVersion.isName(classSuffix) || classSuffix.contains("."))`

dodgex: ah. missed that static... well i'll update the PR when i get the time.

AndroidAnnotations/androidannotations/src/main/java/org/androidannotations/helper/ValidatorHelper.java
```diff
@@ -26,7 +26,6 @@
 import static org.androidannotations.helper.CanonicalNameConstants.HTTP_MESSAGE_CONVERTER;
 import static org.androidannotations.helper.CanonicalNameConstants.INTERNET_PERMISSION;
 import static org.androidannotations.helper.CanonicalNameConstants.WAKELOCK_PERMISSION;
-import static org.androidannotations.helper.ModelConstants.GENERATION_SUFFIX;
```
dodgex: the static import is there. but some lines below.

AndroidAnnotations/androidannotations/src/main/java/org/androidannotations/helper/ModelConstants.java
dodgex: updated

WonderCsabo:
This needs and update due to #1325.

dodgex:
it needs. `generationSuffix()` is for everything that is not a class.

WonderCsabo:
Then this PR is totally incomplete, am i right? I just quickly searched in Eclipse "*_*", and it returned lots of results which are not included here.

dodgex:
well i have no idea anymore. but yeah. for some reason are some `_` missing. but if i remember correctly there are some that do not need to be changed as they have a small scope (like a local variable inside an method).

i'll try to get to this one the next days and update it if i can get some time.

dodgex:
updated the PR. but i only changed thos "_" that are not inside a small block like a method. so all generated methods and fields should now use the `generationSuffix()` but locals and params not. not sure if that is needed?

dodgex:
hmf something is wrong with the formatter/save acitons in the eclipse. crtl-shift-f formats the source correctly but saving removes the indentation.

from what i found it seems to be the `additional actions -> correct indentation` option. @WonderCsabo can you check if it does that for you too?  (i tested in EFragmentHolder.java#130)

WonderCsabo:
Saving removes some indentation for me. Notably in case of chained methods.

dodgex:
any chance to get this merged, or is there still something to change? rebasing this PR is not fun. :D

WonderCsabo:
Thanks for your work on this, and sorry for the lots of rebases. 

> updated the PR. but i only changed thos "_" that are not inside a small block like a method. so all generated methods and fields should now use the generationSuffix() but locals and params not. not sure if that is needed?

Two things:

1. Maybe we could just change anything (locals and params, too). But i am not sure. @yDelouis?
2. I think you are missing fields and methods, too. I just searched for ```_"``` in Eclipse (no regex). And i found lots of results which are fields and methods.

dodgex:
wtf... updated the PR with those that where missing. but i'm pretty sure that i had changed all previously. maybe rebase broke it. would not be the first time.

WonderCsabo:
It seems all needed things changed now. Did you test this PR? I mean changing the suffix and see the functional test still green? This would need the rewrite of the whole functional test suite though. :(

dodgex:
changing the suffix can be considered as a breaking change. the tests are only testing the default suffix. but all these test pass when i run `mvn clean verify`

WonderCsabo:
I mean somehow we should verify that changing the suffix is working! Can you at least add a compile time test? Add a processor parameter which changes the suffix, then verify the new class is generated.

dodgex:
working on a test. currenlty we have `classSuffix` used for class names and `generationSuffix` for fields and methods.

in the current implementation only `classSuffix` is changable with a parameter. should i add an option for `generationSuffix` and test it or do we only provide the ability to change `classSuffix` as the user should never get in touch with anything that uses `generationSuffix` as it is only used for internal methods & fields?

WonderCsabo:
`generationSuffix` should be never changed, the purpose of it is to make the suffix consistent through the codebase. It is enough to test `classSuffix` change. Refactoring `generationSuffix` is already tested in the functional test project and it passed. :)

dodgex:
test added

WonderCsabo:
@dodgex i am wondering, should'nt we remove the `_` suffices from the locals and params?

AndroidAnnotations/androidannotations/src/main/java/org/androidannotations/holder/EFragmentHolder.java
```diff
@@ -134,7 +136,7 @@ private void setFindViewById() {
 	}
 
 	private void setFragmentBuilder() throws JClassAlreadyExistsException {
-		fragmentBuilderClass = generatedClass._class(PUBLIC | STATIC, "FragmentBuilder_");
+		fragmentBuilderClass = generatedClass._class(PUBLIC | STATIC, "FragmentBuilder" + classSuffix());
```
WonderCsabo: I do not think it should use `classSuffx`, but `generationSuffix` instead.

AndroidAnnotations/androidannotations/src/main/java/org/androidannotations/holder/EViewHolder.java
```diff
@@ -155,6 +156,6 @@ public JFieldVar getAlreadyInflated() {
 	}
 
 	private void setAlreadyInflated() {
-		alreadyInflated = generatedClass.field(PRIVATE, JType.parse(codeModel(), "boolean"), "alreadyInflated_", JExpr.FALSE);
+		alreadyInflated = generatedClass.field(PRIVATE, JType.parse(codeModel(), "boolean"), "alreadyInflated" + generationSuffix(), JExpr.FALSE);
```
WonderCsabo: You should also used this in the JavaDoc which is added because of this field. You can leave that constant, and replace `_` with `generationSuffix()`.

WonderCsabo:
@dodgex any update on this?

dodgex:
@WonderCsabo i'm currently not at home. Will try to update this PR next week.

AndroidAnnotations/androidannotations/src/main/java/org/androidannotations/holder/EFragmentHolder.java
dodgex: why? its an generated class that is accessible for the user. so `FragmentBuilderImpl` is better than `FragmentBuilder_` actually i'd rename the [`FragmentBuilder`](https://github.com/excilys/androidannotations/blob/326fe9666df37284245c4e2706f4c042f0b396c1/AndroidAnnotations/androidannotations-api/src/main/java/org/androidannotations/api/builder/FragmentBuilder.java) api class (to e.g. `AbstractFragmentBuilder`) and call the generated class FragmentBuilder so this would be CLEAN.

same to other Builder classes.

AndroidAnnotations/androidannotations/src/main/java/org/androidannotations/holder/EViewHolder.java
dodgex: i hope i understood you correctly. :)

AndroidAnnotations/androidannotations/src/main/java/org/androidannotations/holder/EFragmentHolder.java
WonderCsabo: I do not agree. Generally we use this impl classes when there is an interface for a specific work, and then we provide an implementation for that work. Here, we do not provide implementation for the work, but we generate always the interface actually, too.

dodgex: ok. then i'll change that.

AndroidAnnotations/androidannotations/src/main/java/org/androidannotations/holder/EViewHolder.java
WonderCsabo: :+1:

WonderCsabo:
Thanks for your work on this! Can you update the wiki?

WonderCsabo:
Thanks to @dodgex for editing the wiki.

chpasha:
>i do not think we should remove the suffices from anywhere. The suffix also clearly shows that the class/etc. is generated by us, which is very important if you have a bigger project and/or multiple people are working on it

Hi guys, regarding "clearly showing that the class is generated by us" - why not use @Generated annotation which was meant exactly for that? http://docs.oracle.com/javaee/5/api/javax/annotation/Generated.html . As far as I understand, it is enough to modify setGeneratedClass-Method inside BaseGeneratedClassHolder adding a line like this 
>generatedClass.annotate(Generated.class)
				.param("value", AndroidAnnotationProcessor.class.getName())
                                .param("date", getISO8601StringForDate(new Date()));

maybe even use a specific holder FQN instead of generic AndroidAnnotationProcessor.class

WonderCsabo:
@chpasha that annotation is not available in the Android runtime, and will cause problems if we add it, despite its source only retention. BTW lombok has a `lombok.addGeneratedAnnotation` property for this reason.

chpasha:
I didn't know that, so I just was hacking around and got no errors when running some test android project with this annotation applied. Which problems will it cause since it is discarded by compiler?

WonderCsabo:
Okay, it seems `@Generated` is not a problem, i mistaken it for `@ConstructorProperties`.

In that case, i agree with you @chpasha.

chpasha:
great :)

WonderCsabo:
@chpasha unfortunately i remembered correctly. `Transfuse` and `Parceler` for instance adds an own version of `@Generated`, because that annotation is not available in the android.jar.

chpasha:
I'm not sure I get it :) . This annotation is not included in apk because of it's retention policy. So why should we care about if it is present in android.jar? Like I said, for the sake of experiment I added this annotation to the generation process and it doesn't cause any issues at all.

WonderCsabo:
If the annotation is not in the jar, the compilation will fail. This is the case for Eclipse users for instance.

chpasha:
I'm not sure I understand what it has to do with eclipse :) . I must be missing something, but the annotation is present in JDK6 and later, so it is available during compilation. But don't get me wrong, I'm just curious, I don't insist on introducing or implementing anything as long as I can maintain my own fork :)

P. S. actually it was added in dagger2 https://github.com/square/dagger/issues/265  as well ;)

WonderCsabo:
@chpasha it is indeed JDK6, but that is not in the classpath in case when building an Android project in Eclipse. Only the android.jar is on the classpath.

chpasha:
Ok, then maybe adding an annotation processing option that disables @Generated makes sense? Just for the sake of completeness, since this annotation has not much real value anyway.

WonderCsabo:
Yeah, that makes sense, and as i said `Lombok` does the same. However we have to disable it default, to not break silently compilation for Eclipse users. Which means AS users will also not find the option, and never generate these annotations. :(

Issue #1132, opened 2014-09-10
@Receiver never called with OnAttachOnDetach + local


westito:
LocalBroadcastManager can't register the IntentFilter because the IntentFilter.addAction() called in onCreate AFTER onAttach (where the filter added to the manager). The filter is empty when AA try to register the action in manager. Therefore the action won't be registered.

WonderCsabo:
Fixed.

PR #1286, opened 2014-12-23
fix receiver registered at onAttach

dodgex:
see https://github.com/excilys/androidannotations/issues/1132

### DIFF
--- a/AndroidAnnotations/androidannotations/src/main/java/org/androidannotations/holder/EFragmentHolder.java
+++ b/AndroidAnnotations/androidannotations/src/main/java/org/androidannotations/holder/EFragmentHolder.java
@@ -33,9 +33,11 @@ import javax.lang.model.element.TypeElement;
 import javax.lang.model.type.TypeMirror;
 
 import org.androidannotations.annotations.EFragment;
+import org.androidannotations.annotations.Receiver.RegisterAt;
 import org.androidannotations.helper.ActionBarSherlockHelper;
 import org.androidannotations.helper.AnnotationHelper;
 import org.androidannotations.helper.OrmLiteHelper;
+import org.androidannotations.holder.ReceiverRegistrationHolder.IntentFilterData;
 import org.androidannotations.process.ProcessHolder;
 
 import com.sun.codemodel.JBlock;
@@ -65,7 +67,7 @@ public class EFragmentHolder extends EComponentWithViewSupportHolder implements 
 	private JVar injectBundleArgs;
 	private InstanceStateHolder instanceStateHolder;
 	private OnActivityResultHolder onActivityResultHolder;
-	private ReceiverRegistrationHolder receiverRegistrationHolder;
+	private ReceiverRegistrationHolder<EFragmentHolder> receiverRegistrationHolder;
 	private JBlock onCreateOptionsMenuMethodBody;
 	private JVar onCreateOptionsMenuMenuInflaterVar;
 	private JVar onCreateOptionsMenuMenuParam;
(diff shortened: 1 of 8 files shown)


AndroidAnnotations/androidannotations/src/main/java/org/androidannotations/holder/ReceiverRegistrationHolder.java
```diff
@@ -58,17 +58,25 @@ private JFieldVar createIntentFilterField(IntentFilterData intentFilterData) {
...
 		for (String dataScheme : intentFilterData.getDataSchemeSet()) {
-			initBody.invoke(intentFilterField, "addDataScheme").arg(dataScheme);
+			intentFilterTarget.invoke(intentFilterField, "addDataScheme").arg(dataScheme);
 		}
 
 		return intentFilterField;
 	}
 
+	private JBlock getIntentFilterTarget(IntentFilterData intentFilterData) {
+		if (intentFilterData.registerAt.equals(RegisterAt.OnAttachOnDetach) && holder instanceof EFragmentHolder) {
```
WonderCsabo: I think using polymorphism and method overrides is a better design. Maybe you could create this method in `EComponentHolder` which just returns init body, and override it in `EFragmentHolder` to return the needed block?

WonderCsabo:
Can you add a test?

dodgex:
yeah.

AndroidAnnotations/functional-test-1-5/src/test/java/org/androidannotations/test15/receiver/FragmentWithReceiverTest.java
```diff
@@ -0,0 +1,78 @@
...
+	FragmentManager fragmentManager;
+
+	@Before
+	public void setUp() {
+		fragment = new FragmentWithReceiver_();
+		startFragment(fragment);
+	}
+
+	@Test
+	public void defaultReceiverCalledTest() {
```
WonderCsabo: Remove the `Test` suffix from the method names.

dodgex: whoops. done

AndroidAnnotations/androidannotations/src/main/java/org/androidannotations/holder/ReceiverRegistrationHolder.java
WonderCsabo: @dodgex did you miss this?

dodgex: yeah. sorry. i'll update the PR

AndroidAnnotations/androidannotations/src/main/java/org/androidannotations/holder/EFragmentHolder.java
```diff
@@ -547,6 +549,15 @@ public JBlock getOnDetachBeforeSuperBlock() {
 	}
 
 	@Override
+	public JBlock getIntentFilterTarget(IntentFilterData intentFilterData) {
+		if (RegisterAt.OnAttachOnDetach.equals(intentFilterData.getRegisterAt())) {
+			return getOnAttachAfterSuperBlock();
+		} else {
```
WonderCsabo: Remove the else block, return is enough. Also it maybe more flexible to return super.getIntentFilterTarget here.

AndroidAnnotations/androidannotations/src/main/java/org/androidannotations/holder/EComponentHolder.java
```diff
@@ -76,6 +77,10 @@ public JBlock getInitBody() {
 		return getInit().body();
 	}
 
+	public JBlock getIntentFilterTarget(IntentFilterData intentFilterData) {
```
yDelouis: `getIntentFilterTarget` should be a method of `HasReceiverRegistration`.
`EComponentHolder` does not implement this interface so the method should not be there.
Write this method in both `EActivityHolder`, `EServiceHolder` and `EFragmentHolder`. Even if the implementation is the same for `EActivityHolder` and `EServiceHolder`.

dodgex: i decided to go with `getIntentFilterInitializationBlock` as it only initializes the intentFilter.

WonderCsabo:
FInally merged, thanks!

Issue #1284, opened 2014-12-23
ValidatorParameterHelper fluent API


WonderCsabo:
While i was thinking about implementing https://github.com/excilys/androidannotations/pull/1162#issuecomment-66360768 for `@PrefChange`, i realized it it is not easy to implement with existing `ValidatorParameterHelper` code. Then completely refactoring the class into a fluent API came into my mind. So we can write validations like this:
```java
validator.param.inOrder().required().type(STRING).optional().type(INTEGER).validate(isValid);

validator.param.anyOrder().required().extends(VIEW).optional().type(BOOLEAN).validate(isValid);
```

I think this covers any possibilities, and with this all pamater validation can be handled by calling one method chain (for example an additional `zeroOrOneParameter()` call is not needed).

Issue #957, opened 2014-04-06
do you plan support plugins?


hamsterksu:
Hi,

I think some developers want to add custom annotations to "inject" custom libs, which need to handle activity lifecycle. 

So do you plan support plugins? 

My idea: plugin will provide annotations and handlers for them. Also handler should have access to "parent" holder like EActivityHolder/EFragmentHolder/etc

What do you think about it?

Thanks,
Gennadiy

dodgex:
@yDelouis had you any time to work on this? it seems that the need for a plugin system grows from week to week.

WonderCsabo:
@yDelouis any updates on the plugin feature?

WonderCsabo:
@yDelouis several things came into my mind about the plugin system

#### Copying non-AA annotations

When we copy non AA annotations when subclassing and overriding methods, we now have to exclude annotations coming from the plugins from the copy. Before that we just filtered the org.androidannotations namespace, but the plugins will have a different namespace, so we have to explicitely filter those annotations as well.

#### Handler calling older

I think we should add some more intelligent logic for handler calling order, and it is crucial for plugins. For example some plugin may want to called before another handler, and after another handler etc. I think we should add two methods to `AnnotationHandler`: `shouldBeCalledBefore()` and `shouldBeCalledAfter()`. These methods can return list of annotation .class objects. The `AnnotationHandlers` class should try to satisfy all these constraints and build an ordered list.

#### Bundle

I think we should copy all org.androidannotations plugins to the bundle zip.

#### JavaDoc

Since our classes will be an API for the plugin creators, we should add some JavaDoc at least for the most interesting parts.

#### IDE/build tool

@yDelouis have you tried your implementation with all possibilities (Eclipse with no build tool, IntelliJ with no build tool, m2e, IntelliJ with Gradle, IntelliJ + Maven, others)?

WonderCsabo:
Classes to be promoted to be API:

* `handler` package:
  * `AbstractListenerHandler` (maybe)
  * `AnnotationHandler`
  * `AnnotationHandlers` (only the `add()` method, so not sure)
  * `BaseAnnotationHandler`
  * `BaseGeneratingAnnotationHandler`
  * `HasParameterHandlers`
  * `GeneratingAnnotationHandler`

* `helper` package:
  * `AndroidConstants`
  * `AndroidManifest`
  * `AnnotationHelper`
  * `APTCodeModelHelper`
  * `BundleHelper`
[...]

PR #1291, opened 2014-12-25
ValidatorParameterHelper fluent API

WonderCsabo:
Implements #1284.

This is only a WIP, but i am sharing the work, to get an early response about the direction, @yDelouis @dodgex.

### DIFF
--- a/AndroidAnnotations/androidannotations/src/main/java/org/androidannotations/handler/AbstractSeekBarTouchHandler.java
+++ b/AndroidAnnotations/androidannotations/src/main/java/org/androidannotations/handler/AbstractSeekBarTouchHandler.java
@@ -23,6 +23,7 @@ import javax.lang.model.element.ExecutableElement;
 import javax.lang.model.element.VariableElement;
 
 import org.androidannotations.helper.AndroidManifest;
+import org.androidannotations.helper.CanonicalNameConstants;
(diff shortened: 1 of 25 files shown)


WonderCsabo:
@yDelouis i implemented all of your proposals, expect the annotation checking (too specific case, and the classes are already too complicated). Please check out!

WonderCsabo:
For future reference see relevant discussion in https://github.com/WonderCsabo/androidannotations/pull/2.

WonderCsabo:
FInally this is merged. Thanks @yDelouis for your great work on this one!

dodgex:
First of all: nice work @WonderCsabo and @yDelouis :)

from what i have seen, there is no javadoc on these new api? I think it would be nice to have some kind of documentation how this works. especially as soon as we have a plugin system lots of internals (everything that is a public api for plugins) needs to be well documented.

WonderCsabo:
Actually i thought we should add JavaDoc for the most important API-s used by plugins. We should include this one as well.
Also, we may have to move all classes which are not available for plugins to an internal package.

yDelouis:
I agree, and we should discuss it on the issue #957.

WonderCsabo:
:+1: I just see i already mentioned the JavaDoc there. :)

Issue #1067, opened 2014-07-07
@OptionsItem doesn't allow subclasses to override menu item handling 


jacobtabak:
The generated code for onOptionsItemSelected calls your superclass's onOptionsItemSelected() before the subclass, removing your ability to override functionality in the subclass.  I believe the only workaround is to override onOptionsItemSelected() in your activity (rather than use @OptionsItem annotation) because I don't think there's a way to achieve the same goal using AndroidAnnotations.

Here's the generated code.  My base activity class handles android.R.id.home.

```
    @Override
    public boolean onOptionsItemSelected(MenuItem item) {
        boolean handled = super.onOptionsItemSelected(item);
        if (handled) {
            return true;
        }
        int itemId_ = item.getItemId();
        if (itemId_ == android.R.id.home) {
            onHomeTapped();
            return true;
        }
        return false;
    }
```

WonderCsabo:
Implemented.

PR #1296, opened 2014-12-28
Allow subclasses to override @OptionsItem handling

WonderCsabo:
Implements #1067.

@yDelouis i think this may be a breaking change, as clients could rely on the old behavior in their apps.


WonderCsabo:
@yDelouis do you agree?

WonderCsabo:
OK, so it is not a breaking change then. :)

yDelouis:
It's not mentioned in the doc. So we didn't changed the API but the internal implementation.
But we could move this to 4.0 if you think it's important.

WonderCsabo:
This is OK.

Issue #1326, opened 2015-02-20
@EIntentService IntentBuilder's chaining of @ServiceActions is misleading


saschoar:
When you have annotated your `IntentService` with [`@EIntentService`](
https://github.com/excilys/androidannotations/wiki/Enhance-IntentServices) and have annotated multiple methods with `@ServiceAction`, you can use the IntentBuilder like this:

`MyService_.intent(context).myServiceActionOne().myServiceActionTwo().start();`

However, only the last action (`myServiceActionTwo()` in this case) will be executed, because the intent data is replaced. Please consider adding a hint to the documentation.

WonderCsabo:
It would be better to return an object from these action methods which only
has a start() method. This way we can enforce the correct usage at compile
time.

WonderCsabo:
@yDelouis, @dodgex WDYT?

dodgex:
i think `start()` and `get()` and maybe similar. but only those related to the actual intent for this related action.

yDelouis:
I'm okay with your proposal, @WonderCsabo. I hope it won't be too hard to implement.

WonderCsabo:
The implementation is problematic, so i updated the doc and leave the builder as it is.

PR #1329, opened 2015-02-23
ServiceAction intentbuilder methods are no longer chainable

WonderCsabo:
Related to #1326.

This is a **breaking change**, because valid (but discouraged) uses will no longer compile, like this:
```java
MyService._intent(context).actionOne().extra("key", "value").start();
```

### DIFF
--- a/AndroidAnnotations/androidannotations-api/src/main/java/org/androidannotations/api/builder/IntentServiceStarter.java
+++ b/AndroidAnnotations/androidannotations-api/src/main/java/org/androidannotations/api/builder/IntentServiceStarter.java
@@ -16,25 +16,30 @@
 package org.androidannotations.api.builder;
 
 import android.content.ComponentName;
-import android.content.Context;
 import android.content.Intent;
 
-public abstract class ServiceIntentBuilder<I extends ServiceIntentBuilder<I>> extends IntentBuilder<I> implements IntentServiceStarter {
+/**
+ * Provides methods to start {@link android.app.IntentService IntentService}s.
+ */
+public interface IntentServiceStarter {
 
-	public ServiceIntentBuilder(Context context, Class<?> clazz) {
-		super(context, clazz);
-	}
+	/**
+	 * Starts the {@link android.app.IntentService IntentService} by calling
+	 * {@link android.content.Context#startService(android.content.Intent)
+	 * startService} on the previously given {@link android.content.Context
+	 * Context}.
+	 * 
+	 * @return the result of
+	 *         {@link android.content.Context#startService(android.content.Intent)
+	 *         startService}
+	 */
+	ComponentName start();
 
-	public ServiceIntentBuilder(Context context, Intent intent) {
-		super(context, intent);
-	}
+	/**
+	 * Accessor for the built {@link Intent}.
+	 * 
+	 * @return the created {@link Intent}.
+	 */
+	Intent get();
 
-	@Override
-	public ComponentName start() {
-		return context.startService(intent);
-	}
-
-	public boolean stop() {
-		return context.stopService(intent);
-	}
 }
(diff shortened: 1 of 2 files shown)


dodgex:
looks good.

yDelouis:
We won't be able to set extras after the action. I think it could be annoying if we want to share some extra between different actions.
About this example : 
````java
@EIntentService
public class MyIntentService extends AbstractIntentService {

	@Extra
	protected boolean sharedExtra;

	public MyIntentService() {
		super(MyIntentService.class.getSimpleName());
	}

	@ServiceAction
	void actionOne(String valueString) {}

	@ServiceAction("myAction")
	void actionTwo(String valueString, long valueLong) {}
}
````
What about generating this : 
````java
public final class MyIntentService_
    extends MyIntentService
{

    public final static String ACTION_ONE = "actionOne";
    public final static String ACTION_TWO = "myAction";
    public final static String VALUE_STRING_EXTRA = "valueString";
    public final static String VALUE_LONG_EXTRA = "valueLong";
    public final static String SHARED_EXTRA = "sharedExtra";

    public static MyIntentService_.IntentBuilderWithActions_ intent(Context context) {
        return new MyIntentService_.IntentBuilderWithActions_(context);
    }

    private void init_() {
    }

    @Override
    public void onCreate() {
        init_();
        super.onCreate();
    }

    @Override
    public void onHandleIntent(Intent intent) {
        MyIntentService_.super.onHandleIntent(intent);
        String action = intent.getAction();
        if (ACTION_TWO.equals(action)) {
            Bundle extras = intent.getExtras();
            if (extras!= null) {
                String valueStringExtra = extras.getString(VALUE_STRING_EXTRA);
                long valueLongExtra = extras.getLong(VALUE_LONG_EXTRA);
                super.actionTwo(valueStringExtra, valueLongExtra);
                return ;
            }
        }
        if (ACTION_ONE.equals(action)) {
            Bundle extras = intent.getExtras();
            if (extras!= null) {
                String valueStringExtra = extras.getString(VALUE_STRING_EXTRA);
                super.actionOne(valueStringExtra);
                return ;
            }
        }
    }

    public static class IntentBuilderWithActions_
            extends IntentBuilder_
    {
        public IntentBuilderWithActions_(Context context) {
            super(context);
        }

        public IntentBuilder_ actionOne(String valueString) {
            action(ACTION_ONE);
            valueString(valueString);
            return this;
        }

        public IntentBuilder_ actionTwo(String valueString, long valueLong) {
            action(ACTION_TWO);
            valueString(valueString);
            valueLong(valueLong);
            return this;
        }

        public MyIntentService_.IntentBuilder_ valueString(String valueString) {
            return super.extra(VALUE_STRING_EXTRA, valueString);
        }

        public MyIntentService_.IntentBuilder_ valueLong(long valueLong) {
            return super.extra(VALUE_LONG_EXTRA, valueLong);
        }
    }

    public static class IntentBuilder_
        extends ServiceIntentBuilder<MyIntentService_.IntentBuilder_>
    {

        public IntentBuilder_(Context context) {
            super(context, MyIntentService_.class);
        }

        public MyIntentService_.IntentBuilder_ sharedExtra(boolean sharedExtra) {
            return super.extra(SHARED_EXTRA, sharedExtra);
        }
    }
}
````

This way, we can't chain action but we still can add extras after setting the action.

WonderCsabo:
We would have to change lot of things to generate your code. But I agree that the current solution is not convenient. Maybe we should just add a runtime check to chained action methods? Or drop this PR and expect common sense about clients.

WonderCsabo:
I decided to drop this PR. I updated the wiki about this, as it was [suggested](https://github.com/excilys/androidannotations/issues/1326#issue-58338108).

Issue #1346, opened 2015-03-19
Annotating a field with @EBean crashes the processor


WonderCsabo:
The following code
```
@EBean
public class MyBean {

	@EBean // using @EBean instead of @Bean by coincidence
	MyOtherBean injected;
}
```
results in this stacktrace:

java.lang.ClassCastException: org.eclipse.jdt.internal.compiler.apt.model.VariableElementImpl cannot be cast to javax.lang.model.element.TypeElement
	at org.androidannotations.handler.BaseGeneratingAnnotationHandler.isInnerClass(BaseGeneratingAnnotationHandler.java:52)
	at org.androidannotations.handler.BaseGeneratingAnnotationHandler.validate(BaseGeneratingAnnotationHandler.java:40)
	at org.androidannotations.handler.EBeanHandler.validate(EBeanHandler.java:41)
	at org.androidannotations.handler.BaseAnnotationHandler.validate(BaseAnnotationHandler.java:114)

This is because we are casting to `TypeElement` when checking whether it is an inner class, but the annotated element is clearly not a type, but a variable.
We should check first whether the annotated element is a class at all.

WonderCsabo:
This problem only happens in Eclipse, it seems it tries to call annotation processing on the class even if it has other compilation errors (like using a type targeted annotation on a field like in the example above).
The question is, should we handle this @yDelouis @dodgex ?

yDelouis:
We should handle this, if it's not too complicated.

dodgex:
same as @yDelouis

WonderCsabo:
Okay then, i have a fix in my mind, i will implement it soon.

WonderCsabo:
Actually this way a bug in the Eclipse Java compiler, which is now fixed.

PR #1348, opened 2015-03-19
Fix crash in Eclipse when @EBean used on fields

WonderCsabo:
@yDelouis i am not sure this is the good way to solve this. Now errors about constructors are printed on the field... We can avoid this by adding a check and return in every `BaseGeneratingAnnotationHandler` child, but it would be not too clean.

Related to #1346 .

### DIFF
--- a/AndroidAnnotations/androidannotations/src/main/java/org/androidannotations/handler/BaseGeneratingAnnotationHandler.java
+++ b/AndroidAnnotations/androidannotations/src/main/java/org/androidannotations/handler/BaseGeneratingAnnotationHandler.java
@@ -17,6 +17,7 @@ package org.androidannotations.handler;
 
 import javax.annotation.processing.ProcessingEnvironment;
 import javax.lang.model.element.Element;
+import javax.lang.model.element.ElementKind;
 import javax.lang.model.element.TypeElement;
 
 import org.androidannotations.holder.GeneratedClassHolder;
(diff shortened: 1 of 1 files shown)


yDelouis:
No, it's not the right way. We should invalidate the annotation with something like : 
`validatorHelper.isClassOrInterface(element, valid);`

WonderCsabo:
OK, i can change the code. However this invalidation is totally redundant and unnecessary, since the annotation only can be put on a class because of `@Target(ElementType.TYPE)`. I am wondering maybe we should ask Eclipse guys whether this is an intentional behavior of JDT APT.

dodgex:
i think we should ask the Eclipse guys frist instead of trying to work around. if it is a bug on thier side they should fix it not we. only if they say that this behaviour is intended - what makes absolutely no sense - then we should put effort to fix this one.

WonderCsabo:
I sent a mail to  [jdt-apt-dev@eclipse.org](https://dev.eclipse.org/mhonarc/lists/jdt-apt-dev/), but i just realized that list seems to be dead. I will open a ticket in Eclipse bugzilla if i do not get an answer on the list in a few days (which is likely).

dodgex:
or you can ask on the jdt-dev mailing list too.

WonderCsabo:
Good idea, i asked on jdt-dev.

WonderCsabo:
They responded fast, we are moving the conversation to an Eclipse bugzilla [ticket](https://bugs.eclipse.org/bugs/show_bug.cgi?id=463062).

WonderCsabo:
[Fixed](https://bugs.eclipse.org/bugs/show_bug.cgi?id=463062#c5) in Eclipse!

PR #1294, opened 2014-12-27
Fix for Parcelable[] ClassCastException

WonderCsabo:
Fixes #1208.

PR #1352, opened 2015-03-24
Refactor holders which decorate another holder

WonderCsabo:
This PR contains one commit which was split off from #1294.

### DIFF
--- a/AndroidAnnotations/androidannotations/src/main/java/org/androidannotations/holder/ReceiverRegistrationHolder.java
+++ b/AndroidAnnotations/androidannotations/src/main/java/org/androidannotations/holder/ReceiverRegistrationHolder.java
@@ -34,14 +34,13 @@ import com.sun.codemodel.JDefinedClass;
 import com.sun.codemodel.JExpression;
 import com.sun.codemodel.JFieldVar;
 
-public class ReceiverRegistrationHolder<T extends EComponentHolder & HasReceiverRegistration> {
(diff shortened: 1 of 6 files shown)


AndroidAnnotations/androidannotations/src/main/java/org/androidannotations/holder/ReceiverRegistrationHolder.java
```diff
@@ -34,14 +34,13 @@
 import com.sun.codemodel.JExpression;
 import com.sun.codemodel.JFieldVar;
 
-public class ReceiverRegistrationHolder<T extends EComponentHolder & HasReceiverRegistration> {
+public class ReceiverRegistrationHolder extends GeneratedClassHolderDecorator<EComponentHolder> {
```
yDelouis: Why don't you keep `ReceiverRegistrationHolder` generic ? It prevent us from casting and clearly state that the `holder` must implement `HasReceiverRegistration`.

yDelouis:
I'm okay with the purpose of this PR but I would have kept `ReceiverRegistrationHolder` generic (see my comment).

AndroidAnnotations/androidannotations/src/main/java/org/androidannotations/holder/ReceiverRegistrationHolder.java
WonderCsabo: OK, i'll add back the type parameter.

WonderCsabo:
I addressed your review comment.

PR #1358, opened 2015-03-27
Return empty set as @DefaultStringSet default value

### LINKED ISSUE #1357: @DefaultStringSet does not return empty Set of Strings on default
**AA version: 3.2**

The documentation of ```@DefaultStringSet``` states that an empty Set of Strings is returned on default. The generated code on the other hand returns ```null``` by default. 

**Example:**
Base interface:
```java
@SharedPref(SharedPref.Scope.UNIQUE)
public interface PrefsClass {
    /* Documentation of @DefaultStringSet states:
     * "Use on methods in SharedPref annotated class.
     * The generated method will return an empty java.util.Set of Strings by default."
     */
    @DefaultStringSet
    public Set<String> stringSetPref();
}
```

Generated code:
```java
public final class PrefsClass_ extends SharedPreferencesHelper{
    // ...
    public StringSetPrefField stringSetPref() {
        // returns null instead of empty Set
        return stringSetField("stringSetPref", null);
    }
}
```

Expected code:
```java
public final class PrefsClass_ extends SharedPreferencesHelper{
    // ...
    public StringSetPrefField stringSetPref() {
        // returns a new empty set (not necessarily HashSet)
        return stringSetField("stringSetPref", new HashSet<String>());
    }
}
```

WonderCsabo:
Fixes #1357.


yDelouis:
Great !

Issue #1218, opened 2014-11-07
Cancel delayed function


sytolk:
If I have this delayed function
```java
   @UiThread(delay = 50000)
    public void timerToFinish() {
    }
```

generated code is this:
```java
    @Override
    public void timerToFinish() {
        handler_.postDelayed(new Runnable() {

            @Override
            public void run() {
                PriceActivity_.super.timerToFinish();
            }

        }
        , 50000L);
    }
```

I want to cancel (for some reason) execution of this delayed method before execution:

http://stackoverflow.com/questions/4378533/cancelling-a-handler-postdelayed-process

to do that in AA its need to have this new Runnable() stored.
Maybe all new Runnable its need to be stored in Map in generated class. If I write this:
```java
@UiThread(delay = 50000, id="timerToFinish")
    public void timerToFinish() {
    }

    public void cancelTimerToFinish() {
AA_.cancelDelayed("timerToFinish")
    }
```
and in generated class needs to have:
```java
ConcurrentHashMap runnableMap<String,Runnable>

 public void timerToFinish() {
       runnableMap.put("timerToFinish",new Runnable() {
            @Override
            public void run() {
                PriceActivity_.super.timerToFinish();
            }
        }); 
        handler_.postDelayed(runnableMap.get("timerToFinish"), 50000L);
    }

public void cancelDelayed(String runnableName){
  if(runnableMap.contains(runnableName)
     handler_.removeCallbacks(runnableMap.get(runnableName));
}
```
I have not tested this but I think that will work.

WonderCsabo:
Implemented.

PR #1359, opened 2015-03-28
Cancel @UiThread tasks

WonderCsabo:
Implements #1218.

### DIFF
--- a/AndroidAnnotations/androidannotations-api/src/main/java/org/androidannotations/api/UiThreadExecutor.java
+++ b/AndroidAnnotations/androidannotations-api/src/main/java/org/androidannotations/api/UiThreadExecutor.java
@@ -1,0 +1,84 @@
+/**
+ * Copyright (C) 2010-2015 eBusiness Information, Excilys Group
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may not
+ * use this file except in compliance with the License. You may obtain a copy of
+ * the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed To in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+ * License for the specific language governing permissions and limitations under
+ * the License.
+ */
+package org.androidannotations.api;
+
+import java.util.Iterator;
+import java.util.Map;
+import java.util.Map.Entry;
+import java.util.Set;
+import java.util.WeakHashMap;
+
+import android.os.Handler;
+
+/**
+ * This class provide operations for
+ * {@link org.androidannotations.annotations.UiThread UiThread} tasks.
+ */
+public class UiThreadExecutor {
+
+	private static final Map<Runnable, TaskInfo> TASKS = new WeakHashMap<Runnable, TaskInfo>();
+
+	private UiThreadExecutor() {
+		// should not be instantiated
+	}
+
+	/**
+	 * Store a new task in the map for providing cancellation. This method is
+	 * used by AndroidAnnotations and not intended to be called by clients.
+	 * 
+	 * @param id
+	 *            the identifier of the task
+	 * @param task
+	 *            the task itself
+	 * @param handler
+	 *            the {@link Handler} which runs the task
+	 */
+	public static synchronized void addTask(String id, Runnable task, Handler handler) {
+		TASKS.put(task, new TaskInfo(id, handler));
+	}
+
+	/**
+	 * Cancel all tasks having the specified <code>id</code>.
+	 * 
+	 * @param id
+	 *            the cancellation identifier
+	 */
+	public static synchronized void cancelAll(String id) {
+		Set<Entry<Runnable, TaskInfo>> entrySet = TASKS.entrySet();
+		Iterator<Entry<Runnable, TaskInfo>> iterator = entrySet.iterator();
+
+		while (iterator.hasNext()) {
+			Entry<Runnable, TaskInfo> next = iterator.next();
+			Runnable task = next.getKey();
+			TaskInfo info = next.getValue();
+
+			if (info.id.equals(id)) {
+				info.handler.removeCallbacks(task);
+			}
+		}
+	}
+
+	private static class TaskInfo {
+		String id;
+
+		Handler handler;
+
+		TaskInfo(String id, Handler handler) {
+			this.id = id;
+			this.handler = handler;
+		}
+	}
+}
--- a/AndroidAnnotations/androidannotations-api/src/main/java/org/androidannotations/annotations/UiThread.java
+++ b/AndroidAnnotations/androidannotations-api/src/main/java/org/androidannotations/annotations/UiThread.java
@@ -87,8 +87,39 @@ import java.lang.annotation.Target;
  * 
  * </blockquote>
  * 
+ * <h2>Cancellation</h2>
+ * <p>
+ * You can cancel UiThread tasks if you provide an id (which cannot be an empty
+ * string) with the {@link #id()} parameter. Please note tasks which use
+ * <code>REUSE</code> {@link #propagation()} cannot have an id hence cannot be
+ * cancelled. To cancel all {@link UiThread} tasks with a given id, call
+ * {@link org.androidannotations.api.UiThreadExecutor#cancelAll(String)
+ * UiThreadExecutor#cancelAll(String)}.
+ * </p>
+ * 
+ * <blockquote> <b>Example</b> :
+ * 
+ * <pre>
+ * &#064;EBean
+ * public class MyBean {
+ * 
+ * 	&#064;UiThread(id = &quot;myId&quot;)
+ * 	void uiThreadTask() {
+ * 		// do sg
+ * 	}
+ * }
+ * 
+ * ...
+ * 
+ * UiThreadExecutor.cancelAll(&quot;myId&quot;);
+ * </pre>
+ * 
+ * </blockquote>
+ * 
+ * 
  * @see Background
  * @see android.os.Handler
+ * @see org.androidannotations.api.UiThreadExecutor#cancelAll(String)
  */
 @Retention(RetentionPolicy.CLASS)
 @Target(ElementType.METHOD)
@@ -127,4 +158,17 @@ public @interface UiThread {
 		 */
 		REUSE
 	}
+
+	/**
+	 * Identifier for cancellation.
+	 * 
+	 * To cancel all tasks having a specified id:
+	 * 
+	 * <pre>
+	 * UiThreadExecutor.cancelAll(&quot;my_background_id&quot;);
+	 * </pre>
+	 * 
+	 * @return the id for cancellation
+	 */
+	String id() default "";
 }
--- a/AndroidAnnotations/androidannotations/src/main/java/org/androidannotations/handler/UiThreadHandler.java
+++ b/AndroidAnnotations/androidannotations/src/main/java/org/androidannotations/handler/UiThreadHandler.java
@@ -23,8 +23,11 @@ import javax.lang.model.element.Element;
 import javax.lang.model.element.ExecutableElement;
 
 import org.androidannotations.annotations.UiThread;
+import org.androidannotations.api.UiThreadExecutor;
 import org.androidannotations.helper.APTCodeModelHelper;
 import org.androidannotations.holder.EComponentHolder;
+import org.androidannotations.model.AnnotationElements;
+import org.androidannotations.process.IsValid;
 
 import com.sun.codemodel.JBlock;
 import com.sun.codemodel.JClass;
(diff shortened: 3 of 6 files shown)


Artyomcool:
Hey!  I'm sorry for saying this, but you shouldn't implement it in this way:
1. Runnales will not be removed from the map, so we have a huge leak.
2. You have a huge job on map iteration under synchronization (and explicit synchronization on TASK is overshooting).
There is a better way: after execution runnable should remove itself and key should be an id with set of runnables as value. In this case you will have time complexity O(1) while managing runnables. Also there is no need to store handlers - you can use one static handler in the UiThreadExecutor. Hope it helps ))

WonderCsabo:
Thanks for your comments!

1. They will be, that is why i am using a `WeakHashmap`.
2. BackgroundExecutor.cancelAll does the same, so that is why i thought it will suffice here, too. BTW, i should have synchronized to the .class actually.

I also considered this, but i went this way because we do not need to modify the code generation too much. I am not sure about storing one static `Handler`. That will leak for sure. BTW i checked the `BackgroundExecutor` commit history, and it seems it used a map too, but it was refactored to a list implementation for simplicity.
Lastly, we would still need locking.

AndroidAnnotations/androidannotations-api/src/main/java/org/androidannotations/api/UiThreadExecutor.java
```diff
@@ -0,0 +1,91 @@
...
+	/**
+	 * Cancel all tasks having the specified <code>id</code>.
+	 * 
+	 * @param id
+	 *            the cancellation identifier
+	 */
+	public static synchronized void cancelAll(String id) {
+		Set<Entry<Runnable, TaskInfo>> entrySet = TASKS.entrySet();
+
+		synchronized (TASKS) {
```
Artyomcool: You already have synchronization on the class, thats why you don't need a second synchronized section.

Artyomcool:
Right, I've missed the weak-part around the map :)
Actually, you even don't need to use weak-reference with handler, because handler doesn't hold a reference on runnable object. It is held in a message under looper's queue.
And that's why the single handler will not cause any _new_ leaks (all runnables are held in the main looper anyway).
And yes, I've seen the ```BackgroundExecutor```'s code, and this is why we don't use a cancel mechanism of AA in our project. But I'm really looking forward the cancel ability for UI-tasks, and I don't want it to be slow in a case of many tasks inside the map. Maybe I'll just prepare my own PR so you will be able to choose?

AndroidAnnotations/androidannotations-api/src/main/java/org/androidannotations/api/UiThreadExecutor.java
WonderCsabo: You are right, i forget to remove it here.

WonderCsabo:
> Actually, you even don't need to use a weak-reference with handler 

I am using a `WeakReference` for the `Handler` because the implementation note of the class [JavaDoc](http://docs.oracle.com/javase/7/docs/api/java/util/WeakHashMap.html).

> And that's why the single handler will not cause any *new* leaks (all runnables are held in the main looper anyway).

I see.

> this is why we don't use a cancel mechanism of

I would gladly accept a PR which makes things better, actually i think you already proposed an issue for that. :smile:

> Maybe I'll just prepare my own PR so you will be able to choose?

Okay. You can cherry-pick some of my commits i think.

Artyomcool:
> I am using a WeakReference for the Handler because the implementation note of the class JavaDoc.

Yep, notes says that: Thus care should be taken to ensure that value objects do not strongly refer to their own keys, either directly or indirectly.
And it is true. That is why I've just checked the ```Handler```'s code to see if it holds some references to the runnables. And it doesn't. It references the looper and it's queue, that are holding the runnables through messages, but it doesn't metter - main looper is singleton anyway.


> I would gladly accept a PR which makes things better, actually i think you already proposed an issue for that.

True, thanks for reminding :)

> You can cherry-pick some of my commits i think.

Actually, I was thinking about implementing it on top of yours. Then, if the code is ok, I'll make things in git more clear - looks like it will take some extra work, so I would like to do it only in a case you will like the code :)

WonderCsabo:
> Yep, notes says that: Thus care should be taken to ensure that value objects do not strongly refer to their own keys, either directly or indirectly.

Yeah, you are right, i misinterptered the next sentence.

> Actually, I was thinking about implementing it on top of yours. Then, if the code is ok, I'll make things in git more clear - looks like it will take some extra work, so I would like to do it only in a case you will like the code 

That's OK.

AndroidAnnotations/androidannotations-api/src/main/java/org/androidannotations/api/UiThreadExecutor.java
WonderCsabo: I removed this sync block, and also get rid of the unnecessary `WeakReference`.

Artyomcool:
Well... Looks like I have no idea, how to checkout your branch in git. Could you give me some hint?

WonderCsabo:
This should do:

Git remote add WonderCsabo myurl
Git fetch WonderCsabo
Git checkout WonderCsabo/branch
Git checkout -b newname

Artyomcool:
Thanks, it helped a lot.
I've done it in my branch: https://github.com/Artyomcool/androidannotations/commit/faaadbec506a2e08d93e1b67695dc4c9faf5aad0
It is just a scetch, I hadn't time to check it :sweat_smile: 
The only purpose of this commit - to share an idea of more straightforward way to implement ui tasks cancellation feature.
If it looks good for you, I'll perform all necessary work around git and tests. If you like your solution more - ok, no hard feellings :smile:

WonderCsabo:
Thanks, it seems to be good. Please note that BackgroundExecutor also used maps, but it was refactored to use a list. So the previous contributor decided code simplicity worths the running time complexity (so that is why I did here) .

I wonder can we use a concurrent data structure instead of synchronization? 

You mentioned a static handler. Also we should use a finally block to remove the runnable.

Artyomcool:
Good point about finally block. And you are right about need of wise choice
between complexity and performance. But you are suggest a WeakHashMap, that
is very advance technic (I'm a big fan of weak references actually, but it
needs a lot of care).
If you let me refactor the code with handlers, we won't need a second map
and it will be much simplier.
I've thought about concurrent collections. It will be very tricky, but
looks possible. Should I try to implement it?
04 апр. 2015 г. 17:16 пользователь "Csaba Kozák" <notifications@github.com>
написал:

WonderCsabo:
If you have time, yeah you can try the concurrent way. Of course the static
handler will remove the need of the second map. I am waiting for the
further improvements. ☺

Artyomcool:
Hey, I've done the concurrent way in my branch: https://github.com/Artyomcool/androidannotations/commit/25fa9af78818459ca289145503ae702f78726380. But looks like we shouldn't use it in production, I've done just for fun as a proof of concept. Please look into the code: it is very tricky. It easy to break with even small change. And this is not the only reason to fall back to the simple synchronization way. The most time consuming part is still under synchronization becuase of Android implementation details: when you are canceling a runnable, message queue will perform a full scan under synchronization block (http://grepcode.com/file/repository.grepcode.com/java/ext/com.google.android/android/5.0.2_r1/android/os/MessageQueue.java#MessageQueue.removeMessages%28android.os.Handler%2Cjava.lang.Runnable%2Cjava.lang.Object%29).
Are you agree that it is much better to simplify the code with dummy synchronization block? Or it is better to cover it with a lot of tests?

WonderCsabo:
Thanks for experimenting! Yeah, we should go with the simpler way.

Artyomcool:
Hi there! Looks like I've found the right way to do the job. Please, look at my branch once again. Promise, you'll love it.

WonderCsabo:
Impressive! However it is not clear to me that why don't you removing anything from `TOKENS` in `cancelAll` and after normal task ending. These are still missing, are am i blind? :wink:

Artyomcool:
Your eyes are fine :smile: 
I'm pretty sure, that removing tokens is unneccessary. Every string constant already exists permanently at the constant pool (at the constant pools of every class that using it directly, actually). And we are not concerned about it, right? I just map every ui task id to another object - token. It is extremely low cost  for such purpose. In exchange we get the ability to do nothing special within runnables. I mean - when you are creating a new annonymous class you persistently retain much more memory then with a dummy object. I would document an addTask method to clearify, that is not supposed to be used directly. What do you think?

WonderCsabo:
Yeah, i am aware of String constant storing.

Correct, we will retain some memory which is proportional to the number task IDs used in the app. Another drawback is if the task was finished, and the user cancels it later, we will still search the `MessageQueue` under synchronization for nothing.

Artyomcool:
> Another drawback is if the task was finished, and the user cancels it later, we will still search the MessageQueue under synchronization for nothing.

Well, true. But android do it all the time. Ok, I can actually do just you are asking for: keep track of runnables count and clean up the map. But I'll drop the concurrent collection in this case to simplify the code.

WonderCsabo:
Hey, wait. :smile: Actually i do not see clearly which is the better solution. If you drop the concurrent collection, you will re-introduce the synchronization at our side?

Artyomcool:
True, but lookup for O(1) is actually nothing (for android applications). Even under synchronization.

Artyomcool:
If you will have some issue with this, you will have a lot of performance issues even without AA - synchronization under MessageQueue much worse.

WonderCsabo:
I did not ask this performance-wise, but implementation-wise.

OK, let's move to that way. However do not drop the current code totally. :wink:

Really thanks for working on this in this detail!

Artyomcool:
Ok, but let me suggest another idea for achieving the goal :smile_cat:

Artyomcool:
Take a look:
```
@UiThread
Future<Void> someUiTask() {
    //...some work here
    return Stubs.future();
}

class Stubs {

    private static final Future FUTURE = new Future() {
        public boolean cancel(boolean mayInterrupt) {
            throw new UnsupportedOperationException("Method is not overrided by AA");
        }

        //same implementation of other methods
        //...
    }

    public static <T> Future<T> future() {
         return (Future<T>)FUTURE;
    }
   
}
```
AndroidAnnotations will override the method providing the actual Future, that allows cancel the task, check the state and even fetch a result. We will also need some aggregator to reduce boilerplate, e.g.:
```
class SomeActivity extends Activity {

    FutureAggregation aggregation = new FutureAggreation();

    void someMethod() {
        aggregation.add(someOtherMethod());
    }

   @UiThread
   Future<Void> someOtherMethod() {
      //...
   }

   void onDestroy() {
       super.onDestroy();
       aggregation.cancelAll();
   }

}
```
Moreover, we can generate aggregators for onPause/onStop/onDestory methods like:
```
@EActivity
class SomeActivity extends Activity {
    void someMethod() {
        SomeAAHelperClass.cancelOnDestroy(this, someOtherMethod());
    }

   @UiThread
   Future<Void> someOtherMethod() {
      //...
   }
}

//...
public static void cancelOnDestroy(Activity activity, Future<?> future) {
    SomeMagicInterface generatedInterface = (SomeMagicInterface) activity;
    generatedInterface.getCancelOnDestroyAggregator().add(future);
}
//...
```
It is much more sophisticated, but it covers very interresting cases. Maybe such implementation should wait for the plugin system?

Artyomcool:
Well, looks like this task gives me a lot of fun :smiling_imp:

WonderCsabo:
BTW, i am curious, why we even need the `Object`s? We could just use the String literals themselves as the tokens. Maybe because we need the `putIfAbsent` method?

WonderCsabo:
Your last code is too complicated from the user point of view, sorry.

Artyomcool:
> Your last code is too complicated from the user point of view, sorry

You are right. Most of users will just not need this. I'll wait for a plugin system for that.

Artyomcool:
> BTW, i am curious, why we even need the Objects? We could just use the String literals themselves as the tokens. Maybe because we need the putIfAbsent method?

Well, right vice versa. We need the putIfAbsent because we can't use the literals.
Or you mean smth like that:
```
ConcurrentMap<String, String> map = new ConcurrentHashMap<String,String>();
//...
String literal = ...
literal = map.putIfAbsent(literal, literal);
```
It is an interresting idea, but a little bit tricky and we are winning ~20 bytes per id. Not much actually, but why not... While the ```HANDLER``` is not accessible for users there is no danger in this, but if it is, user will be able to cancel some tasks occasionally. I'm talking about the ```MessageQueue``` implementation - it checks the handler while cancelling messages, so it is safe to use some shared data as a tag. So, should we do it in this way?

Artyomcool:
> ```literal = map.putIfAbsent(literal, literal);```

Sorry, wrong. Should be:
```
String oldLiteral = map.putIfAbsent(literal, literal);
if (oldLiteral != null) {
    literal = oldLiteral;
}
```

WonderCsabo:
No, you are right, we should not.

WonderCsabo:
I am closing this because we will another solution as discussed above.

PR #1366, opened 2015-04-16
Refactor more holder decorators

WonderCsabo:
I left out two holders in the previous PR.


WonderCsabo:
@yDelouis if we merge #1162, we can move common methods from `InstanceStateHolder` and `PreferencesHolder` to another class, so i suggest to wait with merging this PR.

PR #1371, opened 2015-04-20
Invalidate non-abstract inner classes with abstract outer class

### LINKED ISSUE #1345: androidannotations.log file constantly grows
Hi!

I’ve been working with this framework for a while and I've never encountered this problem before:

The "androidannotations.log" file constantly grows, 

and the tasks “:app:assembleDebug” never finishes and the file gets more and more heavy (3 gigas!).

screen shot 2015-03-19 at 15 05 13

I will go to my previous commit -before this mess happened, and see what might be causing this problem.

Any thoughts?

WonderCsabo:
Fixes #1345.


yDelouis:
Thanks.

Issue #1378, opened 2015-04-23
Add services from API 21 and 22


WonderCsabo:
New services appeared in these APIs, and `@SystemService` should support them.

WonderCsabo:
Implemented.

PR #1379, opened 2015-04-23
Add @SystemService support for new services in API 21 and 22

WonderCsabo:
Implements #1378.


AndroidAnnotations/functional-test-1-5/src/main/java/org/androidannotations/test15/ActivityWithServices.java
```diff
@@ -121,6 +125,9 @@
 	@SystemService
 	DropBoxManager dropBoxManager;
 	
+	// @SystemService
+	// EthernetManager ethernetManager; // hidden API
```
dodgex: Hidden API?

WonderCsabo: Yeah, please see this
https://github.com/android/platform_frameworks_base/blob/master/core/java/android/content/Context.java

We support internal services since 1fee3bbc. These are useful for
developers of some kind of core apps, but i cannot find the issue.

dodgex: no issue. just saw that and wonderd what it means. i'm ok with it if it works. have you tested to use that to inject a field? i once played with an hidden api and had to use reflection to use it as the android.jar that is used for compilation did not contain the method.

WonderCsabo: No, these services are intended for custom Android builds. See the [release notes](https://github.com/excilys/androidannotations/wiki/ReleaseNotes#22):

 > and even hidden system services (useful for custom Android builds)

Actually these Services classes are not available in the normal android.jar at compile time, so this only works if you are building against the AOSP tree or you create a custom internal jar.

WonderCsabo: Actually i just noticed that i missed some of the [new hidden services](https://github.com/android/platform_frameworks_base/blob/master/core/java/android/content/Context.java). I will add them later today.

dodgex:
ah okay. so these are support for those who do "special" apps ;) then its ok.

WonderCsabo:
PR updated.

yDelouis:
Thanks.

PR #1380, opened 2015-04-23
Handle special case when injecting AppWidgetManager with @SystemService

### LINKED ISSUE #1377: @SystemService & android.appwidget.AppWidgetManager
Given I inject the AppWidgetManager service into a Fragment like this:

```
@SystemService
protected AppWidgetManager appWidgetManager;
```

Then the generated code looks like this:
```
appWidgetManager = ((AppWidgetManager) getActivity().getSystemService(Context.APPWIDGET_SERVICE));
```

However, this will result in NPE, because AppWidgetManager can not be retrieved like that.

Instead, the following code should be generated:
```
appWidgetManager = AppWidgetManager.getInstance(getActivity());
```

WonderCsabo:
Fixes #1377.

### DIFF
--- a/AndroidAnnotations/androidannotations/src/main/java/org/androidannotations/handler/SystemServiceHandler.java
+++ b/AndroidAnnotations/androidannotations/src/main/java/org/androidannotations/handler/SystemServiceHandler.java
@@ -15,20 +15,27 @@
  */
 package org.androidannotations.handler;
 
+import static com.sun.codemodel.JExpr.assign;
 import static com.sun.codemodel.JExpr.cast;
 import static com.sun.codemodel.JExpr.ref;
 
 import javax.annotation.processing.ProcessingEnvironment;
 import javax.lang.model.element.Element;
+import javax.lang.model.element.TypeElement;
 import javax.lang.model.type.TypeMirror;
 
 import org.androidannotations.annotations.SystemService;
+import org.androidannotations.helper.CanonicalNameConstants;
 import org.androidannotations.holder.EComponentHolder;
 import org.androidannotations.model.AnnotationElements;
 import org.androidannotations.process.IsValid;
 
 import com.sun.codemodel.JBlock;
+import com.sun.codemodel.JClass;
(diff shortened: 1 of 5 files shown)


dodgex:
Looks good. +1 for checking the sdk & current Plattform to decide how to inject. 

I think you verified the results, but is there a way to add a test?

WonderCsabo:
Thanks for checking the PRs!

I wanted to add tests, but unfortunately there is no way to test this, because we do not have the API 21 classes at compile time. Actually we could just provide a dummy class, but we cannot add a new field to `Build.VERSION_CODES`.

WonderCsabo:
Hmm, i just thought about the compile time test (because this should be done there), but maybe we could do it in the functional test with the help of Robolectric.

Actually it is a pain that we still depend on Android 4.x. I tend to create a solution to create a script which installs the Android dependency to the local maven repo, as lots of Android projects the (this is [needed](https://github.com/robolectric/robolectric/blob/master/scripts/install-dependencies.sh) for the compilation of Robolectric itself, for example).

WonderCsabo:
@dodgex i added a commit which tests one case of injection (when only the old method is available). It would be nice to do it in compile time test, but we cannot do that currently.

dodgex:
@WonderCsabo i assumed that it would be hard to get proper tests for that. lets hope no one breaks that accidently :D

yDelouis:
I'm okay with the code but the Travis build has failed. Could you fix this ?

WonderCsabo:
Thanks for checking!

The Travis failure is not related to this PR, this passed. For some reason, it fails while [unzipping](https://travis-ci.org/excilys/androidannotations/jobs/60349974#L4026) the Gradle distribution. I am seeing this problems lately. I try to investigate.

WonderCsabo:
Maybe Travis is unstable [now](https://twitter.com/traviscistatus/status/592902357144498176). I also see other errors, like build freezing while downloading Maven deps.

yDelouis:
OK. Feel free to merge this PR if it builds locally.

PR #1382, opened 2015-04-25
Fix validation of Extra and FragmentArg annotated field

### LINKED ISSUE #1313: Serializable List not working with @FragmentArg
I have a list of serializable objects List\<Leg> legs;
It is annotated with @Extra in my activity and with @FragmentArg in my fragment.

When I send this list from one activity to another via 
RouteLegsActivity_.intent(this).legs(legs).start();
it works find and the generated code is 

```java
public RouteLegsActivity_.IntentBuilder_ legs(List<Leg> legs) {
            return super.extra(LEGS_EXTRA, ((Serializable) legs));
}
```

When I try to do the same to the fragment via RouteLegsFragment_.builder().legs(legs).build()
the generated code is:

```java
public RouteLegsFragment_.FragmentBuilder_ legs(List<Leg> legs) {
            args.putSerializable(LEGS_ARG, legs);
            return this;
}
```

This does not compile because of the missing cast to Serializable.
This seems like a bug to me. Or am I doing something wrong?

yDelouis:
Fix #1313.

### DIFF
--- a/AndroidAnnotations/androidannotations/src/main/java/org/androidannotations/handler/ExtraHandler.java
+++ b/AndroidAnnotations/androidannotations/src/main/java/org/androidannotations/handler/ExtraHandler.java
@@ -15,16 +15,14 @@
  */
 package org.androidannotations.handler;
 
-import static com.sun.codemodel.JExpr.invoke;
-import static com.sun.codemodel.JExpr.lit;
-import static com.sun.codemodel.JMod.FINAL;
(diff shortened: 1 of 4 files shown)


WonderCsabo:
This seems to be okay. Can you revert the unnecessary import modifications so the diffs become more clear?

WonderCsabo:
Merged as of commit 3f6e510.

Thanks!

Issue #1383, opened 2015-04-26
do not "Correct indentation" as save action


dodgex:
The "Correct indentation" Task in save action breaks some formatting that is applied by the formatter (Ctrl+Shift+F). See https://github.com/excilys/androidannotations/pull/1280#issuecomment-74824824

PR #1384, opened 2015-04-26
disable "correct indentation" option

dodgex:
Fixes https://github.com/excilys/androidannotations/issues/1383


WonderCsabo:
Please do the same for the other projects, too. At least the test project has the same issues I think. The api projects does not.

dodgex:
i checked all projects that start with `androidannotations` in my eclipse but as the test project is just called `functional-test-1-5` i missed that one.

PR updated.

WonderCsabo:
Thanks. Actually i have an idea about dropping these auto format settings. We have now Checkstyle, so these are quiet redundant. What i suggest is:

* run the checkstyle:check task in the standard build
* the IDE should run this task, so formatting is still verified. This is done with the m2e-code-quality plugin and eclipse-cs plugin in Eclipse.
* drop the formatter and the save auto cleaner profiles. The contributor still can generate these dynamically if he/she wants with the plugins i listed above, so the same convenience can be achieved.

This is a common setup in OSS project, for example the maven android plugin does the same.

dodgex:
well while i think having the project files in the repo is very convenient, i'd agree. but instead of just removing the formatter settings i'd remove all eclipse project specific files - if possible.

WonderCsabo:
> instead of just removing the formatter settings i'd remove all eclipse project specific files - if possible

I meant that. Actually i really do not like seeing IDE settings files in the repository. :)

dodgex:
but that MIGHT require an update to the wiki. i think we should merge this regardless of the plan to remove the settings and then check if we can get rid of the project files.

WonderCsabo:
i think we should merge this regardless of the plan to remove the settings and then check if we can get rid of the project files.

I am merging this, of course. :)

PR #1395, opened 2015-05-04
Issue 1218: cancelling ui thread tasks

Artyomcool:
PR implements cancelling by using ids, like in background threads.

### DIFF
--- a/AndroidAnnotations/androidannotations-api/src/main/java/org/androidannotations/api/UiThreadExecutor.java
+++ b/AndroidAnnotations/androidannotations-api/src/main/java/org/androidannotations/api/UiThreadExecutor.java
@@ -15,10 +15,14 @@
  */
 package org.androidannotations.api;
 
-import java.util.*;
-import java.util.Map.Entry;
-
 import android.os.Handler;
+import android.os.Looper;
+
+import java.util.Collections;
+import java.util.Set;
+import java.util.concurrent.ConcurrentHashMap;
+import java.util.concurrent.ConcurrentLinkedQueue;
+import java.util.concurrent.atomic.AtomicInteger;
 
 /**
  * This class provide operations for
@@ -26,8 +30,15 @@ import android.os.Handler;
  */
 public class UiThreadExecutor {
 
-	private static final Map<String, List<Runnable>> TASKS = new HashMap<String, List<Runnable>>();
-	private static final Map<Runnable, Handler> HANDLERS = new HashMap<Runnable, Handler>();
+	private static final Handler HANDLER = new Handler(Looper.getMainLooper());
+
+	static final ConcurrentHashMap<String, Set<Runnable>> TASKS = new ConcurrentHashMap<String, Set<Runnable>>();
+
+	private static final Set<Runnable> JUST_POSTED_TASKS = Collections.newSetFromMap(new ConcurrentHashMap<Runnable, Boolean>());
+
+	private static final ConcurrentLinkedQueue<Set<Runnable>> ALLOCATION_STASH = new ConcurrentLinkedQueue<Set<Runnable>>();
+	private static final AtomicInteger ALLOCATION_STASH_SIZE = new AtomicInteger(0);
+	private static final int ALLOCATION_STASH_MAX_SIZE = 16;
 
 	private UiThreadExecutor() {
 		// should not be instantiated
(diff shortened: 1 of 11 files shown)


WonderCsabo:
This has merge conflicts. Can you rebase to develop?

Could you squash the changes of the last two commits to the relevant previous commit(s)? These do not add any interesting info later in the history.

Artyomcool:
done

2015-05-05 0:00 GMT+03:00 Csaba Kozák <notifications@github.com>:

WonderCsabo:
Thanks!

You have one [checkstyle violation](https://travis-ci.org/excilys/androidannotations/jobs/61210883#L517). Please edit the commit which added that.

AndroidAnnotations/functional-test-1-5/src/test/java/org/androidannotations/test15/roboguice/TestSampleRoboApplication_.java
```diff
@@ -1,20 +1,54 @@
...
 // CHECKSTYLE:OFF
 public class TestSampleRoboApplication_ extends Application implements TestLifecycleApplication {
+
 	@Override
 	public void onCreate() {
 		super.onCreate();
 
 		RoboGuice.overrideApplicationInjector(this, RoboGuice.newDefaultRoboModule(this), new RobolectricSampleModule());
+		hackHandler();
+
```
WonderCsabo: There are some unnecessary lines here.

AndroidAnnotations/functional-test-1-5/src/test/java/org/androidannotations/test15/roboguice/TestSampleRoboApplication_.java
```diff
@@ -1,20 +1,54 @@
...
+
+	}
+
+	//TODO remove this after upgrading robolectric to 3+
+	private void hackHandler() {
+		final Handler handler;
+		try {
+			Field handlerField = UiThreadExecutor.class.getDeclaredField("HANDLER");
```
WonderCsabo: Can you use FEST reflect for nicer code?

Artyomcool: Good point! Never heard of FEST refclect but it looks very nice for me.

AndroidAnnotations/functional-test-1-5/src/test/java/org/androidannotations/test15/roboguice/TestSampleRoboApplication_.java
Artyomcool: fixed

WonderCsabo:
You have another Checkstyle violation, unused import in `UiThreadExecutorTest`.

I am sorry, but i just realized what is the clean and Robolectric way of hacking the `Handler`. You have to create a custom shadow. Create a `CustomShadowHandler` class and extend from `ShadowHandler`. Override do needed method. Then create the *org.robolectric.Config.properties* file in *src/test/resources* and add this line:
```properties
shadows=my.package.CustomShadowHandler
```
After that you can revert the modifications of the application class. Actually `onCreate` runs before each test. Also you can get rid of the reflection (i hope). Anyway, it is a more cleaner solution i think.

WonderCsabo:
Also it seem you did not edite the correct commits while rebasing. Just move the test class in 7ff35c9. And change the `Handler` hack in bd6d00e. Now all these changes reside in 2726211, adding unrelated changes to its diff.

Artyomcool:
Should I restore commit history from the reflog and rebase changes again?
As for shadows... Maybe it is better to migrate to Robolectric 3? Any decision here will be just dirty workaround, so there is no a good solution. Should we waste our time to implement a better dirty hack?

WonderCsabo:
>Should I restore commit history from the reflog and rebase changes again?

If it is easier, yes. :)

> Maybe it is better to migrate to Robolectric 3? 

The problem is with 3.0 that there is absolutely no ETA on that. We cannot depend on a `SNAPSHOT` dependency unfortunately.

Artyomcool:
> If it is easier, yes. :)
Do you have some easier options? :)

> The problem is with 3.0 that there is absolutely no ETA on that. We cannot depend on a SNAPSHOT dependency unfortunately.
I understand, but it wasn't my primary intention. I mean, I think we should wait for Robolectric 3.0 with ANY suitable solution. If you insist, that own shadow is much better and cleaner - ok, will do the refactoring. But I don't think so. In my opinion there is no good solution here, so there is no point to waste time on this.

WonderCsabo:
I think nope. :)

I think the own shadow is cleaner about the implementation. The hack is still there though after all, you are right. Also the current solution gets that field with reflection and sets the new anonymous class. The proposed solution does not have this perf impact. However, it effects all tests instead of just the `UiThreadExecutor` related ones.

Okay, you can leave it there, but maybe you should call the hack method in `prepareTest`, i think it makes more sense.

Artyomcool:
Ok, I'll do it in this way.

Artyomcool:
Well, I failed to move commits through history. There are too much conflicts.
If you want to try do it yourself, I can revert branch to state before squashing commits. Should I?

WonderCsabo:
@Artyomcool please do!

Artyomcool:
Done. Now it is right before all rebases, including actualizing develop.

WonderCsabo:
It was a pain to rebase this... It is [here](https://github.com/WonderCsabo/androidannotations/commits/cancelUiThreadTasks). Can you look at the commit diffs to check i did not kill anything accidentally?

Artyomcool:
Good job! You only forgot to move robolectric hack into prepareTest, as you asked. I'm just wondering, how you managed to do this? Looks like you just rewrote every commit by hand.

WonderCsabo:
Thanks. I updated.

I did not rewrite all commits by hand. :smile: Well, i soft reset the last commit, and created two commits of them: adding the hack in the right place, and moving the test to the package. Then i squashed these commits to the right commit in two rebase sessions to keep the number of conflicts as small as possible. Then i rebased the whole thing to develop. You are right, i edited all of the commits not just in the conflicts resolution, because i added formtting and removed Checkstyle errors, etc. But i would not say i rewrote them by hand.

Artyomcool:
Well, where we are?

WonderCsabo:
I think it is ready to be merged.

AndroidAnnotations/androidannotations-api/src/main/java/org/androidannotations/api/UiThreadExecutor.java
```diff
@@ -0,0 +1,123 @@
...
+/**
+ * This class provide operations for
+ * {@link org.androidannotations.annotations.UiThread UiThread} tasks.
+ */
+public class UiThreadExecutor {
+
+	private static final Handler HANDLER = new Handler(Looper.getMainLooper()) {
+		@Override
+		public void handleMessage(Message msg) {
+			Runnable callback = msg.getCallback();
```
WonderCsabo: @Artyomcool i almost forgot... I told you that we should run this in a try-finally block, so we make sure we remove the token even if the client code throws a `RuntimeException`.

WonderCsabo:
i have one (hopefully last) comment. Can you please reset this branch to mine, then edit the "add: cleaning up the map at cost of synchronization" commit?

AndroidAnnotations/androidannotations-api/src/main/java/org/androidannotations/api/UiThreadExecutor.java
Artyomcool: After you've got an exception here ui thread is effectively dead. Everything after that has no point.

WonderCsabo: I do not agree. We store everything in static fields, so they are only cleared if the process is killed. If a `RuntimeException` happen in the ui thread, the `Activity` will likely crash, but it does not mean the process is killed. I think.

Artyomcool: Sorry, but you a not correct here. There is only one ui thread per process. When exception happens during looping, there is no way to restore it, even if you catch it in some way. Also, components never crash by themselves, only with a whole process. And it is very-very good thing, actually.

WonderCsabo: No need to be sorry. :) You are right. I may confused this with something
else, and just wrote down without really thinking it though.
I just tested this, and it just goes up to ZygoteInit.main and dies.

Artyomcool: But we could add some parameter to catch and log exceptions. Not sure if we should.

WonderCsabo: Nope, that is not our responsibility.

WonderCsabo:
Rebased and merged as of a9deed6. Thanks @Artyomcool for your great work on this!

WonderCsabo:
I updated the wiki.

PR #1399, opened 2015-05-05
@PreferenceChange more parameter types

### LINKED ISSUE #1394: @PreferenceChange minor parameter related changes
You are supporting a Boolean parameter on the @PreferenceChange annotation (but not mentioned in documentation), and you are not supporting a

     Set<String>  

parameter (but is reported in the documentation, you probably only meant to say that Set is supported) Finally the Long parameter is not supported (and correctly not reported in the documentation).
Maybe while you are at it you can also change @SharedPref to support Double, and also make @PreferenceChange support doubles?

WonderCsabo:
Fixes #1394.


AndroidAnnotations/androidannotations-api/src/main/java/org/androidannotations/annotations/PreferenceChange.java
```diff
@@ -42,8 +42,13 @@
  * <li>A {@link android.preference.Preference Preference} parameter to know
  * which preference was targeted by this event</li>
  * <li>An {@link Object} or {@link java.util.Set Set of strings} or
- * {@link Boolean} or {@link String} to obtain the new value of the
- * {@link android.preference.Preference Preference}</li>
+ * {@link Boolean} or {@link Float} or {@link Integer} or {@link Long} or
```
dodgex: not sure how it is in english, but in german we would comma seperate the words but the last. also personally i would change the order a little bit

> An Object, String, Set of Strings and also Boolean, Float, Integer, Long or thier corresponding primitive types

yDelouis: +1 for separating with commas and for the order.

WonderCsabo:
Thanks for the suggestion on the JavaDoc, updated.

WonderCsabo:
I just realized i used the new parameter validator API, but it is not available in the master branch we are merging this back. I will change this to only be on top of master and not develop.

WonderCsabo:
Now this branch is based on master.

WonderCsabo:
Correcting myself: of course we have to merge back to a hotfix branch, not directly to master.

AndroidAnnotations/androidannotations/src/main/java/org/androidannotations/handler/PreferenceChangeHandler.java
```diff
@@ -79,13 +80,24 @@ protected void processParameters(HasPreferences holder, JMethod listenerMethod,
...
+
+			if (type.equals(CanonicalNameConstants.PREFERENCE)) {
 				call.arg(preferenceParam);
-			} else if (variableElement.asType().toString().equals(CanonicalNameConstants.OBJECT)) {
+			} else if (type.equals(CanonicalNameConstants.OBJECT)) {
 				call.arg(newValueParam);
+			} else if (type.equals(CanonicalNameConstants.INTEGER) || type.equals(int.class.getName()) || //
+					type.equals(CanonicalNameConstants.FLOAT) || type.equals(float.class.getName()) || //
+					type.equals(CanonicalNameConstants.LONG) || type.equals(long.class.getName())) {
+				JClass wrapperClass = type.startsWith("java") ? holder.refClass(type) : JType.parse(holder.codeModel(), type.replace(".class", "")).boxify();
```
yDelouis: This is quite hard to understand. Is there another way to get the wrapper class ?

yDelouis:
I have only one comment. Otherwise, the code is OK.

AndroidAnnotations/androidannotations/src/main/java/org/androidannotations/handler/PreferenceChangeHandler.java
WonderCsabo: The problem is there are two cases:
type is wrapper
type is primitive

In the first case, we can just the holder.refClass. Unfortunetaly in the second case, holder.refClass returns a low letter int class for example. JType.parse is the way to get correct primitive types, but it only expects the name, without the .class. There is no way to get "int" name from "TypeMirror", so we have to remove ".class" manually.

yDelouis: OK then.

WonderCsabo:
Rebased and merged as of 8d63943.

PR #1424, opened 2015-05-21
If ContextCompat is available, use it to resolve drawables

### LINKED ISSUE #1422: Use ContextCompat to retrieve drawables
Given I'm injecting a drawable into my fragment like that:

```
@DrawableRes(R.drawable.bg_button_with_stroke_orange)
protected Drawable backgroundDrawable;
```

Then Android Annotations generates this code:

```
private void init_() {
    Resources resources_ = context_.getResources();
    backgroundDrawable = resources_.getDrawable(drawable.bg_button_with_stroke_orange);
}
```

Which then generates this compiler warning for `targetSdk=22` builds:

```
TheClass_.java:33: warning: [deprecation] getDrawable(int) in Resources has been deprecated
        backgroundDrawable = resources_.getDrawable(drawable.bg_button_with_stroke_orange);
```

Instead, AndroidAnnoations should, given the class [ContextCompat](https://developer.android.com/reference/android/support/v4/content/ContextCompat.html) is available, generate this:

```
private void init_() {
    backgroundDrawable = ContextCompat.getDrawable(context_, drawable.bg_button_with_stroke_orange);
}
```

Benefits:
* Obviously, the compiler warning would be gone
* More importantly, the theme will be applied to the drawable on Lollipop+ devices

ened:
Tries to fix #1422.

### DIFF
--- a/AndroidAnnotations/androidannotations/src/test/java/org/androidannotations/generation/ContextCompatDrawableTest.java
+++ b/AndroidAnnotations/androidannotations/src/test/java/org/androidannotations/generation/ContextCompatDrawableTest.java
@@ -1,0 +1,56 @@
+/**
+ * Copyright (C) 2010-2015 eBusiness Information, Excilys Group
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may not
+ * use this file except in compliance with the License. You may obtain a copy of
+ * the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed To in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+ * License for the specific language governing permissions and limitations under
+ * the License.
+ */
+package org.androidannotations.generation;
+
+import java.io.File;
+
+import org.androidannotations.AndroidAnnotationProcessor;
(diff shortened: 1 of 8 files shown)


WonderCsabo:
Thanks! Can you squash the commits? Also please add two tests in a separate commit: one when ContextCompat is available and one when it is not. Then assert the generated code is correct.  You can check out ActivityIntentFragmentTest for an example.

ened:
@WonderCsabo please take a look.

AndroidAnnotations/androidannotations/src/test/java/org/androidannotations/generation/ContextCompatDrawableTest.java
```diff
@@ -0,0 +1,57 @@
...
+	private static final String DRAWABLE_VIA_SUPPORT_SIGNATURE = ".*myDrawable = ContextCompat\\.getDrawable\\(this, drawable\\.myDrawable\\);.*";
+
+	@Before
+	public void setUp() {
+		addProcessor(AndroidAnnotationProcessor.class);
+	}
+
+	@Test
+	public void activityCompilesWithRegularDrawable() {
+		addManifestProcessorParameter(ContextCompatDrawableTest.class, "AndroidManifestForDrawable.xml");
```
WonderCsabo: You can move this line to `setUp` from both test methods.

WonderCsabo:
Nice PR, thanks!

PR #1427, opened 2015-05-24
Fix/1422 for corner cases.

ened:


WonderCsabo:
I think the first commit is accidental. Also, you have checkstyle violations.

AndroidAnnotations/androidannotations/src/main/java/org/androidannotations/handler/ResHandler.java
```diff
@@ -91,6 +102,10 @@ public void process(Element element, EComponentHolder holder) {
 				methodBody.assign(ref(fieldName), classes().HTML.staticInvoke("fromHtml").arg(invoke(holder.getResourcesRef(), resourceMethodName).arg(idRef)));
 			} else if (element.getAnnotation(DrawableRes.class) != null && hasContextCompatInClasspath()) {
 				methodBody.assign(ref(fieldName), classes().CONTEXT_COMPAT.staticInvoke("getDrawable").arg(holder.getContextRef()).arg(idRef));
+			} else if (element.getAnnotation(DrawableRes.class) != null && hasContextWithGetDrawableMethod() && !hasContextCompatInClasspath()) {
```
WonderCsabo: Maybe if you just create an if `(element.getAnnotation(DrawableRes.class) != null)` then call a method and handle all `@DrawableRes` related things, it would be cleaner and better structured.

AndroidAnnotations/androidannotations/src/main/java/org/androidannotations/handler/ResHandler.java
```diff
@@ -100,4 +115,42 @@ public void process(Element element, EComponentHolder holder) {
 	protected boolean hasContextCompatInClasspath() {
 		return processingEnvironment().getElementUtils().getTypeElement(CanonicalNameConstants.CONTEXT_COMPAT) != null;
 	}
+	
+	protected boolean hasContextWithGetDrawableMethod() {
```
WonderCsabo: This method name is misleading.

AndroidAnnotations/androidannotations/src/main/java/org/androidannotations/handler/ResHandler.java
```diff
@@ -100,4 +115,42 @@ public void process(Element element, EComponentHolder holder) {
...
+		elseBlock.assign(ref, resourcesRef.invoke("getDrawable").arg(idRef));
+	}
+
+
+	private boolean hasGetDrawable(TypeElement type) {
+		if (type == null) {
+			return false;
+		}
+
+		for (Element element : type.getEnclosedElements()) {
```
WonderCsabo: You can use `ElementFilter.methodsIn(type.getEnclosedElements())` to remove the need of the first `if`.

AndroidAnnotations/androidannotations/src/test/java/org/androidannotations/generation/ActivityWithGetDrawableMethod.java
```diff
@@ -0,0 +1,34 @@
...
+import android.app.Activity;
+import android.graphics.drawable.Drawable;
+
+@EActivity
+public class ActivityWithGetDrawableMethod extends Activity {
+
+	@DrawableRes(R.drawable.myDrawable)
+	Drawable myDrawable;
+
+	public Drawable getDrawable(int drawableId) {
```
WonderCsabo: Maybe we should add a comment to clarify the origin of this method.

AndroidAnnotations/androidannotations/src/test/resources/org/androidannotations/generation/Build.java
```diff
@@ -0,0 +1,32 @@
...
+ * License for the specific language governing permissions and limitations under
+ * the License.
+ */
+package android.os;
+
+/**
+ * We have to put this on resources folder because we want to add it to
+ * classpath only on some unit tests methods
+ */
+public class Build {
```
WonderCsabo: Nice! :+1:

WonderCsabo:
There is a problem: `activityCompilesOnMinSdk21WithoutContextCompat` and `activityCompilesOnMinSdkLower21CompileSdkHigher21WithoutContextCompat` fail, because `ActivityIntentBuilder.hasActivityOptionsInFragment` returns `true`, hence generates a call to a version of `startActivityForResult` which does not exist and a compilation error is raised. However i am not really sure why the test can find that kind of `Fragment` class in the classpath at all. :confused:

ened:
Unfortunately I'm not able to reproduce the failing test for `activityCompilesOnMinSdk21WithoutContextCompat` or `activityCompilesOnMinSdkLower21CompileSdkHigher21WithoutContextCompat` as mentioned by you.

AndroidAnnotations/androidannotations/src/main/java/org/androidannotations/handler/ResHandler.java
```diff
@@ -100,4 +110,52 @@ public void process(Element element, EComponentHolder holder) {
...
+		JConditional guardIf = methodBody._if(holder.classes().BUILD_VERSION.staticRef("SDK_INT").gte(holder.classes().BUILD_VERSION_CODES.staticRef("LOLLIPOP")));
+		JBlock ifBlock = guardIf._then();
+		ifBlock.assign(ref, holder.getContextRef().invoke("getDrawable").arg(idRef));
+
+		JBlock elseBlock = guardIf._else();
+		elseBlock.assign(ref, resourcesRef.invoke("getDrawable").arg(idRef));
+	}
+
+	private boolean hasGetDrawable(TypeElement type) {
+		if (type == null) {
```
ened: Should we inline this method?

AndroidAnnotations/androidannotations/src/main/java/org/androidannotations/handler/ResHandler.java
```diff
@@ -100,4 +110,52 @@ public void process(Element element, EComponentHolder holder) {
...
+
+	private void handleDrawableCompatibility(EComponentHolder holder, JFieldRef ref, JBlock methodBody, JFieldRef idRef, String resourceMethodName) {
+		if (hasContextCompatInClasspath()) {
+			methodBody.assign(ref, classes().CONTEXT_COMPAT.staticInvoke("getDrawable").arg(holder.getContextRef()).arg(idRef));
+		} else if (shouldUseContextGetDrawableMethod() && !hasContextCompatInClasspath()) {
+			methodBody.assign(ref, holder.getContextRef().invoke("getDrawable").arg(idRef));
+		} else if (!shouldUseContextGetDrawableMethod() && hasGetDrawableInContext() && !hasContextCompatInClasspath()) {
+			createCallWithIfGuard(holder, ref, methodBody, idRef);
+		} else {
+			methodBody.assign(ref, invoke(holder.getResourcesRef(), resourceMethodName).arg(idRef));
```
ened: @WonderCsabo I omitted a unit test for this case. Do we need one?

AndroidAnnotations/androidannotations/src/test/resources/org/androidannotations/generation/Build.java
```diff
@@ -0,0 +1,32 @@
...
+	
+	public static class  VERSION {
```
ened: @WonderCsabo please check that `checkstyle` warns about extra spaces between "class" keyword and class name?

WonderCsabo:
I ran the test from within Eclipse, did you try that?

AndroidAnnotations/androidannotations/src/main/java/org/androidannotations/handler/ResHandler.java
WonderCsabo: Nope, this is fine.

AndroidAnnotations/androidannotations/src/main/java/org/androidannotations/handler/ResHandler.java
WonderCsabo: It never hurts.

ened:
No, I'm using TextMate & maven for this. Any chance you can help me out for the Eclipse-testing part, please? ;-)

WonderCsabo:
OK, i found out what is the problem with the classpath. The POM defines 4.1.1.4 version of the Android dependency. However, in our [setup](https://github.com/excilys/androidannotations/wiki/InstallingEnvironment#importing-in-eclipse), we also add the `eclipse-dependencies` project to the classpath, which brings android 1.6 dependency, and overrides 4.1.1.4. That is why Eclipse cannot find the newer `Activity` methods. Maven does, because we do not have the `eclipse-dependencies` in Maven builds obviously.

Actually the POM also defined the 1.6 version Android dep, but 1b9f70e changed it to 4.1.1.4. I think it was a bad idea, because in our tests, we simulate we do not have `Fragment` in classpath, then we simulate we have it and [add it manually](https://github.com/excilys/androidannotations/blob/dadd194f6ef15f0580b14c886c45589761e3d05c/AndroidAnnotations/androidannotations/src/test/java/org/androidannotations/generation/ActivityIntentFragmentTest.java#L57). If we have 4.1.1.4 in the test classpath, we cannot simulate the cases without `Fragment`, and adding that manually makes no sense. @yDelouis, you are the author of that commit, i think we should revert the version incrementation. That means we have to change some tests, though.

AndroidAnnotations/androidannotations/src/test/resources/org/androidannotations/generation/Build.java
WonderCsabo: After lot of investigation, It seems the maven checkstyle plugin does not include test resources at all, due to a bug.

WonderCsabo:
@ened i think we should refactor `ResHandler`. We should make it abstract, and the process method should call an abstract `processResource` method and pass it `idRef` etc. Just like `AbstractListenerHandler`. Then create `AnimationResHandler`, `DrawableResHandler` etc, and a `DefaultResHandler`. Then register the correct classes in `AnnotationHandlers`. This way we do not mix the different cases into one class as we do now.

AndroidAnnotations/androidannotations/src/main/java/org/androidannotations/handler/ResHandler.java
```diff
@@ -100,4 +110,52 @@ public void process(Element element, EComponentHolder holder) {
...
+	}
+
+	private boolean hasGetDrawable(TypeElement type) {
+		if (type == null) {
+			return false;
+		}
+
+		List<? extends Element> allMembers = processingEnvironment().getElementUtils().getAllMembers(type);
+		for (Element element : ElementFilter.methodsIn(allMembers)) {
+			ExecutableElement executableElement = (ExecutableElement) element;
```
WonderCsabo: This is not needed since, `ElementFilter.methodsIn` already returns `ExecutableElement`s.

ened:
@WonderCsabo got it.

WonderCsabo:
@ened please squash the current two commits.

AndroidAnnotations/androidannotations/src/test/resources/org/androidannotations/generation/Build.java
WonderCsabo: @ened so can you please just properly format this code? :)

WonderCsabo:
@ened did you read https://github.com/excilys/androidannotations/pull/1427#issuecomment-105169810?

ened:
@WonderCsabo yes, I'm on it, later tonight hopefully.

WonderCsabo:
OK, thanks, just add some reactions to let me know you saw it. :)

ened:
@WonderCsabo added the abstraction. Regarding the "missing" unit test, it turns out we have it already, I think: `activityCompilesWithRegularDrawable`.

AndroidAnnotations/androidannotations/src/test/resources/org/androidannotations/generation/Build.java
ened: http://stackoverflow.com/a/30445636/375209

WonderCsabo:
Thanks for creating an abstraction. However, that should have go in a separate commit. Any chance you can split the commit into two?

ened:
@WonderCsabo Is it a big a deal? I've been re-basing the entire time. Will have to undo that.

WonderCsabo:
Thanks for all your work on this. I will split it myself, then.

WonderCsabo:
Merged and rebased as of 1d36b13. Thanks!

Issue #1444, opened 2015-06-04
@FragmenArg and co. do not work on generic fields


WonderCsabo:
```java
@EFragment
public class GenericFragment<T extends Serializable> extends Fragment {

  @FragmentArg
  T arg;
}
```
This code will invalidate, because *type does not extend Serializable or Parcelable*. However, this is not true actually, and we have to fix our validation.

WonderCsabo:
I think i have a fix.

PR #1445, opened 2015-06-04
Support @FragmentArg on fields with type variable type

WonderCsabo:
Related to #1444.

The old code has lots of things like this:

```java
DeclaredType declaredType = (DeclaredType) element;
typeString = declaredType.asElement().toString();
elementType = annotationHelper.typeElementFromQualifiedName(typeString);
```

So basically serialize the type to a string, the reload the type from it. It is totally unnecessary in most cases i think, but it is maybe an old workaround that i do not know. All tests pass and my projects are working with the new way. @DayS @yDelouis any thoughts whether the old method was intentional or not?

### DIFF
--- a/AndroidAnnotations/androidannotations/src/main/java/org/androidannotations/helper/ValidatorHelper.java
+++ b/AndroidAnnotations/androidannotations/src/main/java/org/androidannotations/helper/ValidatorHelper.java
@@ -1034,47 +1034,20 @@ public class ValidatorHelper {
 	}
 
 	public void canBePutInABundle(Element element, IsValid isValid) {
+		TypeMirror typeMirror = element.asType();
 		String typeString = element.asType().toString();
 
 		if (!isKnownBundleCompatibleType(typeString)) {
 
-			if (element.asType() instanceof DeclaredType) {
-
-				DeclaredType declaredType = (DeclaredType) element.asType();
-				typeString = declaredType.asElement().toString();
-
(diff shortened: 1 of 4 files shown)


DayS:
I have no idea why this part of the code was done like that. It's, indeed, quite odd.

WonderCsabo:
Thanks! I hope it does not break anything...

yDelouis:
Thanks.

AndroidAnnotations/androidannotations/src/main/java/org/androidannotations/helper/ValidatorHelper.java
```diff
@@ -1034,47 +1034,20 @@ public void hasValidLogLevel(Element element, IsValid isValid) {
...
-					isValid.invalidate();
-				}
+				typeMirror = arrayType.getComponentType();
 			}
 
-			if (elementType != null) {
-				TypeElement parcelableType = annotationHelper.typeElementFromQualifiedName(CanonicalNameConstants.PARCELABLE);
-				TypeElement serializableType = annotationHelper.typeElementFromQualifiedName("java.io.Serializable");
-				if (!annotationHelper.isSubtype(elementType, parcelableType) && !annotationHelper.isSubtype(elementType, serializableType)) {
+			if (typeMirror.getKind() != TypeKind.NONE) {
```
WonderCsabo: @yDelouis i wonder we may have to use `TypeKind.ERROR`, too (or instead)?

Issue #1438, opened 2015-06-02
Add form / multipart parameters from method parameters


WonderCsabo:
In `Retrofit` we can do sg like this:
```java
@FormUrlEncoded
@POST("/user/edit")
User updateUser(@Field("first_name") String first, @Field("last_name") String last);

restClient.updateUser("Csaba", "Kozák");
```

However in AA, we have to do this:

```java
@Post("/user/edit")
User updateUser(MultiValueMap<String, String> formData);

MultiValueMap<String, String> formData = new LinkedMultiValueMap<>();
formData.add("first_name", "Csaba");
formData.add("last_name", "Kozák");
restClient.updateUser(formData);
```

It would be nice to add the `@Field` annotation to AA. Actually it can be used for form-urlencoded and multipart requests, too, since both are handled the same way in `FormHttpMessageConverter` (added to the map).

WonderCsabo:
@yDelouis @dodgex thoughts?

dodgex:
while i switched to Retrofit long time ago, i still think that this might be nice to have for those using AA/RestTemplate. :)

yDelouis:
Why not.

WonderCsabo:
Great!

@dodgex if you are missing functionality from our REST client, feel free to open feature requests/contribute.

dodgex:
@WonderCsabo the only feature i miss is the Retrofit support. ;) I had some problems with RestTemplate that pushed me to to the switch.

Issue #900, opened 2014-01-28
@Get arbitrary parameters 


jkbbwr:
It seems that I am unable to set arbitrary query parameters to a @Get declaration

My endpoint looks like 

`http://api.lmiforall.org.uk/api/v1/ashe/estimateHours?soc=2349&coarse=true`

There are a non trivial amount of parameters to this query, is there a declaration I can use to indicate this to the @Rest interface?

vincentjames501:
This would be super useful. I'm surprised more people haven't requested this for more complex querying.  +1.

WonderCsabo:
Unfortunately this is not a priority for us now, and our queue is very-very long. However, we always gladly accept pull requests.

WonderCsabo:
BTW, we should also add `@Query` which could be added to the fields directly.

WonderCsabo:
This could be easily done if we generate calls to the `UriComponentsBuilder` API.

demonito:
When I needed to create @Get request with many dynamic parameteres, and some of them could be duplicated, I had resolved that problem so:

```java
@Rest(rootUrl = "http://example.com:9080/",
        converters = { GsonHttpMessageConverter.class },
        interceptors = { ApiInterceptor.class })
public interface ExampleApi {

   @Get("content/home/product-type/list?{filters}&domain={domain}") //filters is String like "param1=value1&param1=value2&param3=value3"
   ProductTypeListResponse getProductTypeList(int domain, String filters);

 }

public class ApiInterceptor implements ClientHttpRequestInterceptor {

    private static final String TAG = ApiInterceptor.class.getSimpleName();

    @Override
    public ClientHttpResponse intercept(final HttpRequest request, byte[] body, ClientHttpRequestExecution execution) throws IOException {
        final QueryMultiParamsHttpRequest modifiedRequest = new QueryMultiParamsHttpRequest(request);
        return execution.execute(modifiedRequest, body);
    }
}

public class QueryMultiParamsHttpRequest implements HttpRequest {

    private static final String TAG = QueryParametersBuilder.class.getSimpleName();
    private HttpRequest httpRequest;

    public QueryMultiParamsHttpRequest(final HttpRequest httpRequest) {
        this.httpRequest = httpRequest;
    }

    @Override
    public HttpMethod getMethod() {
        return httpRequest.getMethod();
    }

    @Override
    public URI getURI() {
        final URI originalURI = httpRequest.getURI();
[...]

PR #1448, opened 2015-06-05
Add @Field, @Part and @Path

WonderCsabo:
`@PathParam`: allows to add different url variable name than the method parameter name. It can be expanded later to use url encoding.

```java
@Post("/login{id}")
void login(User user, @Path("id") String identifier);
```

`@Field`: allows to add post parameters as method parameters, just like url variables. Form data and multipart data are supported.

```java
@Post("/login")
void login(@Field String username, @Field String password);
```

About multipart: currently, it requires some boilerplate, but much less than before:
```java
@Post("/profilepicture")
@RequiresHeader("Content-Type")
void upload(@Field String name, @Field FileResource picture);

userRestClient.setHeader("Content-Type", MULTIPART_FORM_DATA);
// normal calls
```
It would be better to remove the need for this.

* we could use #967, to add ```@Headers(@Header(headerName = "Content-Type", value = MULTIPART_FORM_DATA ))```
* create a `@Part` annotation which automatically adds the header, and works like `@Field`
* multi-part post flag for `@Post` which adds the header
* add the header if the parameter type is [`Resource`](http://docs.spring.io/spring-android/docs/2.0.0.M1/api/org/springframework/core/io/Resource.html)
* combination of them

I am open to any suggestions.

Implements #1438.

### DIFF
--- a/AndroidAnnotations/androidannotations-rest-spring/rest-spring/src/main/java/org/androidannotations/rest/spring/handler/PathHandler.java
+++ b/AndroidAnnotations/androidannotations-rest-spring/rest-spring/src/main/java/org/androidannotations/rest/spring/handler/PathHandler.java
@@ -1,0 +1,50 @@
+/**
+ * Copyright (C) 2010-2015 eBusiness Information, Excilys Group
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may not
+ * use this file except in compliance with the License. You may obtain a copy of
+ * the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed To in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+ * License for the specific language governing permissions and limitations under
+ * the License.
+ */
+package org.androidannotations.rest.spring.handler;
+
+import javax.lang.model.element.Element;
+
+import org.androidannotations.AndroidAnnotationsEnvironment;
+import org.androidannotations.ElementValidation;
+import org.androidannotations.handler.BaseAnnotationHandler;
+import org.androidannotations.holder.GeneratedClassHolder;
+import org.androidannotations.rest.spring.annotations.Path;
+import org.androidannotations.rest.spring.helper.RestSpringValidatorHelper;
+
+public class PathHandler extends BaseAnnotationHandler<GeneratedClassHolder> {
+
+	private RestSpringValidatorHelper restValidatorHelper;
+
+	public PathHandler(AndroidAnnotationsEnvironment environment) {
+		super(Path.class, environment);
+		restValidatorHelper = new RestSpringValidatorHelper(environment, getTarget());
+	}
+
+	@Override
+	protected void validate(Element element, ElementValidation validation) {
+		restValidatorHelper.enclosingElementHasOneOfRestMethodAnnotations(element, validation);
+
+		if (validation.isValid()) {
+			restValidatorHelper.urlVariableNameExistsInEnclosingAnnotation(element, validation);
+		}
+	}
+
+	@Override
+	public void process(Element element, GeneratedClassHolder holder) throws Exception {
+		// Don't do anything here.
+	}
+
+}
--- a/AndroidAnnotations/androidannotations-rest-spring/rest-spring/src/main/java/org/androidannotations/rest/spring/handler/PathHandler.java
+++ b/AndroidAnnotations/androidannotations-rest-spring/rest-spring/src/main/java/org/androidannotations/rest/spring/handler/PathHandler.java
@@ -35,6 +35,8 @@ public class PathHandler extends BaseAnnotationHandler<GeneratedClassHolder> {
 
 	@Override
 	protected void validate(Element element, ElementValidation validation) {
+		restValidatorHelper.doesNotHaveFieldAnnotation(element, validation);
+
 		restValidatorHelper.enclosingElementHasOneOfRestMethodAnnotations(element, validation);
 
 		if (validation.isValid()) {
--- a/AndroidAnnotations/androidannotations-rest-spring/rest-spring/src/main/java/org/androidannotations/rest/spring/handler/PostHandler.java
+++ b/AndroidAnnotations/androidannotations-rest-spring/rest-spring/src/main/java/org/androidannotations/rest/spring/handler/PostHandler.java
@@ -15,17 +15,33 @@
  */
 package org.androidannotations.rest.spring.handler;
 
+import java.util.Collections;
+
 import javax.lang.model.element.Element;
 import javax.lang.model.element.ExecutableElement;
 
 import org.androidannotations.AndroidAnnotationsEnvironment;
 import org.androidannotations.ElementValidation;
+import org.androidannotations.handler.AnnotationHandler;
+import org.androidannotations.handler.BaseAnnotationHandler;
+import org.androidannotations.handler.HasParameterHandlers;
+import org.androidannotations.holder.GeneratedClassHolder;
+import org.androidannotations.rest.spring.annotations.Field;
 import org.androidannotations.rest.spring.annotations.Post;
+import org.androidannotations.rest.spring.holder.RestHolder;
 
-public class PostHandler extends RestMethodHandler {
+public class PostHandler extends RestMethodHandler implements HasParameterHandlers<RestHolder> {
+
+	private FieldHandler fieldHandler;
 
 	public PostHandler(AndroidAnnotationsEnvironment environment) {
 		super(Post.class, environment);
+		fieldHandler = new FieldHandler(environment);
+	}
+
+	@Override
+	public Iterable<AnnotationHandler> getParameterHandlers() {
+		return Collections.<AnnotationHandler> singleton(fieldHandler);
 	}
 
 	@Override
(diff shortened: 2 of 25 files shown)


WonderCsabo:
@yDelouis @dodgex any feedback about multipart?

yDelouis:
I would add the annotations `@Headers` and `@Header` for that.
I would have named the new annotation `@Post.Param` `@FormParam`.

WonderCsabo:
@yDelouis adding that header again and again is a little bit dumb i think. Maybe we should create `@MultiPart` which behaves exactly like `@Post.Param`, just adds the header automatically. Like retrofit.

yDelouis:
With retrofit, according to the main page, you have to add `@MultiPart` and `@Part` for each param.
What about copying what they did : 
- add `@PathParam` for replacing a placeholder in the url (or keep it as it is, without any annotation).
- add `@Field` for form post request.
- add `@Part` for multipart parts.
- add `@UrlParam` for url params.

`@Field`and `@Part` are incompatible.
And the headers would be added automatically.

Maybe we should wait the plugin system also.

WonderCsabo:
What is the difference beween `@PathParam` and `@UrlParam`?

yDelouis:
`@PathParam` replaces a placeholder, like we do now without annotation.
`@UrlParam` add will add `?name=value` or `&name=value` to the url.

WonderCsabo:
Are you sure `@UrlParam` is needed? It is sill achievable with `@PathParam`. For complex queries, we should implement `@QueryMap` (#900). Actually `@UrlParam` is also nice, but `@QueryMap` is more important i think.

I rename `@Post.Param` to `@Field` and implement `@Part`.

yDelouis:
`@UrlParam` is not necessary but would be convenient.
`@QueryMap` would be nice too.

WonderCsabo:
These two would only work with `@Get`, am i right?

yDelouis:
No, this could be used with `@Post` also. In my company, the webservices are not really REST services and we have to post data with url params.

WonderCsabo:
Well, in that case we could allow it for any HTTP method.

WonderCsabo:
I realized that we already use the "url variable" term for path parameters in the existing code and messages. Maybe we should use `@QueryParam` instead of `@UrlParam`?

WonderCsabo:
This PR is already big. The work for `@Path`, `@Field` and `@Part` is complete. I will add `@Query` and `@QueryParam` if you merge this, @yDelouis.

yDelouis:
Okay for making another PR.

WonderCsabo:
@yDelouis i am starting to think we should require annotations for all method parameters in a REST method. That way the AA code would be more simpler, and also it would be much more clearer for the caller how to use the parameters. It will break the API, but users will see the change and the necessary actions at compile-time.

yDelouis:
You're right. And with the plugin system, we are breaking everything so we shouldn't worry about that.

WonderCsabo:
:+1:

AndroidAnnotations/androidannotations-rest-spring/rest-spring/src/main/java/org/androidannotations/rest/spring/handler/PathHandler.java
```diff
@@ -0,0 +1,55 @@
...
+	private RestSpringValidatorHelper restValidatorHelper;
+
+	public PathHandler(AndroidAnnotationsEnvironment environment) {
+		super(Path.class, environment);
+		restAnnotationHelper = new RestAnnotationHelper(environment, getTarget());
+		restValidatorHelper = new RestSpringValidatorHelper(environment, getTarget());
+	}
+
+	@Override
+	public void process(Element element, GeneratedClassHolder holder) throws Exception {
```
yDelouis: Why did you put `process` before `validate` ? It's a bit strange to me.

WonderCsabo: You are right, i do not know why i did it.

AndroidAnnotations/androidannotations-rest-spring/rest-spring/src/main/java/org/androidannotations/rest/spring/handler/PathHandler.java
```diff
@@ -0,0 +1,55 @@
...
+import org.androidannotations.ElementValidation;
+import org.androidannotations.handler.BaseAnnotationHandler;
+import org.androidannotations.holder.GeneratedClassHolder;
+import org.androidannotations.rest.spring.annotations.Path;
+import org.androidannotations.rest.spring.helper.RestAnnotationHelper;
+import org.androidannotations.rest.spring.helper.RestSpringValidatorHelper;
+
+public class PathHandler extends BaseAnnotationHandler<GeneratedClassHolder> {
+
+	private RestAnnotationHelper restAnnotationHelper;
```
yDelouis: `restAnnotationHelper` is not used.

AndroidAnnotations/androidannotations-rest-spring/rest-spring/src/main/java/org/androidannotations/rest/spring/handler/PostHandler.java
```diff
@@ -42,4 +72,74 @@ protected String getUrlSuffix(Element element) {
...
+	}
+
+	private abstract class AbstractPostParamHandler extends BaseAnnotationHandler<GeneratedClassHolder> {
+
+		AbstractPostParamHandler(Class<?> targetClass, AndroidAnnotationsEnvironment environment) {
+			super(targetClass, environment);
+		}
+
+		@Override
+		public void process(Element element, GeneratedClassHolder holder) throws Exception {
```
yDelouis: Same here. `process` is called after `validate` so it seems more logical to me to put it after `validate`.

AndroidAnnotations/androidannotations-rest-spring/rest-spring/src/main/java/org/androidannotations/rest/spring/helper/RestAnnotationHelper.java
```diff
@@ -246,14 +283,29 @@ public JVar declareHttpHeaders(ExecutableElement executableElement, RestHolder h
...
+				}
 			}
 		}
 		return null;
 	}
 
+	public String getUrlVariableCorrespondingTo(VariableElement parameter) {
+		Path path = parameter.getAnnotation(Path.class);
+		String parameterName;
+		if (path != null && !path.value().equals("")) {
```
yDelouis: I prefer `!path.value.isEmpty()`, but both are okay.

AndroidAnnotations/androidannotations-rest-spring/rest-spring/src/main/java/org/androidannotations/rest/spring/helper/RestAnnotationHelper.java
```diff
@@ -448,4 +500,41 @@ protected String plainName(JClass jClass) {
...
+				postParameterNameToElementName.put(postParameterName, parameter.getSimpleName().toString());
+			}
+		}
+		return postParameterNameToElementName;
+	}
+
+	private String extractPostParameter(VariableElement parameter, Class<? extends Annotation> clazz) {
+		String value = extractAnnotationParameter(parameter, clazz.getCanonicalName(), "value");
+
+		return !value.equals("") ? value : parameter.getSimpleName().toString();
```
yDelouis: Same here.

AndroidAnnotations/androidannotations-rest-spring/rest-spring/src/main/java/org/androidannotations/rest/spring/helper/RestSpringValidatorHelper.java
```diff
@@ -101,6 +107,10 @@ public void enclosingElementHasRestAnnotation(Element element, ElementValidation
 		enclosingElementHasAnnotation(Rest.class, element, valid, error);
 	}
 
+	public void enclosingElementHasOnOfRestMethodAnnotations(Element element, ElementValidation validation) {
```
yDelouis: There is a typo : `HasOneOf`.

AndroidAnnotations/androidannotations-rest-spring/rest-spring/src/main/java/org/androidannotations/rest/spring/helper/RestSpringValidatorHelper.java
```diff
@@ -391,6 +411,80 @@ public void urlVariableNamesExistInParametersAndHasOnlyOneMoreParameter(Executab
...
+			if (validRestMethodAnnotationNames.contains(annotationMirror.getAnnotationType().toString())) {
+				url = restAnnotationHelper.extractAnnotationParameter(element.getEnclosingElement(), annotationMirror.getAnnotationType().toString(), "value");
+				break;
+			}
+		}
+
+		Set<String> urlVariableNames = restAnnotationHelper.extractUrlVariableNames(url);
+
+		String annotationValue = restAnnotationHelper.extractAnnotationValueParameter(element);
+		String expectedUrlVariableName = !annotationValue.equals("") ? annotationValue : element.getSimpleName().toString();
```
yDelouis: Same here.

AndroidAnnotations/androidannotations-rest-spring/rest-spring/src/main/java/org/androidannotations/rest/spring/helper/RestSpringValidatorHelper.java
```diff
@@ -423,4 +517,42 @@ public void hasSetOfHttpMethodReturnType(ExecutableElement element, ElementValid
 			}
 		}
 	}
+
+	public void doesNotHavePathParamAnnotation(Element element, ElementValidation validation) {
```
yDelouis: The method should be named `doesNotHavePathAnnotation` because the annotation is `@Path`.

yDelouis:
It seems good. Could you rebase and address my comments.
When done, you can merge it by yourself to be sure you won't have to rebase twice.
We will have to edit the wiki then.

WonderCsabo:
Wiki edited.

Issue #1302, opened 2014-12-31
@FragmentById does not work on child of fragment.


wintersandroid:
In API 21+ and possibly earlier, i have only seen this when using api 21. The call getFragmentManager().getFragmentById wont get fragments that are children of a fragment, thus when using FragmentById as a child fragment of a EFragment you will get null.

when finding fragment that is a child of a fragment, getChildFragmentManager().findFragmentById()....

dodgex:
yeah you are right. it seems that we have to call `getChildFragmentManager()`... but i also think, that we might need to have an option to decide if the `getFragmentManager()` should be used or if the `getChildFragmentManager()` should be used. as for not all fragment injecting annotations use the `getFragmentManager()` and if we just change that behaviour this would result in a breaking change.

i think to support this we have to add a parameter `@FragmentById` and  `@FragmentByTag` to let the generated code use the correct manager. i think i would call it `useChildFragmentManager` or something like that with a default value of `false` so that it keeps working for existing applications and then the user can set it to `true` to allow the usage of `getChildFragmentManager()`.

@WonderCsabo @yDelouis any better ideas?

WonderCsabo:
Yeah, a boolean flag would be the most flexible. Although `getChildFragmentManager()` is only available in API level 17 and in never revision of the support lib. If it is not available in the actual API Fragment we should just invalidate, but what should happen if it is available in the API, but the app has a lower `miSdkVersion`?

dodgex:
* if not in the api (due to lower target than 17): invalidate
* if native fragment on minsdk < 17 but with target >= 17: invalidate (as it cant be used on every possible target/device)
* if support fragment but old revision: add a note to the javadoc that this feature requires support-v7 rev `> x` and let it fail on compile time if we can't (in)validate here.

did i forgot a case?

WonderCsabo:
I think it would be better invalidating than generating non-compilable code (we can check whether a method exists, like [here](https://github.com/excilys/androidannotations/pull/1297/files#diff-ac9bc12d96eb86e5f2d7586ad35a3a0bR237)).

dodgex:
yeah. where possible invalidate. but imo users should use up to date android support libs anyway

WonderCsabo:
BTW, we also have to validate that the user only set that parameter to `true` if the enclosing class is a `Fragment`.

dodgex:
yeah. i missed to mention that in my first comment but thought about that too.

[...]

PR #1452, opened 2015-06-06
@FragmentByXXX inject child Fragments

WonderCsabo:
Implements #1302.

This is a breaking change, because `@EFragment`s now always use `getFragmentManager()` instead of `getActivity().getFragmentManager()` (or `getChildFragmentManager()` but only if that is set).

### DIFF
--- a/AndroidAnnotations/androidannotations-core/androidannotations-api/src/main/java/org/androidannotations/annotations/FragmentById.java
+++ b/AndroidAnnotations/androidannotations-core/androidannotations-api/src/main/java/org/androidannotations/annotations/FragmentById.java
@@ -42,7 +42,7 @@ import java.lang.annotation.Target;
  *     android:layout_height="match_parent" &gt;
  * 
  *     &lt;fragment
- *         android:id="@+id/myFragment"
+ *         android:id="@+id/<b>myFragment</b>"
  *         android:name="mypackage.MyFragment_"
  *         android:layout_width="match_parent"
  *         android:layout_height="match_parent" /&gt;
@@ -55,9 +55,9 @@ import java.lang.annotation.Target;
  * // all injected fragment will be the same
  * 
  * 	&#064;FragmentById
- * 	public MyFragment myFragment;
+ * 	public MyFragment <b>myFragment</b>;
  * 	
- * 	&#064;FragmentById(R.id.myFragment)
+ * 	&#064;FragmentById(R.id.<b>myFragment</b>)
  * 	public MyFragment myFragment2;
  * }
  * </pre>
--- a/AndroidAnnotations/androidannotations-core/androidannotations-api/src/main/java/org/androidannotations/annotations/FragmentById.java
+++ b/AndroidAnnotations/androidannotations-core/androidannotations-api/src/main/java/org/androidannotations/annotations/FragmentById.java
@@ -64,6 +64,43 @@ import java.lang.annotation.Target;
  * 
  * </blockquote>
  * 
+ * <p>
+ * To use the <code>getChildFragmentManager()</code> to inject the
+ * <code>Fragment</code>, set the {@link #childFragment()} annotation parameter
+ * to <code>true</code>. You can only do this if the annotated field is in a
+ * class which extends <code>android.app.Fragment</code> or
+ * <code>android.support.v4.app.Fragment</code> and the
+ * <code>getChildFragmentManager()</code> method is available.
+ * </p>
+ * 
+ * <blockquote>
+ * 
+ * Example :
+ * 
+ * <pre>
+ * &lt;LinearLayout xmlns:android="http://schemas.android.com/apk/res/android"
+ *     android:layout_width="match_parent"
+ *     android:layout_height="match_parent" &gt;
+ * 
+ *     &lt;fragment
+ *         android:id="@+id/myChildFragment"
+ *         android:name="mypackage.MyChildFragment_"
+ *         android:layout_width="match_parent"
+ *         android:layout_height="match_parent" /&gt;
+ * &lt;/LinearLayout&gt;
+ * 
+ * 
+ * &#064;EFragment(R.layout.parentfragment)
+ * public class MyParentFragment extends Fragment {
+ * 
+ * 	&#064;FragmentById(<b>childFragment = true</b>)
+ * 	MyChildFragment myFragment;
+ * 
+ * }
+ * </pre>
+ * 
+ * </blockquote>
+ * 
  * @see EFragment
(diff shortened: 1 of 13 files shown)


AndroidAnnotations/androidannotations-api/src/main/java/org/androidannotations/annotations/FragmentById.java
```diff
@@ -55,15 +55,51 @@
...
+ *     &lt;fragment
+ *         android:id="@+id/myChildFragment"
+ *         android:name="mypackage.MyChildFragment_"
+ *         android:layout_width="match_parent"
+ *         android:layout_height="match_parent" /&gt;
+ * &lt;/LinearLayout&gt;
+ * 
+ * 
+ * &#064;EFragment(R.layout.fragment)
+ * public class MyActivity extends Fragment {
```
dodgex: MyFragment extends Fragment? ;)

AndroidAnnotations/androidannotations-api/src/main/java/org/androidannotations/annotations/FragmentByTag.java
```diff
@@ -56,14 +56,51 @@
...
+ *         android:id="@+id/myChildFragment"
+ *         android:tag="myChildFragment"
+ *         android:name="mypackage.MyChildFragment_"
+ *         android:layout_width="match_parent"
+ *         android:layout_height="match_parent" /&gt;
+ * &lt;/LinearLayout&gt;
+ * 
+ * 
+ * &#064;EFragment(R.layout.fragment)
+ * public class MyActivity extends Fragment {
```
dodgex: MyFragment

AndroidAnnotations/androidannotations-api/src/main/java/org/androidannotations/annotations/FragmentById.java
WonderCsabo: Good catch. :)

AndroidAnnotations/androidannotations/src/main/java/org/androidannotations/handler/AbstractFragmentByHandler.java
```diff
@@ -0,0 +1,102 @@
...
+		validatorHelper.extendsFragment(element, valid);
+
+		validatorHelper.isNotPrivate(element, valid);
+
+		validatorHelper.childFragmentUsedOnlyIfEnclosingClassIsFragment(element, valid);
+
+		validatorHelper.getChildFragmentManagerMethodIsAvailable(element, valid, androidManifest);
+
+		if (valid.isValid()) {
+			validatorHelper.getChildFragmentManagerMethodIsAvailable(element, valid, androidManifest);
```
dodgex: duplicate of L68?

WonderCsabo: Yeah, thanks.

WonderCsabo:
@wintersandroid @mgod can you check if this is working for you?

mgod:
@WonderCsabo this fixes it for me. It's a bit of a pain as I expect I'll still be supporting APIs 15 and 16 for a while yet, but I can't tell if making this use the regular fragmentManager on API < 17 is useful.

WonderCsabo:
Sorry I am not really think understand what do you mean. We only supports child Fragments from API 17, since that is when the method become avaialbe. And of course you can use the support lib.

mgod:
I can't use ```childFragment=true``` and have it work for an app that is compiled for min SDK version 15 and target SDK version 21, for example. It will correctly tell me that "The 'childFragmentManager' parameter only can be used if the getChildFragmentManager() method is available in Fragment (from API 17)".

Ideally (for me), it would use the getFragmentManager() method when using API < 17 instead of failing to compile. It seems like this would be fairly safe as if I'm using the childFragmentManager in the code as well, it should give me appropriate warnings that I need to handle that separately in API < 17.

WonderCsabo:
So if the device is below API 17, the following `Fragment` layout:
```xml
<LinearLayout >
    <fragment
        android:id="@+id/myChildFragment"
        android:name="com.foo.views.ChildMyFragment_"  />
</LinearLayout>
```
will add `myChildFragment` into the standard `FragmentManager`? If we want to support that case, we have to create a branch:
```java
if (Build.SDK.VERSION_INT < 17) {
  myChildFragment = getFragmentManager().findFragmentById(R.id.myChildFragment);
} else {
  myChildFragment = getChildFragmentManager().findFragmentById(R.id.myChildFragment);
}
```

This is a fairly reasonable request.
However i do not think this should be the default behavior,  it is better to use Support `Fragment`s to provide a unified API in this case. Also, if we generate this silently, it could lead to potential bugs. So my suggestion is making this an explicit decision of the developer and adding a processing parameter `useChildFragmentCallback`, which is `false` by default, and only generates the structure above if set by the client. Otherwise AA will produce an error in case of API < 17 like the current solution does.

@yDelouis @dodgex @mgod WDYT?

mgod:
So it took me about 10 minutes to get this working with API 15 and support ```Fragment```s, so I think it probably makes more sense to keep the implementation and API for AA as is and make the error alert/docs say something more along the lines of:

"The 'childFragmentManager' parameter only can be used if the getChildFragmentManager() method is available in Fragment (from API 17). If you want to support older API versions, you should use android.support.v4.app.Fragment"

WonderCsabo:
Great! I will update the error message, then.

WonderCsabo:
PR updated.

yDelouis:
It's okay. Rebase and you can merge this PR.

WonderCsabo:
Wiki edited.

WonderCsabo:
It [turned out](http://developer.android.com/about/versions/android-4.2.html#NestedFragments) it may be not a good idea...

> You cannot inflate a layout into a fragment when that layout includes a <fragment>. Nested fragments are only supported when added to a fragment dynamically.

@yDelouis @dodgex

mgod:
Huh. I've been using nested fragments in XML layouts for a while now. It looks like apparently the ```android.support.v4.app.FragmentActivity``` does handle this ok or that the AA implementation somehow avoids the issues (I didn't have nested fragments before using AA). I'd certainly appreciate it if this feature stays around as otherwise I'm going to be stuck reimplementing some screens, but I agree that it doesn't look like it should be encouraged.

WonderCsabo:
@mgod I faced the problem when the parent  Fragment is recreated, I got duplicate ID error. http://stackoverflow.com/a/19815266/747412

PR #1453, opened 2015-06-06
Bugfix @OptionsMenuItem field naming

simopete:
`@OptionsMenuItem` field's name can be same as method parameter's name


WonderCsabo:
Nice PR, thanks!

Issue #204, opened 2012-05-25
Provide constructor injection and setter injection (?)


pyricau:
We currently force the users to have at least package private (default) scopes for their fields. Some users might not appreciate that. We could allow constructor injection / setter injection, this way :

```java
@EBean
public class MyBean {
  
  private View myView;

  private MyOtherBean myOtherBean;

  MyBean(@Bean MyOtherBean myOtherBean) {
    this.myOtherBean = myOtherBean;
  }

  void setMyView(@ViewById View myView) {
    this.myView = myView;
  }

}
```

dodgex:
8 months without feedback. I just started to play with method injection for `@Bean`. I have a working  sample. I also tested with multiple values to inject. while that worked, there are issues with the ability to provide a implementation class in the `@Bean.value()` attribute. so i think it is the best to stay with the one injected element per method approach mentioned above.

WonderCsabo:
Sorry, @dodgex i did not saw this issue.

> adding the annotation on the parameter is not a good idea

Why? We already have multiple annotations working just fine, and i just added two more.

dodgex:
@WonderCsabo actualy i'm not sure why i wrote that. iirc there was the discussion if we want to have annotations on a parameter and we decided to have them after this comment.

btw: if you want to see the current implementation you can see it here: https://github.com/dodgex/androidannotations/tree/204_method_injection

WonderCsabo:
@dodgex it is a good idea. :) I see no problems with it. Actually i think it is much more natural then the method annotation. This is used by every DI system.

dodgex:
@WonderCsabo uhm, i'm not sure if i understand your comment. what is more natural? annotation on param?

WonderCsabo:
Yeah, the param, of course. Becase the value actually be injected to the parameter. Just like with `@Receiver.Extra`, or `@Inject` in guice

dodgex:
okay. shall i create something like `@Bean.Impl` as parameter annotation? that way i could implement multiple beans per method injection.

WonderCsabo:
What is the problem with this?
```java
void setBeans(@Bean(MyBeanImpl.class) MyBean bean, @Bean(MyBeanOtherImpl.class) MyBean bean2) {
 this.bean = bean;
 this. bean2 = bean2;
}
```
:confused:

WonderCsabo:
[...]

PR #1457, opened 2015-06-07
allow method injection #204

dodgex:
this is the WIP branch for method injection. in its initial implementation only for `@Bean` but planned to be expanded to most (if not all) injection related annotations.


dodgex:
related to #204

WonderCsabo:
Nice start! Can you please rebase this onto `develop`?

dodgex:
rebased

AndroidAnnotations/androidannotations-core/androidannotations/src/main/java/org/androidannotations/internal/core/handler/BeanHandler.java
```diff
@@ -70,7 +131,38 @@ public void process(Element element, EComponentHolder holder) throws Exception {
 			block = block._if(beanField.eq(_null()))._then();
 		}
 
-		JInvocation getInstance = injectedClass.staticInvoke(EBeanHolder.GET_INSTANCE_METHOD_NAME).arg(holder.getContextRef());
+		JInvocation getInstance = getInstanceInvocation(element, holder, element);
 		block.assign(beanField, getInstance);
 	}
+
+	private void processeMethod(Element element, EComponentHolder holder) {
```
yDelouis: There is a typo in the method name.

yDelouis:
Seems good. However, if you want to extend this to all injected annotation, I think we will need to helpers in order not to have to duplicate this code structure in all Handlers.

AndroidAnnotations/androidannotations-core/androidannotations/src/main/java/org/androidannotations/internal/core/handler/BeanHandler.java
dodgex: typo fixed

dodgex:
i started this as PoC to see what is possible. but yeah, how that we found that it is possible the plan is to extend it to (if possible) all injections

dodgex:
@WonderCsabo @yDelouis i updated the PR. i refactored the code into a helper class and added method injection for `@SystemService`.

feedback is highly appreciated!

AndroidAnnotations/androidannotations-core/androidannotations-test/src/main/java/org/androidannotations/test/ActivityWithServices.java
```diff
@@ -279,4 +279,28 @@
 	@SystemService
 	WindowManager windowManager;
 
+	AppWidgetManager methodInjectedAppWidgetManager;
```
WonderCsabo: These fields should be `private`.

dodgex:
should they? in other tests similar fields are `default` too. thats why i left the visibiliy.

AndroidAnnotations/androidannotations-core/androidannotations-test/src/main/java/org/androidannotations/test/ActivityWithServices.java
dodgex: should they? in other tests similar fields are default too. thats why i left the visibiliy.

WonderCsabo: Never mind. The fields are packapege private because we have to access them from the test class.

I just said it because we want method injections to be able to inject to `private` fields (at least this is on of the reasons). But we still have to access the fields in the test class, so just forget my comment.

dodgex: i just realized that there is no consistency at all in the test classes. the `BeanInjctedActivity` has all fields `public`

WonderCsabo: These fields should be pp, but it is not a big deal.

AndroidAnnotations/androidannotations-core/androidannotations/src/main/java/org/androidannotations/helper/InjectHelper.java
```diff
@@ -0,0 +1,152 @@
...
+			} else {
+				processField(element, holder);
+			}
+		}
+	}
+
+	private void processParam(Element element, EComponentHolder holder) {
+		ExecutableElement method = (ExecutableElement) element.getEnclosingElement();
+		List<? extends VariableElement> parameters = method.getParameters();
+		List<ParamHelper> paramHelper = methodParameterMap.get(method);
```
WonderCsabo: Please use more meaningful names (`paramHelper` and `paramHelper2`).

AndroidAnnotations/androidannotations-core/androidannotations/src/main/java/org/androidannotations/helper/InjectHelper.java
```diff
@@ -0,0 +1,152 @@
...
+			if (param.equals(element)) {
+				JExpression getInstance = handler.getInstanceInvocation(param, holder, param);
+				paramHelper.add(new ParamHelper(getInstance, paramIndex));
+			}
+		}
+
+		if (paramHelper.size() == paramCount) {
+			String methodName = method.getSimpleName().toString();
+
+			Collections.sort(paramHelper, new Comparator<ParamHelper>() {
```
WonderCsabo: Instead of the anonymous `Comparator` class, it would be nicer to have `ParamHelper` implement `Comparable`.

AndroidAnnotations/androidannotations-core/androidannotations/src/main/java/org/androidannotations/helper/InjectHelper.java
```diff
@@ -0,0 +1,152 @@
...
+		}
+	}
+
+	private void processField(Element element, EComponentHolder holder) {
+		String fieldName = element.getSimpleName().toString();
+		JFieldRef beanField = ref(fieldName);
+		JBlock block = holder.getInitBody();
+
+		// TODO check if always required...
+		boolean hasNonConfigurationInstanceAnnotation = element.getAnnotation(NonConfigurationInstance.class) != null;
```
WonderCsabo: Can you explain these lines?

WonderCsabo:
I am wondering, could it be easier to create a common super `AnnotationHandler` class instead of the `InjectHelper` for method injecting annotations?

AndroidAnnotations/androidannotations-core/androidannotations/src/main/java/org/androidannotations/helper/InjectHelper.java
dodgex: this is from the [`BeanHandler`](https://github.com/excilys/androidannotations/blob/develop/AndroidAnnotations/androidannotations-core/androidannotations/src/main/java/org/androidannotations/internal/core/handler/BeanHandler.java#L68) i added the TODO as I'm not yet sure if it is required for non-beans too.

WonderCsabo: I see. So this checks if the field is annotated with both `@NonConfigurationInstance` and `@Bean`, the bean should not be reinjected after configuration changes. I think this only applies to beans, because views, resources etc should not be annotated with `@NonConfigurationInstance`, as the javadoc says.

WonderCsabo:
This should be rebased.

dodgex:
rebased an adressed some of the comments

dodgex:
about the common super class.

first the question is what injections should be supported? e.g. `@Pref` has `CoreBaseAnnotationHandler` as parent class but the `@Bean` and `@SystemService` have `BaseAnnotationHandler`

not sure what the purpose of `CoreBaseAnnotationHandler` beside the fact that has a `CoreValidationHelper`. if it is ok we could use this as a common base or better extend it for that base.

dodgex:
here is a list of annotations that ca be applied to fields. the checked are already supported. 

- [ ] `@App`
- [x] `@Bean`
- [ ] `@Extra`
- [ ] `@FragmentArg`
- [ ] `@FragmentById`
- [ ] `@FragmentByTag`
- [ ] `@FromHtml`
- [ ] `@HttpsClient`
- [ ] `@InjectMenu`
- [ ] `@OptionsMenuItem`
- [ ] `@PreferenceByKey`
- [ ] `@AnimationRes`
- [ ] `@BooleanRes`
- [ ] `@ColorRes`
- [ ] `@ColorStateListRes`
- [ ] `@DimensionPixelOffsetRes`
- [ ] `@DimensionPixelSizeRes`
- [ ] `@DimensionRes`
- [ ] `@DrawableRes`
- [ ] `@HtmlRes`
- [ ] `@IntArrayRes`
- [ ] `@IntegerRes`
- [ ] `@LayoutRes`
- [ ] `@MovieRes`
- [ ] `@StringArrayRes`
- [ ] `@StringRes`
- [ ] `@TextArrayRes`
- [ ] `@TextRes`
- [ ] `@RootContext`
- [ ] `@Pref`
- [x] `@SystemService`
- [ ] `@ViewById`
- [ ] `@ViewsById`

:no_entry_sign: `@InstanceState`
:no_entry_sign: `@NonConfigurationInstance` 

The last two are not related to injection. :)

WonderCsabo:
I think all listed annotations are valid. And do not forget annotations from plugins (`@RestClient`).
@yDelouis do you agree?

dodgex:
so i need to find/create a common level of `BaseAnnotationHandler` where i can implement the method injection and it also needs to be available for plugins? :)

WonderCsabo:
The helper class is also fine for me, if finding a common super class is
not possible.

dodgex:
currently broken, but that should be fixable with https://github.com/excilys/androidannotations/pull/1566

yDelouis:
I'm okay with the listed annotations. If we missed some, we could add them later.
The helper and the parent class are not incompatible. The parent class could use the helper so if we want a handler to extend another class, we can also add this feature to it.

WonderCsabo:
@dodgex this is no longer blocked.

dodgex:
@WonderCsabo i know, i already rebased after the blocker PR was merged. but i had no time yet to continue work on this.

dodgex:
`@ViewById` and `@ViewsById` depend of the release jcodemodel 2.7.12 see https://github.com/phax/jcodemodel/issues/32. i have added 4 TODO notes for the update to 2.7.12 to remove some workarounds i added to have to code compiling right now.

dodgex:
same for `@PreferenceByKey`

WonderCsabo:
Some annotations are missing from plugins (OrmLiteDao, RestService,? )

dodgex:
arg... yeah. i think i have not checked the plugins when generating the above list of annotations todo.

WonderCsabo:
Are you planning to add those to this PR as well?

dodgex:
yeah. if it is ok.

WonderCsabo:
Yes!

dodgex:
added both

AndroidAnnotations/androidannotations-core/androidannotations/src/main/java/org/androidannotations/helper/BundleHelper.java
```diff
@@ -176,10 +176,12 @@ public BundleHelper(AndroidAnnotationsEnvironment environment, TypeMirror elemen
 			if (isTypeParcelable(type)) {
 				methodNameToSave = "put" + "Parcelable";
 				methodNameToRestore = "get" + "Parcelable";
+				restoreCallNeedCastStatement = true;
```
WonderCsabo: Why are these needed?

AndroidAnnotations/androidannotations-core/androidannotations/src/main/java/org/androidannotations/holder/EComponentWithViewSupportHolder.java
```diff
@@ -95,7 +103,8 @@ protected void setOnViewChanged() {
 		onViewChanged = getGeneratedClass().method(PUBLIC, getCodeModel().VOID, "onViewChanged");
 		onViewChanged.annotate(Override.class);
 		onViewChangedBody = onViewChanged.body();
-		onViewChangedBodyBeforeFindViews = onViewChangedBody.blockSimple();
+		onViewChangedBodyBeforeFindViews = onViewChangedBody.blockSimple(); // TODO use blockVirtual
```
WonderCsabo: #1585 will replace these, right?

AndroidAnnotations/androidannotations-core/androidannotations/src/main/java/org/androidannotations/holder/PreferencesDelegate.java
```diff
@@ -71,12 +72,20 @@ public JBlock getAddPreferencesFromResourceBlock() {
...
+		return addPreferencesFromResourceBlockBeforeInject;
+	}
+
 	private void setAddPreferencesFromResourceBlock() {
 		JMethod method = getGeneratedClass().method(PUBLIC, codeModel().VOID, "addPreferencesFromResource");
 		method.annotate(Override.class);
 		JVar preferencesResIdParam = method.param(int.class, "preferencesResId");
 		method.body().invoke(JExpr._super(), "addPreferencesFromResource").arg(preferencesResIdParam);
 		addPreferencesFromResourceBlock = method.body();
+		addPreferencesFromResourceBlockBeforeInject = addPreferencesFromResourceBlock.blockSimple(); // TODO use blockVirtual()
```
WonderCsabo: I implement this in #1585.

AndroidAnnotations/androidannotations-core/androidannotations/src/main/java/org/androidannotations/internal/core/handler/FragmentArgHandler.java
```diff
@@ -40,25 +45,43 @@
...
 
 	@Override
 	public void validate(Element element, ElementValidation validation) {
-		validatorHelper.enclosingElementHasEFragment(element,  validation);
+		injectHelper.validate(FragmentArg.class, element, validation);
 
 		validatorHelper.isNotPrivate(element, validation);
 
-		validatorHelper.canBePutInABundle(element, validation);
+		Element param = element;
```
WonderCsabo: I saw this construct multiple times, maybe it should be moved to a helper method?

AndroidAnnotations/androidannotations-core/androidannotations/src/main/java/org/androidannotations/helper/BundleHelper.java
dodgex: because there will be an incompatible types error otherwise.

AndroidAnnotations/androidannotations-ormlite/ormlite-api/src/main/java/org/androidannotations/ormlite/api/OrmLiteInjectHelper.java
```diff
@@ -0,0 +1,8 @@
+package org.androidannotations.ormlite.api;
+
+/**
+ * @param <T> type to inject
+ */
+public interface OrmLiteInjectHelper<T> {
+	T inject();
+}
```
WonderCsabo: Can you explain why this is needed?

dodgex: i found no better way to get arround the exception.

AndroidAnnotations/androidannotations-core/androidannotations/src/main/java/org/androidannotations/internal/core/handler/FragmentArgHandler.java
dodgex: I'll add a helper method for thos constructs

WonderCsabo: Very well!

WonderCsabo:
You have a [missing license header](https://travis-ci.org/excilys/androidannotations/jobs/84784505#L1181).

AndroidAnnotations/androidannotations-ormlite/ormlite-api/src/main/java/org/androidannotations/ormlite/api/OrmLiteInjectHelper.java
WonderCsabo: What is the problem? This was the same with field injection. We wrapped the inejction in try-catch blocks. Now we have to wrap the setter method call in a try-catch block. Or is that hard to implement?

dodgex: currently that would be hard to implement as the handler itself only returns an `IJExpresson` that "returns" the desired object.

WonderCsabo: What about the following: create a special inject helper which is a subclass of `InjectHelper`. In process(), it calls super.process first. Then if we should handle an exception, it removes the last statement from the block (you may have to use reflection for that), creates the try-catch and adds that black with try-bock containing the removed statement). Because the last statement int the block should be the method invocation or the assignment.

dodgex: i think i found a solution.

dodgex: its gone.

WonderCsabo:
@dodgex you should add new compile-time tests to verify the validation of injection setter methods.

Also you currently allow this:
```
@Extra
void setExtras(@Extra int e) {

}
```
This should be illegal, because we do not know what annotation parameters should be used (from method or from parameter).

dodgex:
validation + compile time test added

WonderCsabo:
@yDelouis i reviewed this PR and @dodgex addressed my comments. But this is a huge PR, so please review it, too.

dodgex:
jcodemodel 2.8.0 has been released and i update the with the changes PR.

dodgex:
rebased

dodgex:
also i removed the [WIP] as this PR should be ready - beside missing review by @yDelouis

AndroidAnnotations/androidannotations-core/androidannotations/src/main/java/org/androidannotations/holder/HasMethodInjection.java
```diff
@@ -0,0 +1,27 @@
...
+ * the License.
+ */
+package org.androidannotations.holder;
+
+import javax.lang.model.element.Element;
+
+import com.helger.jcodemodel.IJExpression;
+import com.helger.jcodemodel.JBlock;
+
+public interface HasMethodInjection<T extends GeneratedClassHolder> {
```
yDelouis: This is strange that this interface is in `holder` package but is implemented by `Handler`. 
`Holders` have injection methods, not `Handlers`. So this interface should be implemented by `Holders`.

AndroidAnnotations/androidannotations-core/androidannotations/src/main/java/org/androidannotations/internal/core/handler/FragmentArgHandler.java
```diff
@@ -67,12 +86,18 @@ public void process(Element element, EFragmentHolder holder) {
 			argKey = fieldName;
 		}
 
-		TypeMirror actualType = codeModelHelper.getActualType(element, holder);
+		DeclaredType enclosingClassType;
+		if (param.getKind() == ElementKind.PARAMETER) {
```
yDelouis: This kind of code is duplicated (in various ways) in several `handlers`. Is there a way to put this code in `InjectHelper` ?

AndroidAnnotations/androidannotations-core/androidannotations/src/main/java/org/androidannotations/helper/InjectHelper.java
```diff
@@ -0,0 +1,200 @@
...
+			Element enclosingElement = element.getEnclosingElement();
+			if (enclosingElement instanceof ExecutableElement) {
+				processParam(element, holder);
+			} else {
+				processField(element, holder);
+			}
+		}
+	}
+
+	private void validateEnclosingElement(Element element, ElementValidation valid) {
```
yDelouis: Instead of passing a value of the enum, `validateEnclosingElement()` could be a method of the interface `HasMethodInjection`.

AndroidAnnotations/androidannotations-ormlite/ormlite/src/main/java/org/androidannotations/ormlite/handler/OrmLiteDaoHandler.java
```diff
@@ -96,7 +116,7 @@ public void process(Element element, EComponentHolder holder) {
 		catchBlock.body() //
 				.staticInvoke(getClasses().LOG, "e") //
 				.arg(holder.getGeneratedClass().name()) //
-				.arg("Could not create DAO " + fieldName) //
+				.arg("Could not create DAO") //
```
yDelouis: Maybe you could pass the `element` as an argument of the method `wrapInjection` so that we can improve the log.

yDelouis:
Seems good. Except for the interfaces `HasMethodInjection` and `HasMethodInjectionWrapper` which are not in the right package and not easy to understand.

WonderCsabo:
@dodgex you also has to rebase this PR, and we merged #1585.

dodgex:
yay :/ #1585 sounds like merge conflicts. but i'm more afraid of #1619. i'll give it a try. :)

AndroidAnnotations/androidannotations-core/androidannotations/src/main/java/org/androidannotations/holder/HasMethodInjection.java
dodgex: i renamed the interface to `MethodInjectionHandler` and moved it into handler package

AndroidAnnotations/androidannotations-core/androidannotations/src/main/java/org/androidannotations/helper/InjectHelper.java
dodgex: moved that method into `MethodInjectionHandler` and implemented the validation in each handler

AndroidAnnotations/androidannotations-ormlite/ormlite/src/main/java/org/androidannotations/ormlite/handler/OrmLiteDaoHandler.java
dodgex: how to log for a method injection with multiple daos?

dodgex:
@yDelouis i moved the `HasMethodInjectionWrapper` to `MethodInjectionHandler.WrapsInjection` i hope that is ok.

dodgex:
adressed the comments and rebased. please don't require me to do such a rebase again :( it took more than 90 minutes :/

AndroidAnnotations/androidannotations-core/androidannotations/src/main/java/org/androidannotations/internal/core/handler/FragmentArgHandler.java
dodgex: i only count two. this one here and the one in [ExtraHandler](https://github.com/excilys/androidannotations/pull/1457/files#diff-dd36efc89c87354095c6bea0c49233a4R120). not sure if it is worth a method in `InjectHelper`.

dodgex:
@WonderCsabo @yDelouis sorry to bother you, but do you have any other feedback*? there is [yet another PR] that requries a rebase (either here or there). I'm ok with rebasing this PR again when the other is merged first. but rebasing a PR like this is no fun. :D

Thanks! :)

*) e.g. a response to [this](https://github.com/excilys/androidannotations/pull/1457#discussion_r44861231) or [this](https://github.com/excilys/androidannotations/pull/1457#discussion_r45061630) comment

WonderCsabo:
@dodgex i think you forget to update all Javadoc. :/

dodgex:
this feature is planned as a "top-secret-hidden-feature". no javadoc or wiki needed... :/

dodgex:
... i'll try to add the javadoc this evening or this weeked. :smile:

WonderCsabo:
Great, thanks.

yDelouis:
I reviewed the changes you made about the interfaces I did not like and some example of how they are used. 
It seems good to me.

WonderCsabo:
@dodgex any update on this one?

dodgex:
@WonderCsabo sadly i did not found time this weekend, I try to add the docs asap.

dodgex:
@WonderCsabo unfortunately i'm currently to busy to get a hand on the documentation part.

ened:
@dodgex Could you spare a few minutes to list down which documentation is needed, e.g. which part and where. Perhaps we can help out.

dodgex:
all modified annotations need a note with small sample that it is possible to use them on methods too

dodgex:
[here](https://github.com/dodgex/androidannotations/blob/5ab361a5faffff5c8350a415478d2c0436214098/AndroidAnnotations/androidannotations-core/androidannotations/src/test/java/org/androidannotations/generation/ActivityWithValidMethodInjection.java)(and in the other tests) you can see the usage.

dodgex:
when one of those annotations is used on the method, it may have one parameter that is a valid type to be injected. or one can have a method with the annotations on multiple parameters (but always ALL given parameters need the annotation - and it is not possible to mix parameters of diffrent annotations)

dodgex:
so what is possible:
```java
@Extra
	void annotationOnMethod(String extra1) {
	}

	void annotationOnParameter(@Extra String extra2) {
	}

	void annotationOnParameters(@Extra String extra3, @Extra String extra4) {
	}
```

but not
```java
	void annotationOnParameter(@Extra String extra2, String noAnnotation) {
	}

	void annotationOnParameters(@Extra String extra3, @Bean MyBean bean) {
	}
```

dodgex:
@WonderCsabo i just realized that i totally missed this PR. i might get some air the next days and i hope to get it done before christmas. could you review the change of the last commit if it is enough or if it is necessary to add more information to the annotation itself. th wiki will get a note that this will be available as of 4.0 but i think the javadoc does not need this info?

WonderCsabo:
The version info is not needed in JD. Your commit is nice. However, on some annotations, we write "use it on fields" etc. So in those cases that has to be changed as well.

dodgex:
yeah, i wanted to check the jdocs if they mention fields and reword that as well. but not add those notes for all annotations.

WonderCsabo:
That is fine.

dodgex:
skipping `@HttpsClient`for now. (see https://github.com/excilys/androidannotations/issues/1640) @WonderCsabo do we still consider removing httpsclient?

dodgex:
not yet edited: all `*Res` annotations

dodgex:
rebased onto develop

dodgex:
and squashed commits + added last javadocs. I hope i didnt miss anything

AndroidAnnotations/androidannotations-core/androidannotations/src/main/java/org/androidannotations/internal/core/handler/BeanHandler.java
```diff
@@ -25,52 +22,58 @@
...
 		AbstractJClass injectedClass = getJClass(annotationHelper.generatedClassQualifiedNameFromQualifiedName(typeQualifiedName));
+		return injectedClass.staticInvoke(EBeanHolder.GET_INSTANCE_METHOD_NAME).arg(holder.getContextRef());
+	}
 
-		String fieldName = element.getSimpleName().toString();
-		JFieldRef beanField = ref(fieldName);
-		JBlock block = holder.getInitBodyInjectionBlock();
-
-		boolean hasNonConfigurationInstanceAnnotation = element.getAnnotation(NonConfigurationInstance.class) != null;
-		if (hasNonConfigurationInstanceAnnotation) {
```
WonderCsabo: @dodgex why this was removed?

AndroidAnnotations/androidannotations-core/androidannotations/src/main/java/org/androidannotations/holder/EComponentWithViewSupportHolder.java
```diff
@@ -126,34 +126,6 @@ public JInvocation findViewById(JFieldRef idRef) {
 		return findViewById;
 	}
 
-	public void processViewById(JFieldRef idRef, AbstractJClass viewClass, JFieldRef fieldRef) {
-		assignFindViewById(idRef, viewClass, fieldRef);
-	}
-
-	public void assignFindViewById(JFieldRef idRef, AbstractJClass viewClass, JFieldRef fieldRef) {
-		String idRefString = idRef.name();
```
WonderCsabo: Where do you handle this now?

AndroidAnnotations/androidannotations-core/androidannotations/src/main/java/org/androidannotations/internal/core/handler/RootContextHandler.java
```diff
@@ -26,55 +24,63 @@
...
 
-		JBlock body = holder.getInitBodyInjectionBlock();
 		IJExpression contextRef = holder.getContextRef();
 
 		if (CanonicalNameConstants.CONTEXT.equals(typeQualifiedName)) {
-			body.assign(ref(fieldName), contextRef);
+			return contextRef;
 		} else {
 			AbstractJClass extendingContextClass = getEnvironment().getJClass(typeQualifiedName);
-			JConditional cond = body._if(holder.getContextRef()._instanceof(extendingContextClass));
```
WonderCsabo: Where do you handle this now?

dodgex: [here](https://github.com/excilys/androidannotations/pull/1457/files#diff-60c076e2f5fcbae18f868083cf6ba72dR78)

WonderCsabo: But all the logging is gone.

AndroidAnnotations/androidannotations-core/androidannotations/src/main/java/org/androidannotations/holder/EComponentWithViewSupportHolder.java
dodgex: the call is replaced with `getFoundViewHolder`

dodgex: the unchecked annotations has moved to `createFoundViewAndIfnotNullBlock`

AndroidAnnotations/androidannotations-core/androidannotations/src/main/java/org/androidannotations/internal/core/handler/RootContextHandler.java
dodgex: Hmm you are right... not sure how to bring the logging back :/ as the generated statement can be used as method param. thats not possible with a logging statement inside. i would need a helper method or something

WonderCsabo: Can't you just add the logging statement before the line of the injecthelper adds the assignmanet, method call? You have to use `process()` for that of course.

dodgex: i think that would generate wired code for methods with multiple parameters. as the elements are not processed in a logical order - similar to the calling order for `@AfterXxx`methods?

if i'm not wrong, this could result in something like this:
```java
if(!(bla instanceof SomeContext)) {
   // logging
}
callInjecteion(xyz); // this is another injection.
if(!(bla2 instanceof AnotherContext)) {
   // logging
}
callContextInjection(bla, bla2);
```

while it wont make sense to inject more than one context, it is possible.

AndroidAnnotations/androidannotations-core/androidannotations/src/main/java/org/androidannotations/internal/core/handler/BeanHandler.java
dodgex: hm, i thought i moved it somewhere else. but it looks like it got lost... need to find a way to support an edge case like this... (similar to the logging :x)

AndroidAnnotations/androidannotations-core/androidannotations/src/main/java/org/androidannotations/internal/core/handler/RootContextHandler.java
dodgex: i think i have an idea... i could introduce something like `injectHelper.prependExpression(expr)` that adds the expression before the injection.

dodgex: but that does not feel clean :(

WonderCsabo: Yeah you are right, there is no defined order how we get the annotations to
process. Let's try it, I think this is not a big deal.

dodgex: for methods `void methodInjectedContext(Activity methodInjectedContext)` and `void multiInjectedContext(@RootContext Service multiInjectedContext, @RootContext Application multiInjectedContext2)` it generates
```java
if (!(context_ instanceof Activity)) {
    Log.w("SomeBean_", (("Due to Context class "+ context_.getClass().getSimpleName())+", the @RootContext Activity won't be populated"));
}
methodInjectedContext(((context_ instanceof Activity)?((Activity) context_):null));
if (!(context_ instanceof Service)) {
    Log.w("SomeBean_", (("Due to Context class "+ context_.getClass().getSimpleName())+", the @RootContext Service won't be populated"));
}
if (!(context_ instanceof Application)) {
    Log.w("SomeBean_", (("Due to Context class "+ context_.getClass().getSimpleName())+", the @RootContext Application won't be populated"));
}
multiInjectedContext(((context_ instanceof Service)?((Service) context_):null), ((context_ instanceof Application)?((Application) context_):null));
```

WonderCsabo: Cool. Let's do this.

dodgex: okay. i added it for now as separate commit. should i squash it with the original commit that updates the `RequestContextHandler`?

WonderCsabo: Nope. Wait until the other thing is resolved. (NonConfiguration)

AndroidAnnotations/androidannotations-core/androidannotations/src/main/java/org/androidannotations/internal/core/handler/BeanHandler.java
dodgex: thats it. :)
```java
this.maintainedDependency = ((maintainedDependency == null)?EmptyDependency_.getInstance_(this):maintainedDependency);
this.recreatedDependency = EmptyDependency_.getInstance_(this);
```

dodgex:
@WonderCsabo as discussed on hangouts the repo for comparing generated code: https://github.com/dodgex/aa_generation_compare

i'll delete that repository after this PR is merged.

WonderCsabo:
Sorry @dodgex for being late in the party, but after looking at the generated code, i realised another approach would be better.

Currently you are passing the same expression to the method parameter and assign it to the field. To make this work, you had to squeeze the whole injection to be an inline statement. I think this resulted and much uglier generated code, and complicated processor code. Moreover, it is possible that in the future we will have injection which cannot be designed to be inline. We already saw that we have problems with this.

I suggest another approach. Let's do the same old generation for both fields and methods as we did before. The idea is to create local variables for the method parameters. Then inject to these local variables as we injected to the fields, there is (almost) no difference. After it is done, in case of methods we just pass the injected local variables. For fields, there is no further action needed.

For example:

```java
{
  String stringExtra;
  if (extras_.containsKey(STRING_EXTRA)) {
    stringExtra = extras_.getString(STRING_EXTRA);
  }
  methodInjection(stringExtra);
}
```
In case of method injection we need a block with parenthesis, to avoid name clashes of the local variable (it is possible the different methods with the same name have same named method parameters, so the best way to avoid clashes is this).

This way we could generate the same code as before, and also keep the generation code almost intact. The main difference is the inject helper must provide the block, and also it must provide the enclosing element (which is not the immediate enclosing element in case of method parameters).

WDYT?

dodgex:
while i hate the extra ammount of work this means, i think i see the point and it makes sense. :/ :)

WonderCsabo:
Yes, we should have figured this out earlier... :(

dodgex:
wow this makes stuff complicated :/

dodgex:
for the extras sample above i now generate the following:
```java
{
    int tmp_intExtra = intExtra;
    if (extras_.containsKey(INT_EXTRA)) {
        tmp_intExtra = extras_.getInt(INT_EXTRA);
    }
    this.intExtra = tmp_intExtra;
}
// or for methods...
{
    methodInjectedExtra(extras_.getString(METHOD_INJECTED_EXTRA));
}
{
    multiInjectedExtra(extras_.getString(MULTI_INJECTED_EXTRA), extras_.getString(MULTI_INJECTED_EXTRA2_EXTRA));
}
```

so we do not override fields... methods have no "previous" state, so the may get the default values returned by the `getXXX` methods of `Bundle`

dodgex:
do we want the generated `{}` block only for fields? if yes, that would make stuff even more complicated...

WonderCsabo:
It would be nice to not have `{}` with fields...

Although the example is wrong. We want this
```java
if (extras_.containsKey(INT_EXTRA)) {
  injectedField = extras_.getInt(INT_EXTRA);
}
// or for methods...
{
    if (extras_.containsKey(INT_EXTRA)) {
        tmp_intExtra = extras_.getInt(INT_EXTRA);
    }
    methodInjectedExtra(tmp_intExtra);
}
```
So the injection is just the same for fields as for methods.

dodgex:
what value should `tmp_intExtra` have for the method injection? an `int` cannot be `null` and by by directly passing the `getXXX` we provide the default value to the method without any need to care about the default value of a certain type.

also for field injection, the way the code currently works this is not possible. if we want it, i have to put some work into rewriting the InjectHelper.

WonderCsabo:
I would implement this in a following way:
* `InjectHelper`'s responsibility to pass the element to be assigned. For fields, it just passes the field. For methods, it creates the temporary variable and passes it. Also it passes the enclosing element which is the enclosing element for fields and the enclosing element of enclosing element in case of parameters.
* The `Handler` then just creates the injection as usual. It will be completely transparent wether it is injecting a method or a field: it is just getting the assign target, the original variable (field or parameter) and the enclosing element.
* The block can be also hidden from the `Handler`. We could create a method in `InjectHelper`, `getInjectionBlock(originalBlock)`. In case of methods, the inject helper will create a `{}` block in `originalBlock` and return that for the same method. In case of fields, it just returns `originalBlock`.

Yes, `InjectHelper` must be changed for this. :/ But after it the `Handler`s job will be much easier. Yes, i know that means lot of changes in the handlers should be reverted and/or changed...

WonderCsabo:
In case of method injection `tmp_intExtra` will have the default value of
the type. We cannot do anything about that. Actually it is the also the
case for fields, but there the client can assign a "default" value, but
this is not a general thing and is not defined in the contract of the
annotations.

dodgex:
PR updated

AndroidAnnotations/androidannotations-core/androidannotations/src/main/java/org/androidannotations/helper/BundleHelper.java
WonderCsabo: Are you sure?

WonderCsabo: I think now with the temporary variable, this is not needed.

AndroidAnnotations/androidannotations-core/androidannotations/src/main/java/org/androidannotations/internal/core/handler/AbstractFragmentByHandler.java
```diff
@@ -17,37 +17,50 @@
...
+		injectHelper = new InjectHelper<>(validatorHelper, this);
 	}
 
 	@Override
 	protected void validate(Element element, ElementValidation validation) {
-		validatorHelper.enclosingElementHasEnhancedViewSupportAnnotation(element, validation);
+		injectHelper.validate(targetClass, element, validation);
 
-		validatorHelper.extendsFragment(element, validation);
+		Element param = element;
```
WonderCsabo: Shouldn't this method receive the real param? I mean a new method in `MethodInjectionHandler`, which will receive the first method parameter if the annotation is on the method, the method parameter if the annotation is on the parameter, and the field is the annotation is on the field.
So it is transparent for the handler itself.

AndroidAnnotations/androidannotations-core/androidannotations/src/main/java/org/androidannotations/internal/core/handler/ContextCompatAwareResHandler.java
```diff
@@ -66,14 +64,11 @@ private boolean hasTargetMethodInContext() {
 		return hasTargetMethod(context, androidRes.getResourceMethodName());
 	}
 
-	private void createCallWithIfGuard(EComponentHolder holder, JFieldRef ref, JBlock methodBody, JFieldRef idRef) {
+	private IJExpression createCallWithIfGuard(EComponentHolder holder, JFieldRef idRef) {
 		JVar resourcesRef = holder.getResourcesRef();
-		JConditional guardIf = methodBody._if(getClasses().BUILD_VERSION.staticRef("SDK_INT").gte(getClasses().BUILD_VERSION_CODES.staticRef(minSdkPlatformName)));
```
WonderCsabo: Can't we move back to the normal if instead of the inline if here? (as no we have the temporary variable)

AndroidAnnotations/androidannotations-core/androidannotations/src/main/java/org/androidannotations/helper/BundleHelper.java
dodgex: yeah, i think you are right. it was required for the inline condition

AndroidAnnotations/androidannotations-core/androidannotations/src/main/java/org/androidannotations/internal/core/handler/ContextCompatAwareResHandler.java
dodgex: we could, but we then would need a way to notify [this line](https://github.com/excilys/androidannotations/pull/1457/files#diff-6eeb2ce2606461aac88c842f34e4a083R79) that we already added the assignment based on the `if` here. it could be easily done by returning `null`. do we want that? (see latest commit)

AndroidAnnotations/androidannotations-core/androidannotations/src/main/java/org/androidannotations/internal/core/handler/AbstractFragmentByHandler.java
dodgex: a new method in an interface that is implemented by ALL related handlers just to not have this if in ONE handler?

WonderCsabo: Hmm. No other handlers do this?

AndroidAnnotations/androidannotations-core/androidannotations/src/main/java/org/androidannotations/internal/core/handler/ContextCompatAwareResHandler.java
WonderCsabo: Why not.

AndroidAnnotations/androidannotations-core/androidannotations/src/main/java/org/androidannotations/internal/core/handler/AbstractFragmentByHandler.java
dodgex: i grepped for `element instanceof ExecutableElement` and `((ExecutableElement) element)` and found none else

AndroidAnnotations/androidannotations-core/androidannotations/src/main/java/org/androidannotations/internal/core/handler/ContextCompatAwareResHandler.java
dodgex: okay

AndroidAnnotations/androidannotations-core/androidannotations/src/main/java/org/androidannotations/internal/core/handler/AbstractFragmentByHandler.java
WonderCsabo: OK.

AndroidAnnotations/androidannotations-core/androidannotations/src/main/java/org/androidannotations/internal/core/handler/ContextCompatAwareResHandler.java
dodgex: PR updated

WonderCsabo:
@dodgex can you update the diff repo?

dodgex:
@WonderCsabo updated

AndroidAnnotations/androidannotations-core/androidannotations/src/main/java/org/androidannotations/internal/core/handler/ExtraHandler.java
```diff
@@ -72,12 +89,34 @@ public void process(Element element, HasExtras holder) {
...
 			String docComment = getProcessingEnvironment().getElementUtils().getDocComment(element);
-			createIntentInjectionMethod(element, (HasIntentBuilder) holder, extraKeyStaticField, fieldName, docComment);
+			createIntentInjectionMethod(param, (HasIntentBuilder) holder, extraKeyStaticField, fieldName, docComment);
 		}
+
+		JMethod injectExtrasMethod = holder.getInjectExtrasMethod();
+		JVar extras = holder.getInjectExtras();
+
+		DeclaredType enclosingClassType;
+		if (param.getKind() == ElementKind.PARAMETER) {
```
WonderCsabo: I thought the injecthelper already passes the `enclosingElement`. You dropped that idea?

AndroidAnnotations/androidannotations-core/androidannotations/src/main/java/org/androidannotations/internal/core/handler/ExtraHandler.java
```diff
@@ -72,12 +89,34 @@ public void process(Element element, HasExtras holder) {
 		}
 
 		JFieldVar extraKeyStaticField = createStaticExtraField(holder, extraKey, fieldName);
-		injectExtraInComponent(element, holder, extraKeyStaticField, fieldName);
-
 		if (holder instanceof HasIntentBuilder) {
 			String docComment = getProcessingEnvironment().getElementUtils().getDocComment(element);
-			createIntentInjectionMethod(element, (HasIntentBuilder) holder, extraKeyStaticField, fieldName, docComment);
+			createIntentInjectionMethod(param, (HasIntentBuilder) holder, extraKeyStaticField, fieldName, docComment);
```
WonderCsabo: Did you handle to create only one intent builder method for the multi injected method?

AndroidAnnotations/androidannotations-core/androidannotations/src/main/java/org/androidannotations/internal/core/handler/ExtraHandler.java
dodgex: it does not provide the enclosing element and never did. the `enclosingElement` here is the enhanced class.

what provided is, is `element` and `param` where `element` is always the annotated element (the field, method or one of the arguments) and `param` is either the annotated element (for field and or arguments in a method with mutliple arguments) or the first/only argument for method injection where the the annotation is on the method.

AndroidAnnotations/androidannotations-core/androidannotations/src/main/java/org/androidannotations/internal/core/handler/ExtraHandler.java
dodgex: i have not found a good solution yet

AndroidAnnotations/androidannotations-core/androidannotations/src/main/java/org/androidannotations/internal/core/handler/ExtraHandler.java
WonderCsabo: OK, i searched through the code and the enclosing element is only queried here in this way, so this is not duplicated.

AndroidAnnotations/androidannotations-core/androidannotations/src/main/java/org/androidannotations/internal/core/handler/ExtraHandler.java
dodgex: i would have to do some check like for the `enclosingElement` and create a separate `createIntentInjectionMethod` for that i think

WonderCsabo: You have to add a map in the handler.

If the element is a method parameter (or an annotated method with multiple params), whose method has multiple parameters, then you have to check if all other parameters in the map or not. If they are already in, you generate the method.
If not, you just add it to the map.

If the element is a field or a method with only one param, then you can generate the extra method.

WonderCsabo: Of course, you have to handle multi injected method extra method generation differently, because you have to create and set all the parameters.

BTW, this applies to `FragmentArg`, too.

dodgex: this is only for methods with the annotation on params. a method with annotation on the method is only allowed to have 1 param.

dodgex: also this would duplicate some of the code already in `InjectHelper` that is used to manage the injection with that method

WonderCsabo: Yep, forgot that part...

WonderCsabo: Hmm. Maybe you can expose some internals then from injecthelper, which could be used by the handler.

The injecthelper could mark the method parameters which were already handled. Then the handler could query this information, and the the same logic i wrote down.

dodgex: currently the map in `InjectHelper` only has a list of the variable references it should use the incovation per method, so this map would not help for the handler. i'd need to store some more data then.

dodgex: but might be possible.

dodgex: are we ok in having an interface to request a callback if all methods are handled instead of polling from the handler?

WonderCsabo: We can do that, yes.

dodgex:
@WonderCsabo as you suggested i refactored the generation of FragmentBuilder and IntentBuilder methods to create one method that contains all arguments for methods with multiple injections. for now it is a dedicated commit, to allow easier reviewing. i split & merge it into the related commits if you reviewed it.

dodgex:
i should do clean builds before push :(

dodgex:
fixed and the diff repo is updated too

dodgex:
updated last commit and diff repo

WonderCsabo:
@dodgex can you update the wiki sometime?

dodgex:
I'll try to do it the next few days.

dodgex:
Checklist for Wiki Updates:



- [ ] `@App`
- [ ] `@Bean`
- [ ] `@Extra`
- [ ] `@FragmentArg`
- [ ] `@FragmentById`
- [ ] `@FragmentByTag`
- [ ] `@HttpsClient`
- [ ] `@InjectMenu`
- [ ] `@OptionsMenuItem`
- [ ] `@PreferenceByKey`
- [ ] `@AnimationRes`
- [ ] `@BooleanRes`
- [ ] `@ColorRes`
- [ ] `@ColorStateListRes`
- [ ] `@DimensionPixelOffsetRes`
- [ ] `@DimensionPixelSizeRes`
- [ ] `@DimensionRes`
- [ ] `@DrawableRes`
- [ ] `@HtmlRes`
- [ ] `@IntArrayRes`
- [ ] `@IntegerRes`
- [ ] `@LayoutRes`
- [ ] `@MovieRes`
- [ ] `@StringArrayRes`
- [ ] `@StringRes`
- [ ] `@TextArrayRes`
- [ ] `@TextRes`
- [ ] `@RootContext`
- [ ] `@Pref`
- [ ] `@SystemService`
- [ ] `@ViewById`
- [ ] `@ViewsById`

dodgex:
wiki on my fork is updated. also i have added the commit `6ba5e68` that updates the year in FAQ to 2016. :)

dodgex:
added `@RestService` and `@OrmLiteDao`

dodgex:
@WonderCsabo I assume you missed that i updated the wiki?

WonderCsabo:
Thanks for pinging. I will check it in a couple of hours.

dodgex:
no problem

WonderCsabo:
I merged your wiki changes. Thanks again for your great work!

Issue #1440, opened 2015-06-03
Utilize ParameterizedTypeReference in REST client


WonderCsabo:
Currently for parameterized response types, we generate the following:
```java
@Get("/")
List<Lesson> getLessons();
```
```java
restTemplate.exchange(rootUrl, HttpMethod.GET, null, List_Lesson.class).getBody();

// ...

public class List_Lesson extends ArrayList<Lesson> {

}
```

This was a clever solution, but unnecessary since Spring for Android 2.0, which has built-in support for parameterized response types:

```java
restTemplate.exchange(rootUrl, HttpMethod.GET, null,
new ParameterizedTypeReference<List<Lesson>>() { }).getBody();
```
I think we should use it it available.

yDelouis:
Implemented.

PR #1467, opened 2015-06-15
Use ParameterizedTypeReference in REST method calls

WonderCsabo:
Generic response class is natively supported in Spring for Android 2.0. This change utilizes that support, if it is available and the reponse class is generic. It falls back to the old implementation otherwise.

Unfortunetaly there is no way to add this to the compile or functional tests.

Related to #1440.

### DIFF
--- a/AndroidAnnotations/androidannotations-rest-spring/rest-spring/src/main/java/org/androidannotations/rest/spring/helper/RestAnnotationHelper.java
+++ b/AndroidAnnotations/androidannotations-rest-spring/rest-spring/src/main/java/org/androidannotations/rest/spring/helper/RestAnnotationHelper.java
@@ -287,12 +287,24 @@ public class RestAnnotationHelper extends TargetAnnotationHelper {
 		JExpression responseClassExpr = nullCastedToNarrowedClass(holder);
 		TypeMirror returnType = executableElement.getReturnType();
 		if (returnType.getKind() != TypeKind.VOID) {
+			if (getElementUtils().getTypeElement(RestSpringClasses.PARAMETERIZED_TYPE_REFERENCE) != null //
+					&& returnType.getKind() == TypeKind.DECLARED && !((DeclaredType) returnType).getTypeArguments().isEmpty() //
+					&& !returnType.toString().startsWith(RestSpringClasses.RESPONSE_ENTITY)) {
+				return createParameterizedTypeReferenceAnonymousSubclassInstance(returnType, holder);
+			}
+
 			JClass responseClass = retrieveResponseClass(returnType, holder);
 			if (responseClass != null) {
 				responseClassExpr = responseClass.dotclass();
 			}
 		}
 		return responseClassExpr;
+	}
+
+	public JExpression createParameterizedTypeReferenceAnonymousSubclassInstance(TypeMirror returnType, RestHolder holder) {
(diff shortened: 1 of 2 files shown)


WonderCsabo:
Unfortunately this does not work for multiple reasons:

* there are too much cases and i did not figure out how to handle them with both `ParameterizedTypeReference` and the old way
* CodeModel import generation is buggy with generic anonymous classes

However i am leaving this branch if anybody wants to continue.

dodgex:
might this https://github.com/excilys/androidannotations/issues/1489 help for the second point?

WonderCsabo:
Yeah, i also thought about that. But i am stranded about the first. :(

dodgex:
could you elaborate on the issue? what cases do you mean? maybe we can find a solution together :)

WonderCsabo:
Thanks for helping out! I really appreciate that.

In [this](https://github.com/excilys/androidannotations/blob/develop/AndroidAnnotations/androidannotations-rest-spring/rest-spring-test/src/main/java/org/androidannotations/rest/spring/test/MyService.java) class, there are lots of cases when arrays and generics are used together. There are two problems with my solution:

* for one of the cases i generate this `List<Event>[].class`, which is invalid
* i think the `ParameterizedTypeReference` is still not used in some cases where it could be

You can check out this branch and see for yourself. You have to upgrade the Spring Android version in the POM to `2.0.0.M2`, and also have to add the Spring repo:
```xml
<repositories>
    <repository>
        <id>spring-milestones</id>
        <name>Spring Milestones</name>
        <url>http://repo.spring.io/milestone</url>
    </repository>
</repositories>
```

dodgex:
I'll try to take a look on it this weekend.

AndroidAnnotations/androidannotations-rest-spring/rest-spring/src/main/java/org/androidannotations/rest/spring/helper/RestAnnotationHelper.java
```diff
@@ -359,7 +375,18 @@ private JClass resolveResponseClass(TypeMirror expectedType, RestHolder holder)
...
 			ArrayType arrayType = (ArrayType) expectedType;
-			return resolveResponseClass(arrayType.getComponentType(), holder).array();
+
+			TypeMirror componentType = arrayType.getComponentType();
+
+			if (componentType.getKind() == TypeKind.DECLARED) {
+				DeclaredType declaredType = (DeclaredType) componentType;
+
+				if (useTypeReference && getElementUtils().getTypeElement(RestSpringClasses.PARAMETERIZED_TYPE_REFERENCE) != null) {
+					return codeModelHelper.typeMirrorToJClass(expectedType);
```
dodgex: shouldnt the `expectedType` be `declaredType`? if not you should remove the unused variable `declaredType`.

dodgex:
i fixed your branch: https://github.com/dodgex/androidannotations/commits/1440_restParameterizedTypeReference

WonderCsabo:
Good job! Let's open a PR. However, we cannot merge that yet, since the `List` importing problem is still here. I really hope switching to `phax/jcodemodel` will solve that.

dodgex:
shall i open the PR from my branch?

WonderCsabo:
Yes, please, so until that other people can comment.

PR #1481, opened 2015-06-24
Add Gitter notifications from Travis

WonderCsabo:
This needs `GITTER_WEBHOOK_URL` variable added in Travis settings.

@yDelouis if you are OK with this i merge and add that.


WonderCsabo:
It turned out that Travis secure environment variable cannot be used for this, we have to use a secure key instead, as explained here: https://github.com/gitterHQ/gitter/issues/196. I fixed that in edf729e7.

Issue #1464, opened 2015-06-10
@WindowFeature and support library


camillebeaumont:
The `@WindowFeature` annotation create `requestWindowFeature(x)` in the `init_()` function even if the activity use the support library (instead of using `supportRequestWindowFeature(x)`)

WonderCsabo:
Good catch! Wanna contribute?

camillebeaumont:
Ok why not. I Find out how to do it.
However, I have an issue. `supportRequestWindowFeature` is from `ActionBarActivity` and not from `FragmentActivity` and unfortunately there is no appcompat-v7 library on Maven central, only the v4 support library. So I can't make a 
Any clues ?

WonderCsabo:
This is not a problem at all. It means we cannot test it in the functional test, but we can in the compile-time test (by adding a fake `ActionBarActivity`). You can generate arbitrary code. :) Check out [this](https://github.com/excilys/androidannotations/blob/develop/AndroidAnnotations/androidannotations/src/test/java/org/androidannotations/generation/ActivityIntentFragmentTest.java) test, which adds `Fragment` to the classpath. The added classes must be in *src/test/resources*, like [this](https://github.com/excilys/androidannotations/tree/develop/AndroidAnnotations/androidannotations/src/test/resources/org/androidannotations/generation).

If you have further questions, do not hesitate to ask..

WonderCsabo:
BTW, you *also*  have to check `AppCompatActivity`, since `ActionBarActivity` is deprecated.

camillebeaumont:
Yes, of course ! I think of this minute ;)

WonderCsabo:
@camillebeaumont did you manage to make progress on this one?

WonderCsabo:
Implemented.

PR #1508, opened 2015-07-31
@WindowFeature uses correct method when using support library

simopete:
Implements #1464


WonderCsabo:
Seems to be OK. However `develop` is currently frozen because of the plugin system work, so we cannot merge PRs currently. After the plugins are done, this has to be rebased and we can merge it.

Issue #1525, opened 2015-08-19
copy javadoc for Activity @Extra and @FragmentArg to thier builder


dodgex:
if it is possible to read javadoc comments from processed elements/fields it would be nice to copy the javadoc for Activity `@Extra` and `@FragmentArg` to the related builder. that allows to have a documented "api".

e.g.
```java
/**
 * [any javadoc]
 */
@Extra("MY_EXTRA_KEY")
protected String myExtra;
```
-> 
```java
/**
 * [any javadoc]
 */
public MyActivity_.IntentBuilder_ myExtra(long myExtra) {
    return super.extra(MY_EXTRA, myExtra);
}
```

dependeing on the possibilities of the code generation extend the javadoc to be valid:
```java
/**
 * [any javadoc]
 *
 * @param myExtra The value to set
 * @return The IntentBuilder to chain calls
 */
```

thoughts?

dodgex:
maybe even generate the javadoc all the time, but add the text from the field ?

```java
/**
 * [optional: copy field javadoc]
 *
 * Intent/Extra Key: "MY_EXTRA_KEY".
 * 
 * @param myExtra The value to set
 * @return The IntentBuilder to chain calls
 */
```

yDelouis:
Shall we count on you to submit a PR as soon as the project is not frozen due to the plugin refacto any more ?

dodgex:
If this enhancement is welcome i'll try to get a PR up. Any eta on the plugin refactoring?

yDelouis:
It's hard to give an ETA for the plugin refacto but I think we are close to the end. All it remains, I think, is to define the API for plugins.

dodgex:
Okay, so I still got some time for this PR. ;)

WonderCsabo:
Before you get too excited: I am not sure the processing API passes the doc
at all. Does it?

dodgex:
@WonderCsabo it passes. 
```java
String comment = processingEnv.getElementUtils().getDocComment(element);
```

dodgex:
is there another annotation beside `@Extra` and `@FragementArg` that creates a method for a builder?

WonderCsabo:
ServiceAction?

dodgex:
`@ServiceAction` uses the same code to generate the methods as `@Extra` so there is the javadoc with the `@param` and `@return` already. only having additional custom javadoc text is not possible there as the setter methods are generated by parameters

WonderCsabo:
[...]

PR #1526, opened 2015-08-19
copy javadoc for Activity @Extra and @FragmentArg to thier Builder

dodgex:
see https://github.com/excilys/androidannotations/issues/1525

### DIFF
--- a/AndroidAnnotations/androidannotations-core/androidannotations/src/main/java/org/androidannotations/helper/APTCodeModelHelper.java
+++ b/AndroidAnnotations/androidannotations-core/androidannotations/src/main/java/org/androidannotations/helper/APTCodeModelHelper.java
@@ -584,4 +584,10 @@ public class APTCodeModelHelper {
 
 		generatedElement.annotate(SuppressWarnings.class).param("value", annotationValue);
 	}
+
+	public void addTrimmedDocComment(JMethod method, String docComment) {
+		if (docComment != null) {
+			method.javadoc().append(docComment.replaceAll("\r", "").trim());
+		}
+	}
 }
--- a/AndroidAnnotations/androidannotations-core/androidannotations/src/main/java/org/androidannotations/internal/core/handler/ExtraHandler.java
+++ b/AndroidAnnotations/androidannotations-core/androidannotations/src/main/java/org/androidannotations/internal/core/handler/ExtraHandler.java
@@ -75,7 +75,8 @@ public class ExtraHandler extends BaseAnnotationHandler<HasExtras> {
 		injectExtraInComponent(element, holder, extraKeyStaticField, fieldName);
 
 		if (holder instanceof HasIntentBuilder) {
-			createIntentInjectionMethod(element, (HasIntentBuilder) holder, extraKeyStaticField, fieldName);
+			String docComment = getProcessingEnvironment().getElementUtils().getDocComment(element);
+			createIntentInjectionMethod(element, (HasIntentBuilder) holder, extraKeyStaticField, fieldName, docComment);
 		}
 	}
 
@@ -105,7 +106,7 @@ public class ExtraHandler extends BaseAnnotationHandler<HasExtras> {
 		ifContainsKey.assign(extraField, restoreMethodCall);
 	}
 
-	private void createIntentInjectionMethod(Element element, HasIntentBuilder holder, JFieldVar extraKeyStaticField, String fieldName) {
-		holder.getIntentBuilder().getPutExtraMethod(element.asType(), fieldName, extraKeyStaticField);
+	private void createIntentInjectionMethod(Element element, HasIntentBuilder holder, JFieldVar extraKeyStaticField, String fieldName, String docComment) {
+		holder.getIntentBuilder().getPutExtraMethod(element.asType(), fieldName, extraKeyStaticField, docComment);
 	}
 }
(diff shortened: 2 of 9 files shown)


AndroidAnnotations/androidannotations/src/main/java/org/androidannotations/handler/FragmentArgHandler.java
```diff
@@ -113,5 +113,12 @@ private void createBuilderInjectionMethod(Element element, EFragmentHolder holde
 		JVar arg = method.param(paramClass, fieldName);
 		method.body().invoke(builderArgsField, bundleHelper.getMethodNameToSave()).arg(argKeyStaticField).arg(arg);
 		method.body()._return(_this());
+
+		String docComment = processingEnv.getElementUtils().getDocComment(element);
+		if (docComment != null) {
+			method.javadoc().append(docComment.replaceAll("\r", "").trim());
```
WonderCsabo: Can you explain this line?

dodgex: i assume you ask for the `.replaceAll("\r", "")`. I removed the `\n` because it breaks the output.

e.g.
```
/**
 * Hello
World
 *
 */
```
instead of
```
/**
 * Hello
 * World
 *
 */
```

WonderCsabo: You mean you removed the `\r`, right? But why is this needed? CodeModel assumes only `\n` line ending and break with `\r\n`?

dodgex: Yes i mean that i removed the `\r`, sorry.

the `.getDocComment(element)` call returns comment with `\r\n` but when i apped that to the javadoc i get the line-break without an `*` in the line (like the first snippet). removing the `\r` resulted in having a correctly formatted javadoc output.

WonderCsabo: I see. Can you move this duplicated code then to `APTCodeModelHelper`?

dodgex: updated

WonderCsabo: I would extract this whole block into one method:

if (docComment != null) {   method.javadoc().append(codeModelHelper.
trimDocComment(docComment));  }

WonderCsabo: I would extract this whole block into one method:
```java
if (docComment != null) {
  method.javadoc().append(docComment.replaceAll("\r", "").trim());
}
```

dodgex: done

WonderCsabo:
Thanks! Seems to be okay. After the plugin system is merged, i will have to ask you to rebase this unfortunately.

dodgex:
I'm just waiting for the moment i can rebase ;-)

WonderCsabo:
Please rebase this onto `develop`.

dodgex:
rebased

WonderCsabo:
It seems you have compilation errors.

dodgex:
yeah... currently fighting with my eclipse to get AA working.

WonderCsabo:
Sorry... We changed to IntelliJ. I did not update the instructions, yet.

dodgex:
the androidannotations-test project is missing project.properties :/

dodgex:
should be fixed

dodgex:
TODO: copy ServiceAction javadoc

WonderCsabo:
@dodgex will you add `@ServiceAction` javadoc gen in this PR?

dodgex:
yeah. it is planned. but i had no time yet. maybe this evening.

WonderCsabo:
OK, i will wait with merging, then.

dodgex:
PR updated with javadoc for `@ServiceAction`. here it is a plain copy of the original doc.

WonderCsabo:
Thanks for handling `@ServiceAction` as well. And it seems it also handles javadoc for parameters correctly:

```java
/**
 * this is a javadoc comment
 *
 * @param param This is something
 */
@ServiceAction
void action(int param) {
}
```
```java
/**
 * this is a javadoc comment
 * 
 *  @param param This is something
 * 
 */
public ServiceWithServiceAction_.IntentBuilder_ action(int param) {
	action(ACTION);
	super.extra(PARAM_EXTRA, param);
	return this;
}
```

Can i ask you to add the param to the test? Also please add a `@return` tag to the `ServiceAction` javadoc as you did for the other two annotations (and also test that)..

Thanks for your work!

WonderCsabo:
Well, the only little problem with `@param` that it is has extra space before it. But i think that is okay, since JavaDoc parsers should be able to handle that (IntelliJ's can).

dodgex:
no idea where the space comes from. but i'm pretty sure that JavaDoc parsers do not care for spaces at all (if not in a `<pre>` block)

dodgex:
PR updated

WonderCsabo:
I know where it comes from. The javadoc which is added [here](https://github.com/dodgex/androidannotations/commit/425c958ed43431b42c73c3b676c97f25b508603b#diff-8978079fb392ef4e876921a75c72d549R126), is interpreted as the main JavaDoc. When codemodel renders it, it adds a space after the starts. But the `@param` already has a space, that is why it is duplicated. But this is not a problem.

WonderCsabo:
After the final review of this PR, i realized there is another last annotations, which can use JavaDoc copy: `@SharedPref` and `@DefaultXXX`. These are not so important, but if you want to contribute that too, go ahead.

WonderCsabo:
Follow up issue for shared prefs: https://github.com/excilys/androidannotations/issues/1551.

Issue #1478, opened 2015-06-24
Use better algorith to find AndroidManifest.xml


WonderCsabo:
`AndroidManifestFinder` tries to find the file, but it still only knows the old Eclipse project structure, where the xml was in the project root. Now, most projects use a different structure.

**Gradle:**
Dummy file: *./build/generated/source/apt{/flavor}{/buildtype}/***
Manifest: *./src/main/AndroidManifest.xml or *./build/manifests/full{/flavor}{/buildtype}/AndroidManifest.xml* or ./build/bundles{/flavor}{/buildtype}/AndroidManifest.xml*

**Maven:**
Dummy file: *./target/generated-sources/***
Manifest: *./src/main/AndroidManifest.xml* (better would be the merged manifest in target, but that is not always present at the time of processing)

**Eclipse:**
Dummy file: *./apt_generated/***
Manifest: *./AndroidManifest.xml*

Instead of the current behavior, we should try to match regex patterns according to the paths above. Then we can easily calculate the manifest path after we know the project root folder found by the regex.
If this algorithm fails to find the manifest, users can still explicitly specify it via the `androidManifestFile` processing option.

PR #1543, opened 2015-09-11
Use a better AndroidManifest.xml finder

WonderCsabo:
Implements #1478.

### DIFF
--- a/AndroidAnnotations/androidannotations-core/androidannotations/src/main/java/org/androidannotations/internal/helper/AndroidManifestFinder.java
+++ b/AndroidAnnotations/androidannotations-core/androidannotations/src/main/java/org/androidannotations/internal/helper/AndroidManifestFinder.java
@@ -22,8 +22,12 @@ import java.io.FileInputStream;
 import java.io.FileNotFoundException;
 import java.io.IOException;
 import java.util.ArrayList;
+import java.util.Arrays;
+import java.util.Collections;
 import java.util.List;
 import java.util.Properties;
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
 
 import javax.lang.model.util.Elements;
 import javax.xml.parsers.DocumentBuilder;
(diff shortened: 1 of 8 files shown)


AndroidAnnotations/androidannotations-core/androidannotations/src/main/java/org/androidannotations/internal/helper/AndroidManifestFinder.java
```diff
@@ -103,43 +112,102 @@ private File findManifestInSpecifiedPath(String androidManifestPath) throws File
...
+	File findManifestInKnownPathsStartingFromGenFolder(String sourcesGenerationFolder) throws FileNotFoundException {
+		Matcher gradleMatcher = GRADLE_GEN_FOLDER.matcher(sourcesGenerationFolder);
+		Matcher mavenMatcher = MAVEN_GEN_FOLDER.matcher(sourcesGenerationFolder);
+		Matcher eclipseMatcher = ECLIPSE_GEN_FOLDER.matcher(sourcesGenerationFolder);
+
+		AndroidManifestFinderStrategy strategy = null;
+
+		File androidManifestFile = null;
+
+		if (gradleMatcher.matches()) {
```
yDelouis: I think you could add a method `matches(String sourceGenerationFolder` in the implementations of `AndroidManifestFinderStrategy` to avoid the `if..else if..else`. That way, you could create a list a `AndroidManifestFinderStrategy`, iterate it to find the right strategy. The constants `XXX_GEN_FOLDER` would be a inner constant of each strategy.

yDelouis:
This is a great improvement. I have only one comment though. Could you address it ?

WonderCsabo:
Done. Is this what you thought?

yDelouis:
Yes, but I wouldn't have passed the variable `sourcesGenerationFolder` in the constructor but as a parameter of the method `applies()`.

WonderCsabo:
@yDelouis then we should also pass it in the `findAndroidManifestFile()`, and calculate the `Matcher` again with the found strategy.  Or save it in the `applies()` method, but i think that would be bad design.
What would you like to do exactly? :)

yDelouis:
You're right.

Issue #1489, opened 2015-07-08
Change to phax/jcodemodel


WonderCsabo:
[This](https://github.com/phax/jcodemodel) is a fork of the original sun project,  and is actively developed. Lots of things are added and fixed, for example we can remove the `JSuperWildcard` class, use [normal](https://github.com/excilys/androidannotations/blob/develop/AndroidAnnotations/androidannotations/src/main/java/org/androidannotations/handler/RoboGuiceHandler.java#L208) `synchronized` here, get rid of reflection, enum contants in annotation parameters, remove ugly code when adding override annotation, better R class imports   etc.

WonderCsabo:
Implemented. Some fixes are still needed from `jcodemodel`, if @phax does them and releases a new version, we should update our dependency and use those enhancements.

PR #1548, opened 2015-09-14
Use phax/jcodemodel instead of com.sun.codemodel

WonderCsabo:
Implements #1489.


yDelouis:
It must be rebased.
I don't like the naming of the class of JCodemodel (`AbstractJClass`, `IJAnnotable`, etc..) but it will help a lot.

Issue #1551, opened 2015-09-15
Copy JavaDoc to @SharedPref class and @DefaultXXX methods


WonderCsabo:
We do not override `@DefaultXXX` annotated methods, but generate a totally new class with new methods. This is why JavaDoc on the original methods cannot be inherited, and needed to be copied to the generated methods.

WonderCsabo:
I assign this issue to @dodgex .

dodgex:
how to handle the user provided `@return` ?

e.g.
```java
	/**
	 * The Age!
	 *
	 * @return the age as int
	 */
	@DefaultInt(42)
	int age();
```
generates:
```java
    /**
     * The Age!
     * 
     *  @return the age as int
     */
    public IntPrefField age() {
        return intField("age", 42);
    }
```
what is actually not correct.

also i'd like to add something like `* Defaults to: 42` but that should be above the `@return` and other tags

dodgex:
i think we should just copy the javadoc and ignore the default value. or we have to add the default value to the top of the javadoc

dodgex:
in my current implementation i add `<p><b>Defaults to</b>: 42</p>` before the given javadoc comment

WonderCsabo:
I think:

```java
  /**
     * The Age!
     * 
     *  @return The PrefField object for chaining calls to work with the preference
     */
    public IntPrefField age() {
```

And for the editor methods:

@return The PrefEditorField object for chaining calls to edit the preference

Otherwise i think your proposal is OK, @dodgex .

dodgex:
the problem is, that depending on compiler/ide configuration the javadoc on the interface has to be valid -> requires `@return`

we would need to remove/replace the `@return` if we want to add that message on the `age()`

WonderCsabo:
You are right. Then stick with your proposal.

WonderCsabo:
I think we should copy the javadoc to the editor methods as well.

dodgex:
oh. i did not realize that there are more methods. :/ i'll add them.

dodgex:
i updated the PR btw :)

WonderCsabo:
Implemented.

PR #1553, opened 2015-09-17
copy @SharedPref javadoc

dodgex:
see https://github.com/excilys/androidannotations/issues/1551

### DIFF
--- a/AndroidAnnotations/androidannotations-core/androidannotations/src/test/java/org/androidannotations/generation/GenerateJavaDocTest.java
+++ b/AndroidAnnotations/androidannotations-core/androidannotations/src/test/java/org/androidannotations/generation/GenerateJavaDocTest.java
@@ -67,4 +67,34 @@ public class GenerateJavaDocTest extends AAProcessorTestHelper {
 		assertGeneratedClassMatches(generatedFile, ".*\\*  @param param this is a param");
 		assertGeneratedClassMatches(generatedFile, ".*\\* @return");
 	}
+
+	@Test
+	public void generateJavaDocForSharedPref() throws IOException {
+		CompileResult result = compileFiles(SharedPrefWithJavaDoc.class);
+		File generatedFile = toGeneratedFile(SharedPrefWithJavaDoc.class);
+
+		assertCompilationSuccessful(result);
+
+		// CHECKSTYLE:OFF
+		String[] fieldDoc = { //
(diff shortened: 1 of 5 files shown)


AndroidAnnotations/androidannotations-core/androidannotations/src/test/java/org/androidannotations/generation/GenerateJavaDocTest.java
```diff
@@ -67,4 +67,17 @@ public void generateJavaDocForServiceAction() throws IOException {
...
 	}
+
+	@Test
+	public void generateJavaDocForSharedPref() throws IOException {
+		CompileResult result = compileFiles(SharedPrefWithJavaDoc.class);
+		File generatedFile = toGeneratedFile(SharedPrefWithJavaDoc.class);
+
+		assertCompilationSuccessful(result);
+
+		assertGeneratedClassMatches(generatedFile, ".*\\* <p><b>Defaults to</b>: 42</p>");
```
WonderCsabo: It would be more correct to use `assertGeneratedClassContains`, since we have to check that both the normal and the editor method has the javadoc.

dodgex: i hope that is what you mean

AndroidAnnotations/androidannotations-core/androidannotations/src/test/java/org/androidannotations/generation/GenerateJavaDocTest.java
```diff
@@ -67,4 +67,28 @@ public void generateJavaDocForServiceAction() throws IOException {
...
+		File generatedFile = toGeneratedFile(SharedPrefWithJavaDoc.class);
+
+		assertCompilationSuccessful(result);
+
+		// CHECKSTYLE:OFF
+		String[] fieldDoc = { //
+				"     * <p><b>Defaults to</b>: 42</p>", //
+				"     * The Age!", //
+				"     *  @return the age as int", //
+				"     *  @deprecated use {@link #ageLong()}", //
```
WonderCsabo: please add the method signatures as well

dodgex: added

WonderCsabo:
Thanks!

PR #1555, opened 2015-09-17
Allow to provide a default value for @DefaultStringSet

dodgex:
based on https://github.com/excilys/androidannotations/pull/1553

This PR allows to provide a default value for a `@DefaultStringSet` annotated method in a `@SharedPref` interface.

Usage:
```java
	@DefaultStringSet({"a", "b", "c"})
	Set<String> setWithDefault();
```

That generates:
```java
   public StringSetPrefField setWithDefault() {
        return stringSetField("setWithDefault", new HashSet<String>(Arrays.asList("a", "b", "c")));
    }
```

### DIFF
--- a/AndroidAnnotations/androidannotations-core/androidannotations-api/src/main/java/org/androidannotations/annotations/sharedpreferences/DefaultStringSet.java
+++ b/AndroidAnnotations/androidannotations-core/androidannotations-api/src/main/java/org/androidannotations/annotations/sharedpreferences/DefaultStringSet.java
@@ -51,6 +51,13 @@ import org.androidannotations.annotations.ResId;
 public @interface DefaultStringSet {
 
 	/**
+	 * The default value of the preference.
+	 *
+	 * @return the default value
+	 */
+	String[] value();
+
+	/**
 	 * The R.string.* field which refers to the key of the preference.
 	 * 
 	 * @return the resource name of the preference key
--- a/AndroidAnnotations/androidannotations-core/androidannotations-test/src/main/java/org/androidannotations/test/prefs/SomePrefs.java
+++ b/AndroidAnnotations/androidannotations-core/androidannotations-test/src/main/java/org/androidannotations/test/prefs/SomePrefs.java
@@ -22,6 +22,7 @@ import org.androidannotations.annotations.sharedpreferences.DefaultFloat;
 import org.androidannotations.annotations.sharedpreferences.DefaultInt;
 import org.androidannotations.annotations.sharedpreferences.DefaultLong;
 import org.androidannotations.annotations.sharedpreferences.DefaultString;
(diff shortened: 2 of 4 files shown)


WonderCsabo:
Seems to be cool, i added two comments. Could you please rebase this onto `develop`?

athkalia:
Hello, 

I am having some trouble with this change, it seems that this default value is mandatory. Unfortunately, I am getting my default value through a 

```
    <MultiSelectListPreference android:key="@string/prefernce_key"
                                   android:title="@string/preference_title"
                                   android:dialogTitle="@string/dialog_title"
                                   android:entries="@array/preference_entries"
                                  android:entryValues="@array/preference_entries_values"
                                   android:defaultValue="@array/preference_default_entry_values"/>
```

so this change doesn't quite suit me. Can't you keep it as optional ? (by the way the comment in the DefaultStringSet still claims that 

```
 * <p>
 * Use on methods in {@link SharedPref} annotated class. The generated method
 * will return an empty {@link java.util.Set} of Strings by default.
 * </p>
```

Cheers

WonderCsabo:
The JavaDoc has to be updated for sure. But we require the default value on any `@DefaultXXX` annotation. You can just specify an empty array, the behavior will be the same before this change. Alternativaly, you can just remove the `@DefaultStringSet` annotation, but that way you cannot specify `keyRes`.

athkalia:
thanks for the response. I noticed that

```
value = {} 
```
does not work. (it does work but generates code that doesn't compile)

value = ''" does work, but it's not quite the same.

In version 3.3.2 
a 
```
new HashSet<String>(0)
```
would get generated,
In version 4.0 with 
```
value = ""
```
 we get a 
```
new HashSet<String>(Arrays.asList(""))
```
and with 
```value = {}
```
we get a 
```new HashSet<String>(Arrays.asList())
```
 (doesn't compile)

WonderCsabo:
Thanks for checking! "" is shortcut for {""} so that is okay. But we should
handle the empty array. @dodgex wanna fix this?

athkalia:
Last food for thought. Can't the @DefaultRes also add support for Set<String> and link the default value to an R.array.* value ?

dodgex:
@WonderCsabo cant fix it before 15.11. and the i have to find time.

dodgex:
i created PR https://github.com/excilys/androidannotations/pull/1624 with a fix and javadoc update

WonderCsabo:
@athkalia the empty array is now fixed thanks to @dodgex.

Mapping Android resources to pref default values is the behavior of the `@DefaultRes` annotation. I think we can extend it by supporting a resource string array as default value of a string set. Please open a new issue for that.

athkalia:
Lovely! Will do, thanks

Issue #1156, opened 2014-09-22
Override @OptionsMenu declared from super classes.


mathieudebrito:
Hi there, It would be cool to be able to override OptionsMenu declared from super classes.
It would be really appreciated especially for stacked fragments.
```java
@EActivity
@OptionsMenu(value=R.menu.my_menu, override=true)
public class MyFragment extends Fragment {
  // ...
}
```
Should produce : 
```java
@Override
public void onCreateOptionsMenu(Menu menu, MenuInflater inflater) {
  menu.clear();
  inflater.inflate(my.package.R.menu.rides_menu, menu);
  super.onCreateOptionsMenu(menu, inflater);
}
```
Best

WonderCsabo:
@yDelouis i started implementing the feature and just realized that @mathieudebrito is talking about two things:

* not adding menus from *parent* classes
* cleaning the menu before inflation (so it removes the menu from *other* classes)

I implemented the second case. Should we add another parameter? I mean `override`, which could only add menus from the leaf class.

However, this starts to be too complicated. :confused:

mathieudebrito:
Hi @WonderCsabo, from my point of view, the main idea is to be able to clean the menu before inflating the one that specifies the cleaning options.

It's true that `@OptionsMenu(value=menuResId, override=true)` is misleading. It should be used to override a menu from the superclass.

In my case, `@OptionsMenu(value=menuResId, cleanBeforeInflate=true)` would clean the menu before inflating the new one. So even if a superclass inflated a menu, only the current leaf should be able to display its menu.

```java
@Override
public void onCreateOptionsMenu(Menu menu, MenuInflater inflater) {
  super.onCreateOptionsMenu(menu, inflater);
  menu.clear();
  inflater.inflate(menuResId, menu);
}
```

WonderCsabo:
@mathieudebrito 

> So even if a superclass inflated a menu, only the current leaf should be able to display its menu.

Let's assume we have the following classes:

```java
@EActivity
@OptionsMenu(R.menu.parent)
public class Parent extends Activity {

}

@EActivity
@OptionsMenu(R.menu.child)
public class Child extends Parent {

}
```
the following method will be generated for `Child`:
```java
    @Override
    public boolean onCreateOptionsMenu(Menu menu) {
        MenuInflater menuInflater = getMenuInflater();
        menuInflater.inflate(R.menu.parent, menu);
        menuInflater.inflate(R.menu.child, menu);
[...]

PR #1556, opened 2015-09-17
[WIP] Add new options for OptionsMenu

shiraji:
See [#1156 (comment)](https://github.com/excilys/androidannotations/issues/1156#issuecomment-69502841)

I'm not fast coder. It may take really long time to fix this issue. If someone wants to do this instead, feel free to do so.

### DIFF
--- a/AndroidAnnotations/androidannotations-core/androidannotations-api/src/main/java/org/androidannotations/annotations/OptionsMenu.java
+++ b/AndroidAnnotations/androidannotations-core/androidannotations-api/src/main/java/org/androidannotations/annotations/OptionsMenu.java
@@ -71,4 +71,18 @@ public @interface OptionsMenu {
 	 * @return the resource names of the menus
 	 */
 	String[] resName() default "";
+
+	/**
+	 * Whether to prevent from displaying parent's menu.
+	 *
+	 * @return <b>true</b> if prevent from displaying parent's menu, <b>false</b> otherwise
+	 */
+	boolean overrideParent() default false;
+
+	/**
+	 * Whether to clean before inflate menu.
+	 *
+	 * @return <b>true</b> if clean before inflate menu, <b>false</b> otherwise
+	 */
+	boolean cleanBeforeInflate() default false;
 }
--- a/AndroidAnnotations/androidannotations-core/androidannotations/src/main/java/org/androidannotations/holder/EActivityHolder.java
+++ b/AndroidAnnotations/androidannotations-core/androidannotations/src/main/java/org/androidannotations/holder/EActivityHolder.java
@@ -65,15 +65,10 @@ public class EActivityHolder extends EComponentWithViewSupportHolder implements 
 	private OnActivityResultDelegate onActivityResultDelegate;
 	private ReceiverRegistrationDelegate<EActivityHolder> receiverRegistrationDelegate;
 	private PreferenceActivityDelegate preferencesHolder;
+	private OnCreateOptionMenuDelegate onCreateOptionMenuDelegate;
 	private JMethod injectExtrasMethod;
 	private JBlock injectExtrasBlock;
 	private JVar injectExtras;
-	private JBlock onCreateOptionsMenuMethodBody;
-	private JVar onCreateOptionsMenuMenuInflaterVar;
-	private JVar onCreateOptionsMenuMenuParam;
-	private JVar onOptionsItemSelectedItem;
-	private JVar onOptionsItemSelectedItemId;
-	private JBlock onOptionsItemSelectedMiddleBlock;
 	private NonConfigurationHolder nonConfigurationHolder;
 	private JBlock initIfNonConfigurationNotNullBlock;
 	private JVar initNonConfigurationInstance;
@@ -107,6 +102,7 @@ public class EActivityHolder extends EComponentWithViewSupportHolder implements 
 		onActivityResultDelegate = new OnActivityResultDelegate(this);
 		receiverRegistrationDelegate = new ReceiverRegistrationDelegate<>(this);
 		preferencesHolder = new PreferenceActivityDelegate(this);
+		onCreateOptionMenuDelegate = new OnCreateOptionMenuDelegate(this);
 		setSetContentView();
 		intentBuilder = new ActivityIntentBuilder(this, androidManifest);
 		intentBuilder.build();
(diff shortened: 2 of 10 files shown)


shiraji:
Can anyone check my code at this point? I want to make sure I went to the right direction. 

*INPUT*

```java
@OptionsMenu(value = {R.menu.menu}, overrideParent = true, cleanBeforeInflate = true)
@EActivity
public class OptionsMenuActivity extends Activity {
}
```
*OUTPUT*

```java
...
    @Override
    public boolean onCreateOptionsMenu(Menu menu) {
        menu.clear();
        MenuInflater menuInflater = getMenuInflater();
        menuInflater.inflate(org.androidannotations.menu.R.menu.menu, menu);
        return true;
    }
...
```

### What I did

* Created new options as you see #1156 
* Refactored onCreateMenu/onCreateItemMenu creation methods to OnCreateOptionMenuDelegate/OnCreateOptionMenuFragmentDelegate

### TODO

* Creating test cases (Especially make sure the combination of optionsMenu, optionsMenuItem, and injecting optionsMenuItem works for new options)

WonderCsabo:
First of all, thanks for contributing to AA. :)

I looked at your code, and the refactors seem reasonable. However you have to move all annotation handlers related to options menu after the `OptionsMenuHandler`, [here](https://github.com/excilys/androidannotations/blob/develop/AndroidAnnotations/androidannotations-core/androidannotations/src/main/java/org/androidannotations/internal/core/CorePlugin.java#L170). Otherwise the `onCreateOptionsMenu` method will be created regardless of the configuration in the `@OptionsItem` configuration. (If i see correctly, only `InjectMenuHandler` is at the wrong place.) Then, null checks will be unnecessary [here](https://github.com/shiraji/androidannotations/commit/25ff6eaf8f93ee9d4efa850c5abbc981d4da396c#diff-f629373b6e81fbd1bfe0dae080816365R46), so you can remove them.

shiraji:
> First of all, thanks for contributing to AA. :)

Your welcome. And thank you for managing this great library!

> Otherwise the onCreateOptionsMenu method will be created regardless of the configuration in the @OptionsItem configuration

You are right. 

I came up with another idea to handle this issue. How about accssing `@OptionMenu` from `InjectMenuHandler` and `OptionsMenuItemHandler`?

```java
	public void process(Element element, HasOptionsMenu holder) {
		String fieldName = element.getSimpleName().toString();
		OptionMenu optionMenu = element.getEnclosingElement().getAnnotation(OptionMenu.class);
		OnCreateOptionMenuDelegate.CreateOptionAnnotationData createOptionAnnotationData;
		if (optionMenu == null) {
			createOptionAnnotationData = new OnCreateOptionMenuDelegate.CreateOptionAnnotationData(false, false);
		} else {
			createOptionAnnotationData = new OnCreateOptionMenuDelegate.CreateOptionAnnotationData(optionMenu.overrideParent(), optionMenu.cleanBeforeInflate());
		}
		JBlock body = holder.getOnCreateOptionsMenuMethodBody(createOptionAnnotationData);
		JVar menuParam = holder.getOnCreateOptionsMenuMenuParam(createOptionAnnotationData);

		body.assign(JExpr._this().ref(fieldName), menuParam);
	}
```

It may violate what you design handlers...but it avoids complexity of registering handlers.

WonderCsabo:
I prefer leaving the current handler architecture intact, so each annotation should be processed by one and only one handler (we already violate that in same cases, but we should try to avoid that).

shiraji:
Make sense. I will move `InjectMenuHandler`.

WonderCsabo:
@shiraji actually, please just fixup the commits, the not add separate fixing commits. You can do this with the git interactive rebase.

WonderCsabo:
Hmm, something is wrong. You removed the null check and now the build throws `NullPointerException`s.

shiraji:
Let me fix the problems. 

Sorry about the commit messages. I was thinking I will rebase all commits for future developers after you check my code. You are saying I should not rebase these commits?

shiraji:
Right...I don't think I can remove null check since I pass null as a parameter at [here](https://github.com/excilys/androidannotations/pull/1556/files#diff-a33f1216ca0b6ed5fabaa0e18579aceaR50) and [here](https://github.com/excilys/androidannotations/pull/1556/files#diff-59733c99950f40af85a0fd3630f9c2abR56)

shiraji:
OK, I fixed problem with NPE. I'm so sorry but I did not run `mvn test` at last commit. I really wanted to complete this pull request. 

I did not add [comments](https://github.com/excilys/androidannotations/pull/1556/files#diff-9af002b00d59db050aaf8fb98df7494eR171) for `OptionsItemHandler` because I only see it creates `onOptionsItemSelected`. Could you point where does `OptionsItemHandler` create `onCreateOptionsMenu`?

shiraji:
I created multiple test cases. Found a huge problem...

```java
@EActivity
@OptionsMenu(R.menu.my_menu1)
public class MyActivity extends Activity {
}

@EActivity
@OptionsMenu(value = R.menu.my_menu2, cleanBeforeInflate = true, overrideParent = true)
public class MySubActivity extends MyActivity {
}
```

Since `@OptionsMenu` on MyActivity handles first, the options always become `false`. 

Well, I have no idea how to fix this problem. I don't want to change the order of processing annotations for only this case...mmm

shiraji:
Instead of using `element`, I used `holder.getAnnotatedElement()`. Again, it may violate what you design for handling ancestors.

Would you mind if I rebase this commit with 25ff6ea ?

WonderCsabo:
I am not sure, to be honest. I think the whole concept of `overrideParent` is a little bit flawed.

@yDelouis @dodgex any suggestions?

WonderCsabo:
@shiraji this should be rebased.

shiraji:
I will rebase the commit as soon as I have time to do it. But, then the question is would you merge this pull request? I agree with your opinion. I think we should wait until more people want this feature.

shiraji:
Squashed

shiraji:
May I close this pull request?

If there are more people want this function, I will recreate pull request.

yDelouis:
I agree with @WonderCsabo, I find these parameters too complicated. And the use case is too specific to be handled by AA.

WonderCsabo:
Okay then, i am closing this. Sorry @shiraji for your extra work on this one. :(

shiraji:
No worry.

PR #1559, opened 2015-09-19
Added parceler `@Extra` and `@InstanceState` support

johncarl81:
This PR offers [Parceler](http://parceler.org) integration which wraps and unwraps beans annotated with `@Parcel` transparently when they are used in `@Extra` or `@InstanceState` aspects.  This is a rework of this previous PR: #977

### DIFF
--- a/AndroidAnnotations/androidannotations-core/androidannotations-test/pom.xml
+++ b/AndroidAnnotations/androidannotations-core/androidannotations-test/pom.xml
@@ -82,6 +82,17 @@
 			<version>140</version>
 			<scope>test</scope>
 		</dependency>
+		<dependency>
+			<groupId>org.parceler</groupId>
+			<artifactId>parceler</artifactId>
+			<version>1.0.3</version>
+			<scope>provided</scope>
+		</dependency>
+		<dependency>
+			<groupId>org.parceler</groupId>
+			<artifactId>parceler-api</artifactId>
+			<version>1.0.3</version>
+		</dependency>
 	</dependencies>
 	<build>
 		<plugins>
--- a/AndroidAnnotations/androidannotations-core/androidannotations-test/pom.xml
+++ b/AndroidAnnotations/androidannotations-core/androidannotations-test/pom.xml
@@ -33,6 +33,7 @@
 		<main.basedir>${project.parent.parent.basedir}</main.basedir>
 		<android-rome-version>1.0.0-r2</android-rome-version>
 		<checkstyle.excludes>**/R.java,**/BuildConfig.java,</checkstyle.excludes>
+		<parceler.version>1.0.3</parceler.version>
 	</properties>
 
 	<dependencies>
@@ -85,13 +86,13 @@
 		<dependency>
 			<groupId>org.parceler</groupId>
 			<artifactId>parceler</artifactId>
-			<version>1.0.3</version>
+			<version>${parceler.version}</version>
 			<scope>provided</scope>
 		</dependency>
 		<dependency>
 			<groupId>org.parceler</groupId>
 			<artifactId>parceler-api</artifactId>
-			<version>1.0.3</version>
+			<version>${parceler.version}</version>
 		</dependency>
 	</dependencies>
 	<build>
--- a/AndroidAnnotations/androidannotations-core/androidannotations-test/src/main/java/org/androidannotations/test/parceler/ParcelerBean.java
+++ b/AndroidAnnotations/androidannotations-core/androidannotations-test/src/main/java/org/androidannotations/test/parceler/ParcelerBean.java
@@ -1,0 +1,44 @@
+/**
+ * Copyright (C) 2010-2015 eBusiness Information, Excilys Group
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may not
+ * use this file except in compliance with the License. You may obtain a copy of
+ * the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed To in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+ * License for the specific language governing permissions and limitations under
+ * the License.
+ */
+package org.androidannotations.test.parceler;
+
+import org.parceler.Parcel;
+import org.parceler.ParcelConstructor;
+
+
+@Parcel
+public class ParcelerBean {
+
+	String name;
+	int age;
+
+	public ParcelerBean() {
+	}
+
+	@ParcelConstructor
+	public ParcelerBean(String name, int age) {
+		this.name = name;
+		this.age = age;
+	}
+
+	public String getName() {
+		return name;
+	}
+
+	public int getAge() {
+		return age;
+	}
+}
--- a/AndroidAnnotations/androidannotations-core/androidannotations/src/main/java/org/androidannotations/helper/AnnotationHelper.java
+++ b/AndroidAnnotations/androidannotations-core/androidannotations/src/main/java/org/androidannotations/helper/AnnotationHelper.java
@@ -22,19 +22,8 @@ import java.lang.annotation.Annotation;
 import java.lang.reflect.InvocationTargetException;
 import java.lang.reflect.Method;
 import java.util.ArrayList;
-import java.util.Collection;
-import java.util.HashMap;
-import java.util.HashSet;
-import java.util.LinkedHashMap;
-import java.util.LinkedHashSet;
-import java.util.LinkedList;
 import java.util.List;
 import java.util.Map;
-import java.util.Set;
-import java.util.SortedMap;
-import java.util.SortedSet;
-import java.util.TreeMap;
-import java.util.TreeSet;
 
 import javax.annotation.processing.ProcessingEnvironment;
 import javax.lang.model.element.AnnotationMirror;
@@ -66,40 +55,6 @@ public class AnnotationHelper {
 
 	public static final String DEFAULT_FIELD_NAME_VALUE = "value";
 	public static final String DEFAULT_FIELD_NAME_RESNAME = "resName";
-	private static final String PARCEL_ANNOTATION = "org.parceler.Parcel";
-	private static final Map<String, Integer> SUPPORTED_PARCEL_TYPES = new HashMap<>();
-
-	static {
-		SUPPORTED_PARCEL_TYPES.put(Collection.class.getName(), 1);
-		SUPPORTED_PARCEL_TYPES.put(List.class.getName(), 1);
-		SUPPORTED_PARCEL_TYPES.put(ArrayList.class.getName(), 1);
-		SUPPORTED_PARCEL_TYPES.put(Set.class.getName(), 1);
-		SUPPORTED_PARCEL_TYPES.put(HashSet.class.getName(), 2);
-		SUPPORTED_PARCEL_TYPES.put(TreeSet.class.getName(), 2);
-		SUPPORTED_PARCEL_TYPES.put(CanonicalNameConstants.SPARSE_ARRAY, 1);
-		SUPPORTED_PARCEL_TYPES.put(Map.class.getName(), 2);
-		SUPPORTED_PARCEL_TYPES.put(HashMap.class.getName(), 2);
-		SUPPORTED_PARCEL_TYPES.put(TreeMap.class.getName(), 2);
-		SUPPORTED_PARCEL_TYPES.put(Integer.class.getName(), 0);
-		SUPPORTED_PARCEL_TYPES.put(Long.class.getName(), 0);
-		SUPPORTED_PARCEL_TYPES.put(Double.class.getName(), 0);
-		SUPPORTED_PARCEL_TYPES.put(Float.class.getName(), 0);
-		SUPPORTED_PARCEL_TYPES.put(Byte.class.getName(), 0);
-		SUPPORTED_PARCEL_TYPES.put(String.class.getName(), 0);
-		SUPPORTED_PARCEL_TYPES.put(Character.class.getName(), 0);
-		SUPPORTED_PARCEL_TYPES.put(Boolean.class.getName(), 0);
-		SUPPORTED_PARCEL_TYPES.put(byte[].class.getName(), 0);
-		SUPPORTED_PARCEL_TYPES.put(char[].class.getName(), 0);
-		SUPPORTED_PARCEL_TYPES.put(boolean[].class.getName(), 0);
-		SUPPORTED_PARCEL_TYPES.put(CanonicalNameConstants.IBINDER, 0);
-		SUPPORTED_PARCEL_TYPES.put(CanonicalNameConstants.BUNDLE, 0);
-		SUPPORTED_PARCEL_TYPES.put(CanonicalNameConstants.SPARSE_BOOLEAN_ARRAY, 0);
-		SUPPORTED_PARCEL_TYPES.put(LinkedList.class.getName(), 1);
-		SUPPORTED_PARCEL_TYPES.put(LinkedHashMap.class.getName(), 2);
-		SUPPORTED_PARCEL_TYPES.put(SortedMap.class.getName(), 2);
-		SUPPORTED_PARCEL_TYPES.put(SortedSet.class.getName(), 1);
-		SUPPORTED_PARCEL_TYPES.put(LinkedHashSet.class.getName(), 1);
-	};
 
 	private static final Logger LOGGER = LoggerFactory.getLogger(AnnotationHelper.class);
 
@@ -140,39 +95,7 @@ public class AnnotationHelper {
 		return getElementUtils().getTypeElement(qualifiedName);
 	}
 
-	public boolean isParcelType(TypeMirror typeMirror) {
-		return isParcelType(typeMirror, true);
-	}
-
-	public boolean isParcelType(TypeMirror typeMirror, boolean root) {
-		if (typeMirror instanceof DeclaredType && getElementUtils().getTypeElement(PARCEL_ANNOTATION) != null) {
-			DeclaredType declaredType = (DeclaredType) typeMirror;
-			TypeElement element = (TypeElement) declaredType.asElement();
-
-			String name = element.getQualifiedName().toString();
-
-			if (isAnnotatedWith(element, PARCEL_ANNOTATION)) {
-				return true;
-			}
-
-			if (SUPPORTED_PARCEL_TYPES.containsKey(name)) {
-				boolean genericsMatch = true;
-
-				Integer genericsSize = SUPPORTED_PARCEL_TYPES.get(name);
-				if (genericsSize == declaredType.getTypeArguments().size()
-						&& (!root || genericsSize > 0)) {
-					for (int i = 0; i < genericsSize; i++) {
-						genericsMatch &= isParcelType(declaredType.getTypeArguments().get(i), false);
-					}
-
-					return genericsMatch;
-				}
-			}
-		}
-		return false;
-	}
-
-	private boolean isAnnotatedWith(Element element, String qualifiedName) {
+	public boolean isAnnotatedWith(Element element, String qualifiedName) {
 		for (AnnotationMirror annotationMirror : element.getAnnotationMirrors()) {
 			if (annotationMirror.getAnnotationType().toString().equals(qualifiedName)) {
 				return true;
--- a/AndroidAnnotations/androidannotations-core/androidannotations/src/main/java/org/androidannotations/helper/AnnotationHelper.java
+++ b/AndroidAnnotations/androidannotations-core/androidannotations/src/main/java/org/androidannotations/helper/AnnotationHelper.java
@@ -22,8 +22,19 @@ import java.lang.annotation.Annotation;
 import java.lang.reflect.InvocationTargetException;
 import java.lang.reflect.Method;
 import java.util.ArrayList;
+import java.util.Collection;
+import java.util.HashMap;
+import java.util.HashSet;
+import java.util.LinkedHashMap;
+import java.util.LinkedHashSet;
+import java.util.LinkedList;
 import java.util.List;
 import java.util.Map;
+import java.util.Set;
+import java.util.SortedMap;
+import java.util.SortedSet;
+import java.util.TreeMap;
+import java.util.TreeSet;
 
 import javax.annotation.processing.ProcessingEnvironment;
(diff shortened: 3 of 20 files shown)


AndroidAnnotations/androidannotations-core/androidannotations-test/src/main/java/org/androidannotations/test/parceler/ParcelerBean.java
```diff
@@ -0,0 +1,44 @@
...
+	}
+
+	public String getName() {
+		return name;
+	}
+
+	public int getAge() {
+		return age;
+	}
+}
```
dodgex: no newline at end of file

johncarl81: Oops.. done.

AndroidAnnotations/androidannotations-core/androidannotations/src/main/java/org/androidannotations/internal/core/helper/IntentBuilder.java
```diff
@@ -135,21 +138,36 @@ public JInvocation getSuperPutExtraInvocation(TypeMirror elementType, JVar extra
...
+				TypeMirror stringType = elementUtils.getTypeElement(STRING).asType();
+				if (!typeUtils.isSubtype(elementType, stringType)) {
+					extraParameterArg = cast(environment.getClasses().SERIALIZABLE, extraParameterArg);
+				}
 			}
 		}
 		return _super().invoke("extra").arg(extraKeyField).arg(extraParameterArg);
 	}
 
+	public boolean isAnnotatedWith(TypeMirror typeMirror, String qualifedName) {
```
dodgex: you already added this method in `AnnotationHelper` any chance to use that one?

johncarl81: I certainly could, but I wasn't sure how to get ahold of AnnotationHelper other than making `isAnnotatedWith` static.  What do you suggest?

AndroidAnnotations/androidannotations-core/androidannotations-test/src/main/java/org/androidannotations/test/parceler/ParcelerBean.java
johncarl81: #1560

AndroidAnnotations/androidannotations-core/androidannotations/src/main/java/org/androidannotations/internal/core/helper/IntentBuilder.java
dodgex: I'd pass the `AnnotationHelper` to the intent builder.

Add it to the constructor, then update the `ActivityIntentBuilder` and `ServiceIntentBuilder` to constructors and thier instantiations. where they are instantiated there is `annotationHelper` field from parent class available.

but i think you should do that in a separate commit before doing the work that adds the parceler support

johncarl81: Got it... done.

WonderCsabo: I think you still forgot to remove the method from here.

AndroidAnnotations/androidannotations-core/androidannotations/src/main/java/org/androidannotations/helper/BundleHelper.java
```diff
@@ -172,6 +173,10 @@ public BundleHelper(AndroidAnnotationsEnvironment environment, TypeMirror elemen
 			if (isTypeParcelable(type)) {
 				methodNameToSave = "put" + "Parcelable";
 				methodNameToRestore = "get" + "Parcelable";
+			} else if (annotationHelper.isAnnotatedWith(type, "org.parceler.Parcel")) {
```
WonderCsabo: `Parcels.wrap/unwrap` method [can handle](http://stackoverflow.com/questions/29674199/use-android-annotations-with-parceler/29680080#comment47641892_29691052) `List<AnnotatedWithParcel>` AFAIK, or even more. We should also allow those.

AndroidAnnotations/androidannotations-core/androidannotations/src/main/java/org/androidannotations/internal/core/handler/FragmentArgHandler.java
```diff
@@ -106,7 +106,10 @@ private void createBuilderInjectionMethod(Element element, EFragmentHolder holde
 		JClass paramClass = codeModelHelper.typeMirrorToJClass(type);
 
 		JMethod method = builderClass.method(PUBLIC, holder.narrow(builderClass), fieldName);
-		JVar arg = method.param(paramClass, fieldName);
+		JExpression arg = method.param(paramClass, fieldName);
+		if (bundleHelper.isParcelerBean()) {
```
WonderCsabo: We should add this to `BundleHelper` in some way, to remove code duplication.

WonderCsabo:
Sorry, but i am not sure the refactor with passing the `AnnotationHelper` instance around is needed. You could just create a new `AnnoationHelper` in the intent builder constructor.

```java
annotationHelper = new AnnotationHelper(environment);
```

AndroidAnnotations/androidannotations-core/androidannotations/src/main/java/org/androidannotations/internal/core/handler/FragmentArgHandler.java
WonderCsabo: Maybe we could just create the whole expression with `BundleHelper` just like when we restore, instead of only get the method name, and leave the caller (AnnotationHandler) to create the whole saving expression.

AndroidAnnotations/androidannotations-core/androidannotations/src/main/java/org/androidannotations/internal/core/helper/IntentBuilder.java
johncarl81: It got missed in the rebase.. should be removed now.

AndroidAnnotations/androidannotations-core/androidannotations/src/main/java/org/androidannotations/internal/core/handler/FragmentArgHandler.java
johncarl81: Sounds good... done.

AndroidAnnotations/androidannotations-core/androidannotations/src/main/java/org/androidannotations/helper/BundleHelper.java
johncarl81: Parceler can also handle boxed primitives in collections:
`List<Integer>`, `Map<Integer, String>`
A mix:
`Map<String, AnnotatedWithParcel>`
And nested collections:
`Map<List<String>, List<AnnotatedWithParcel>>`
Would you like these cases handled as well?

yDelouis:
Could you please update the javadoc of the annotations `@Extra`, `@FragmentArg` and `@SavedInstanceState` to specify that these annotations are compatible with classes annotated with `@Parceler` ?

yDelouis:
And I think you forgot to handle inner `Extra` annotations like `@OnActivityResult.Extra`. You have to modify the class `ExtraParameterHandler`.

AndroidAnnotations/androidannotations-core/androidannotations/src/main/java/org/androidannotations/helper/BundleHelper.java
WonderCsabo: If you can add it, it would be great. However we should also check that we have `Parceler` on the classpath (when the user adds a `List` of primitives for examples, we cannot be sure because we are not checking any annotation).

WonderCsabo:
@yDelouis Those extra annotations are already handled by the `BundleHelper`.

WonderCsabo:
@johncarl81 this branch also has to be rebased.

yDelouis:
@WonderCsabo, you're right, sorry.

AndroidAnnotations/androidannotations-core/androidannotations/src/main/java/org/androidannotations/helper/BundleHelper.java
johncarl81: I've been going with the assumption that if you annotate your bean with `@Parcel` that would mean Parceler is on the classpath.  Perhaps we should just support collections with supported types and at least one bean annotated with `@Parcel`... would that be acceptable instead of checking explicitly that Parceler is on the classpath?

WonderCsabo: @johncarl81 it is very easy to check it is on the classpath: `elementUtils.getTypeElement("org.parceler.Parcer") != null`. We use this type of check for lots of things throughout AA.

johncarl81: Ok, that will work nicely, I'll move forward with that then.

WonderCsabo: Great!

johncarl81: I'm finding that matching types following the critieria we discussed may be a bit overzelous.   For instance, this is matching the `ArrayList<String> listExtra` in `ExtraInjectedActivity`.  Should we be serializing this through Parceler?

johncarl81: I guess previously `ArrayList<String>` was being handled as a `Serializable`.  Maybe it is better to use Parcelable to serialize it.

WonderCsabo: No, i think we should only serialize those objects through Parceler which are not in plain Android.

johncarl81: Agreed... I think I have a solution, pushing shortly.

AndroidAnnotations/androidannotations-core/androidannotations/src/main/java/org/androidannotations/helper/ValidatorHelper.java
```diff
@@ -546,8 +546,10 @@ public void canBePutInABundle(Element element, ElementValidation valid) {
 			if (typeMirror.getKind() != TypeKind.NONE) {
 				TypeMirror parcelableType = annotationHelper.typeElementFromQualifiedName(CanonicalNameConstants.PARCELABLE).asType();
 				TypeMirror serializableType = annotationHelper.typeElementFromQualifiedName("java.io.Serializable").asType();
-				if (!annotationHelper.isSubtype(typeMirror, parcelableType) && !annotationHelper.isSubtype(typeMirror, serializableType)) {
-					valid.addError("Unrecognized type. Please let your attribute be primitive or implement Serializable or Parcelable");
+				if (!annotationHelper.isSubtype(typeMirror, parcelableType)
+						&& !annotationHelper.isSubtype(typeMirror, serializableType)
+						&& !annotationHelper.isParcelType(typeMirror)) {
```
WonderCsabo: I just realized we have a problem here. We only allow `Parcelable`, `Serializable` (and now Parceler supported) types. But that will leave out most of [these types for example](https://github.com/johncarl81/androidannotations/blob/parceler/AndroidAnnotations/androidannotations-core/androidannotations/src/main/java/org/androidannotations/helper/BundleHelper.java#L47). We have to fix this later...

AndroidAnnotations/androidannotations-core/androidannotations/src/main/java/org/androidannotations/helper/AnnotationHelper.java
```diff
@@ -55,6 +66,40 @@
 
 	public static final String DEFAULT_FIELD_NAME_VALUE = "value";
 	public static final String DEFAULT_FIELD_NAME_RESNAME = "resName";
+	private static final String PARCEL_ANNOTATION = "org.parceler.Parcel";
```
WonderCsabo: Can you move all Parceler-related things to another helper class (which can extend `AnnotationHelper`)?

WonderCsabo: I would move `PARCEL_ANNOTATION` to `CanonicalNameConstants` and use that constant everywhere.

AndroidAnnotations/androidannotations-core/androidannotations/src/main/java/org/androidannotations/helper/AnnotationHelper.java
```diff
@@ -95,6 +140,47 @@ public TypeElement typeElementFromQualifiedName(String qualifiedName) {
...
+						&& (!root || genericsSize > 0)) {
+					for (int i = 0; i < genericsSize; i++) {
+						genericsMatch &= isParcelType(declaredType.getTypeArguments().get(i), false);
+					}
+
+					return genericsMatch;
+				}
+			}
+		}
+		return false;
```
WonderCsabo: Very nice algorithm.

AndroidAnnotations/androidannotations-core/androidannotations/src/test/java/org/androidannotations/internal/helper/AndroidManifestFinderTest.java
```diff
@@ -122,4 +122,4 @@ private void createFile(File file) throws IOException {
 			fail("Could not creates test manifest" + file);
 		}
 	}
-}
\ No newline at end of file
+}
```
WonderCsabo: This is done on develop, please revert.

AndroidAnnotations/androidannotations-core/androidannotations/src/main/java/org/androidannotations/internal/core/helper/IntentBuilder.java
```diff
@@ -135,16 +139,18 @@ public JInvocation getSuperPutExtraInvocation(TypeMirror elementType, JVar extra
...
-				if (!typeUtils.isSubtype(elementType, stringType)) {
-					extraParameterArg = cast(environment.getClasses().SERIALIZABLE, extraParameterArg);
-				}
-			} else {
+			if (typeUtils.isSubtype(elementType, parcelableType)) {
 				TypeMirror serializableType = elementUtils.getTypeElement(SERIALIZABLE).asType();
 				if (typeUtils.isSubtype(elementType, serializableType)) {
 					extraParameterArg = cast(environment.getClasses().PARCELABLE, extraParameterArg);
 				}
+			} else if (!BundleHelper.METHOD_SUFFIX_BY_TYPE_NAME.containsKey(elementType.toString()) && annotationHelper.isParcelType(elementType)) {
```
WonderCsabo: I just realized that `METHOD_SUFFIX_BY_TYPE_NAME` is not containing everything which is supported by `Bundle`.

AndroidAnnotations/androidannotations-core/androidannotations-test/pom.xml
```diff
@@ -82,6 +82,17 @@
 			<version>140</version>
 			<scope>test</scope>
 		</dependency>
+		<dependency>
+			<groupId>org.parceler</groupId>
+			<artifactId>parceler</artifactId>
+			<version>1.0.3</version>
```
WonderCsabo: Can you extract the version to a property?

AndroidAnnotations/androidannotations-core/androidannotations/src/main/java/org/androidannotations/internal/core/helper/IntentBuilder.java
```diff
@@ -135,16 +139,18 @@ public JInvocation getSuperPutExtraInvocation(TypeMirror elementType, JVar extra
...
+				extraParameterArg = environment.getCodeModel().ref("org.parceler.Parcels").staticInvoke("wrap").arg(extraParameterArg);
```
WonderCsabo: Please update the comment `// Cast to Parcelable or Serializable if needed`. This is what i hate about comments, they rot...

AndroidAnnotations/androidannotations-core/androidannotations/src/main/java/org/androidannotations/helper/ValidatorHelper.java
WonderCsabo: I am stupid, ignore this.

AndroidAnnotations/androidannotations-core/androidannotations/src/main/java/org/androidannotations/helper/BundleHelper.java
```diff
@@ -225,6 +231,10 @@ public JExpression getExpressionToRestoreFromBundle(JClass variableClass, JExpre
 			expressionToRestore = JExpr.invoke(bundle, methodNameToRestore).arg(extraKey);
 		}
 
+		if (parcelerBean) {
+			expressionToRestore = environment.getCodeModel().ref("org.parceler.Parcels").staticInvoke("unwrap").arg(expressionToRestore);
```
WonderCsabo: Can you extract `"org.parceler.Parcels"` to `CanonicalNameConstants`?

WonderCsabo:
This is OK, however i realized `METHOD_SUFFIX_BY_TYPE_NAME` is outdated and we also do not support `SparceArray<T extends Parcelable>`.

AndroidAnnotations/androidannotations-core/androidannotations/src/main/java/org/androidannotations/internal/core/helper/IntentBuilder.java
johncarl81: This code is begging for a more unified approach.  For example, check out the way I did this 
in Parceler:
[Generators](https://github.com/johncarl81/parceler/blob/master/parceler/src/main/java/org/parceler/internal/Generators.java)
[Generators setup](https://github.com/johncarl81/parceler/blob/master/parceler/src/main/java/org/parceler/internal/ParcelerModule.java#L219-L258)
[Generators validation](https://github.com/johncarl81/parceler/blob/master/parceler/src/main/java/org/parceler/internal/ParcelableAnalysis.java#L575)
[Generators usage - generates read code](https://github.com/johncarl81/parceler/blob/master/parceler/src/main/java/org/parceler/internal/ParcelableGenerator.java#L284)
and in Transfuse:
[Bundle Property Repository](https://github.com/johncarl81/transfuse/blob/master/transfuse/src/main/java/org/androidtransfuse/analysis/repository/BundlePropertyBuilderRepository.java)
[Bundle Property Setup](https://github.com/johncarl81/transfuse/blob/master/transfuse/src/main/java/org/androidtransfuse/config/TransfuseAndroidModule.java#L219-L234)
[Bundle Property validation](https://github.com/johncarl81/transfuse/blob/master/transfuse/src/main/java/org/androidtransfuse/gen/variableBuilder/ExtraInjectionNodeBuilder.java#L93)
[Bundle Property usage - generates write code](https://github.com/johncarl81/transfuse/blob/master/transfuse/src/main/java/org/androidtransfuse/gen/IntentFactoryStrategyGenerator.java#L153)

WonderCsabo: Totally agree, the current code is just too complicated...

I also thought about creating a provider thing, which we discussed in your
original PR.
That way we could completely split off Parceler support to a plugin,
instead of adding Parceler-related
code directly to AA.

Your code is very clean and robust, congratulations!

WonderCsabo:
Can you rebase this? We switched our codemodel dependency, which uses some new class names and namespace.

johncarl81:
Sure!  Done.

WonderCsabo:
Thanks!

@yDelouis can we merge this?

yDelouis:
Javadoc for `@Extra`, `@FragmentArg` and `@SavedInstanceState` are still missing. But except, that, it is okay.

WonderCsabo:
And from even more annotations, you are right.

johncarl81:
Ok, let me know if you like that language.

By the way, if you guys would like great looking javadocs with less ceremony I'd recommend [Asciidoclet](http://asciidoctor.org/news/2013/06/03/asciidoclet-announcement/)

WonderCsabo:
We also have to add info for:

* `OnActivityResult.Extra`
* `Receiver.Extra`
* `ReceiverAction.Extra`
* `ServiceAction.Extra` (actually we already miss parameter type info on that :( )

johncarl81:
I don't see `ServiceAction.Extra'...

WonderCsabo:
Sorry. Just ServiceAction.

johncarl81:
Alright, I think I got them all... let me know if there's anything else.

johncarl81:
Ping

WonderCsabo:
Sorry for not adding feedback, i am very busy at work :(

I think this can be merged. @yDelouis ?

yDelouis:
I think so ! Thanks !

johncarl81:
Fantastic!

Thanks @WonderCsabo for your help on this effort, both on this PR and your attention to Eclipse support in Parceler.  Without your diligence this feature would have never reached this point.

WonderCsabo:
@johncarl81 i hope i could help, and thank you for implementing this in AA, it was a quick task.

@yDelouis we have to update the wiki. I think we can add this information in the third party integration section. Is that OK?

WonderCsabo:
@johncarl81 i created a [wiki page](https://github.com/excilys/androidannotations/wiki/ParcelerIntegration) for Parceler integration, i hope everything is correct.

johncarl81:
Looks spot on @WonderCsabo.

WonderCsabo:
Great, thanks for checking.

WonderCsabo:
@johncarl81 FYI i tried this in Gradle, and the `JCodeModel#ref` method throwed an `NoClassDefFoundError` about `android.os.Parcelable` when you are referencing the `Parcels` utility class. So i changed to use `JCodeModel#directClass` instead 12342808.

johncarl81:
Ah, good catch.

PR #1561, opened 2015-09-20
Fix in order parameter validation

### LINKED ISSUE #1443: In order parameter validation is broken
For example the following validation:

```java
validatorHelper.param.inOrder() //
				.type(boolean.class.getName()).optional() //
				.type(long.class.getName()).optional() //
				.type(int.class.getName()).optional() //
				.validate(executableElement, valid);
```

allows this method:
```java
void select(AdapterView<Adapter> a, long c, boolean b, Bundle bundle) { }
```

yDelouis:
Fixes #1443.

### DIFF
--- a/AndroidAnnotations/androidannotations-core/androidannotations/pom.xml
+++ b/AndroidAnnotations/androidannotations-core/androidannotations/pom.xml
@@ -56,6 +56,10 @@
 			<version>4.1.1.4</version>
 			<scope>provided</scope>
 		</dependency>
+		<dependency>
+			<groupId>org.mockito</groupId>
+			<artifactId>mockito-core</artifactId>
+		</dependency>
 	</dependencies>
 
 	<build>
--- a/AndroidAnnotations/androidannotations-core/androidannotations/src/test/java/org/androidannotations/helper/ValidatorParameterHelperTest.java
+++ b/AndroidAnnotations/androidannotations-core/androidannotations/src/test/java/org/androidannotations/helper/ValidatorParameterHelperTest.java
@@ -1,0 +1,348 @@
+/**
+ * Copyright (C) 2010-2015 eBusiness Information, Excilys Group
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may not
+ * use this file except in compliance with the License. You may obtain a copy of
+ * the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed To in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+ * License for the specific language governing permissions and limitations under
+ * the License.
+ */
+package org.androidannotations.helper;
+
+import static org.junit.Assert.assertFalse;
+import static org.junit.Assert.assertTrue;
+import static org.mockito.Matchers.any;
+import static org.mockito.Mockito.doAnswer;
+import static org.mockito.Mockito.doReturn;
+import static org.mockito.Mockito.mock;
+
+import java.lang.annotation.Annotation;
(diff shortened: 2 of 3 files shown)


AndroidAnnotations/androidannotations-core/androidannotations/pom.xml
```diff
@@ -56,6 +56,11 @@
 			<version>4.1.1.4</version>
 			<scope>provided</scope>
 		</dependency>
+		<dependency>
+			<groupId>org.mockito</groupId>
+			<artifactId>mockito-core</artifactId>
+			<version>1.10.8</version>
```
WonderCsabo: Missing `test` scope!

WonderCsabo: Moreover, we use Mockito in other modules as well. So you can pull it to the root POM in `dependencyManagement`, just like we do with JUnit.

AndroidAnnotations/androidannotations-core/androidannotations/src/test/java/org/androidannotations/helper/ValidatorParameterHelperTest.java
```diff
@@ -0,0 +1,321 @@
+package org.androidannotations.helper;
+
```
WonderCsabo: OK, this file should be reformatted.

WonderCsabo:
The test suite is far from full coverage, but it is a good start. Thanks for this PR, we really needed this fix, and we can move on.

It seems Travis is down, so that is why it did not tell you have lots of Checkstyle errors in the test class and a missing licence header. Please fix this, as well as my other comments. After it, the PR can be merged.

WonderCsabo:
We have a problem:

```
ItemClicksHandledActivity.java:[116,9] org.androidannotations.annotations.ItemSelect can only have the following parameters: [ [ boolean or java.lang.Boolean ], [ extending java.lang.Object (optional) ],  ] in the order above
ItemClicksHandledActivity.java:[131,9] org.androidannotations.annotations.ItemSelect can only have the following parameters: [ [ boolean or java.lang.Boolean ], [ extending java.lang.Object (optional) ],  ] in the order above
ItemClicksHandledActivity.java:[142,9] org.androidannotations.annotations.ItemSelect can only have the following parameters: [ [ boolean or java.lang.Boolean ], [ extending java.lang.Object (optional) ],  ] in the order above
```

The problem here that `ItemSelect` really accepts any type (any reference type or primitive type), except that if the second parameter is primitive int, than it is treated as a position, otherwise as the element.
But `anyType()` does not accept any types, only any classes. We should rename it to `anyClass()` or such, then introduce a new `anyType()` validator which indeed accepts any type (primitive or class).

AndroidAnnotations/androidannotations-core/androidannotations/pom.xml
yDelouis: Fixed.

AndroidAnnotations/androidannotations-core/androidannotations/src/test/java/org/androidannotations/helper/ValidatorParameterHelperTest.java
yDelouis: I worked on my professional computer and forgot to setup eclipse formatter on it.

yDelouis:
I addressed your comments.
I hope @dodgex is okay for setting him as author of the last commit since I copied-pasted his gist.

dodgex:
@yDelouis sure i'm okay with that. Thanks :)

WonderCsabo:
@yDelouis i just saw the `getWrapperType()` method, and it is half-done, not sure why. Can you fix that?

WonderCsabo:
@yDelouis also please rebase this so the build can go green.

WonderCsabo:
@yDelouis any update on this?

PR #1562, opened 2015-09-20
use AA 4.0-SNAPSHOT

dodgex:
the `maveneclipse` sample is still using AA 3.3.2


WonderCsabo:
Actually, this is intentional. We use the last stable version, because when a user comes to our GH page, he will just go to the examples, and in that case she should see the latest stable version. We use 3.3.2 in the Eclipse and Maven examples. We cannot do that with Gradle ones, because it is not easy to inject a simply variable (the version) in the command line when we build the CI (in the CI we use the latest snapshot version).

We could assume that users are smart and will checkout the tag or switch to the master branch to see the examples. But i think they do not do that. At least i also forget to do that when poking around other projects.

dodgex:
ah okay...

dodgex:
i wont assume that they checkout the correct tag.

WonderCsabo:
@dodgex i am not sure this is a good thing or not. I just checked the Dagger project, which uses the latest snapshot in its examples.

dodgex:
i'd prefer that too. having the version that is current in the scope. snapshot on dev and the fixed version on master & tags.

WonderCsabo:
It is easier to manage for sure. I am okay to change that. @yDelouis WDYT?

dodgex:
i reopen the PR. depending on @yDelouis either merge or close it again. :)

WonderCsabo:
@dodgex this PR should also remove the flag in Travis, since it is unnecessary. Also, you can add the same warning next to the version which we added in the Gradle projects.

However, i am not sure what to do with the Eclipse project...

WonderCsabo:
Also, we should rename the `maveneclipse` project to just `maven`. The current name is quite misleading.

dodgex:
i think i'd just remove the pure eclipse sample.

WonderCsabo:
Since we do not provide the bundles anymore, it has a point. Nobody should use pure IDE projects (Eclipse, IntelliJ, etc does not matter).

dodgex:
i updated the PR.

yDelouis:
I think it's better than the example project uses the snapshot because it will force us to update the example while changing AA. For example, by creating the plugin system, examples won't work anymore.

WonderCsabo:
We already used the snapshot for the examples in CI for the same reason.
They work, because they do not use annotations moved to plugins.

dodgex:
before merging the PR (if wanted) i will update the package inside the maven sample as there is still the `maveneclipse` present.

examples/HelloWorldEclipse/src/org/androidannotations/helloworldeclipse/ActivityWithExtra.java
```diff
@@ -1,42 +0,0 @@
-package org.androidannotations.helloworldeclipse;
```
WonderCsabo: The Eclipse example contained much more example code than the Gradle or Maven one. Can you copy all the example codes to both samples?

dodgex:
- rebased
- squashed commits
- updated maven and gradle samples with code from eclipse sample
- removed eclipse sample

WonderCsabo:
Thanks, this will makes things easier.

I updated to releasing wiki page.

PR #1563, opened 2015-09-21
Use visibility modifier from overridden method

### LINKED ISSUE #1493: @Background on @EBean - Protected changed to Public
Hi @all,

if I mark a protected method in an @EBean as @Background the view level for this method is changed from protected to public. If you need to call the Bean directly (like MyBean_.getInstance_(context)) the hidden methods are viewable and usable now!

WonderCsabo:
Fixes #1493.


yDelouis:
Good.

Issue #1323, opened 2015-02-16
Problems with "id" package name prefix


larvyde:
This happens when I tried using @OptionsItem on my app which happens to have a package name that starts with id. AA imports id.co.packagename.R.id so when it tries to inflate(id.co.packagename.R.menu.logout) java thinks the 'id' in the package name is the generated resource class R.id. I can't change the package name since it's tied to the company domain (indonesian domains have the ccTLD .id).

Since AA imports R.id, R.layout, etc, why doesn't it import R.menu as well, which would solve this? Either that or don't import anything and just use the fully qualified name (it's generated code anyway).

I'm using version 3.2. I apologise if this had been fixed in another version.

WonderCsabo:
Unfortunetaly this is a little bit more complicated.

When we process your code, we do not know what way you passed the menu id, we only get the integer value. So if we want to know the R.xxx.yyy field, we have to scan the R classes and find the value you passed. 

We could just use the value itself, but we have to get the field for two reasons in this case:

* before generating code, we always validate the usage of the annotation, so we do not generate not compilable or not working code, but add a compile time error message instead. In this case, we have to be sure that the integer value indeed corresponds to a menu.
* We want to generate as nice and readable code as possible. That's why we use the R.xxx.yyy constants in our generated code, not just simply the passed value.

It seems our R class scanner has a bug in it, so it gets confused with id as the package name prefix.

However, the imports are fully created by our Java code generator dependency, codemodel. Unfortunately it has that bad import strategy, which leads to a non-compilable code in your case. I am afraid we cannot configure it.

larvyde:
Alright, that is understandable.
For posterity, if anyone runs into this problem, I ended up sidestepping the issue by copying over the generated code from Activity_.java and changed the inflate() line which only adds some 4 or so lines of code so it's not too big a deal.

WonderCsabo:
Thanks for sharing the workaround us and for pointing out the problem in the first place.

WonderCsabo:
I am closing this issue for now, because there is no way to fix this currently. :cry:

WonderCsabo:
@larvyde sorry, but can you explain the problem here again.

I tested this, and this code was created:
```java
import id.androidannotations.gradle.R.id;
// ... 
@Override
public boolean onOptionsItemSelected(MenuItem item) {
	int itemId_ = item.getItemId();
	if (itemId_ == id.myItem) {
		myItemSelected();
		return true;
	}
[...]

PR #1566, opened 2015-09-24
Upgrade to JCodeModel 2.7.11

WonderCsabo:
Taking advantage of some API changes, as well fixing #1323.

### DIFF
--- a/AndroidAnnotations/androidannotations-core/androidannotations/src/main/java/org/androidannotations/internal/rclass/RInnerClass.java
+++ b/AndroidAnnotations/androidannotations-core/androidannotations/src/main/java/org/androidannotations/internal/rclass/RInnerClass.java
@@ -32,6 +32,8 @@ import org.androidannotations.helper.CaseHelper;
 import org.androidannotations.rclass.IRInnerClass;
 
 import com.helger.jcodemodel.AbstractJClass;
+import com.helger.jcodemodel.JClassAlreadyExistsException;
+import com.helger.jcodemodel.JDirectClass;
 import com.helger.jcodemodel.JFieldRef;
 
 public class RInnerClass implements IRInnerClass {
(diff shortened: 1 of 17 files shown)


AndroidAnnotations/androidannotations-core/androidannotations/src/main/java/org/androidannotations/internal/rclass/RInnerClass.java
```diff
@@ -122,9 +124,30 @@ public static JFieldRef extractIdStaticRef(AndroidAnnotationsEnvironment environ
 			int fieldSuffix = layoutFieldQualifiedName.lastIndexOf('.');
 			String fieldName = layoutFieldQualifiedName.substring(fieldSuffix + 1);
 			String rInnerClassName = layoutFieldQualifiedName.substring(0, fieldSuffix);
+			int innerClassSuffix = rInnerClassName.lastIndexOf('.');
 
-			AbstractJClass refClass = environment.getJClass(rInnerClassName);
-			return refClass.staticRef(fieldName);
+			String rClassQualifiedName = rInnerClassName.substring(0, innerClassSuffix);
```
yDelouis: It is more complicated. Why it is better ?

AndroidAnnotations/androidannotations-core/androidannotations/src/main/java/org/androidannotations/internal/rclass/RInnerClass.java
```diff
@@ -122,9 +124,30 @@ public static JFieldRef extractIdStaticRef(AndroidAnnotationsEnvironment environ
...
+				if (clazz.name().equals(innerClassSimpleName)) {
+					innerClass = clazz;
+					break;
+				}
+			}
+			if (innerClass == null) {
+				try {
+					innerClass = rClass._class(innerClassSimpleName);
+				} catch (JClassAlreadyExistsException e) {
+					// never happens, since we already checked the inner class does not exist
```
yDelouis: If it never happens, I think we should throw a `RuntimeException`. So that, it will crash and users will report it.

yDelouis:
Is it possible to remove the method in `APTCodeModelHelper` where we use reflection ?

AndroidAnnotations/androidannotations-core/androidannotations/src/main/java/org/androidannotations/internal/rclass/RInnerClass.java
WonderCsabo: It will import just the R class instead of R.string, and fix #1323.

AndroidAnnotations/androidannotations-core/androidannotations/src/main/java/org/androidannotations/internal/rclass/RInnerClass.java
WonderCsabo: It logically never happens, we will never throw that exception. :)

WonderCsabo:
That method is a thorn in my flesh, but i think we can only achieve that through reflection.

AndroidAnnotations/androidannotations-core/androidannotations/src/main/java/org/androidannotations/internal/rclass/RInnerClass.java
yDelouis: Okay, thanks.

dodgex:
Hey Guys, are there any News on this PR? :)

WonderCsabo:
From my side, it can be merged. @yDelouis ?

yDelouis:
You're right. We didn't agree about throwing the exception but it's not a big deal.

WonderCsabo:
Yeah, thanks!

Issue #1466, opened 2015-06-11
Calling a @Background/@WakeLock method in a @AfterInject method causes NullPointerException


feugatos:
Hi there, I'm experience a bug with AndroidAnnotations 3.3.1.

I have an @EFragment support fragment with an @AfterInject method where I want to perform some initialization. Calling a @Background/@WakeLock annotated method from within the @AfterInject method results in a NullPointerException.

```
java.lang.NullPointerException: Attempt to invoke virtual method 'android.os.PowerManager$WakeLock android.os.PowerManager.newWakeLock(int, java.lang.String)' on a null object reference : ... )
```

Looking at the generated code I see this:

```
private void init_(Bundle savedInstanceState) {
        OnViewChangedNotifier.registerOnViewChangedListener(this);
        Resources resources_ = getActivity().getResources();
        ...
        restoreSavedInstanceState_(savedInstanceState);
        init(); <- This is my @AfterInject method
        powerManager_ = ((PowerManager) getActivity().getSystemService(Context.POWER_SERVICE));
    }
```

The problem is that the powerManager_ is initialized after calling my @AfterInject method and thus the NullPointerException.

Shouldn't any required System Components/Services be initialized before calling any annotated method?

The fix for this would be to call any annotated methods after having first the reference to the required System Components/Services.

Putting my initialization code in a @AfterViews method doesn't result in the same exception.

Thanks in advance.

yDelouis:
The injection of `powerManager_` should be in `init();`, before calls to methods annotated with `@AfterInject`. This is indeed a bug, and it's caused by the fact that the annotation `@WakeLock` must be processed after other annotations and it's badly managed.

@WonderCsabo, We should refactor a bit this part to make handlers more independent. In `init_` we should create a `block` or a `method` for both `inject`, `afterInject`, `injectViews`, `afterInjectViews`, etc... So that the order of the handlers won't matters anymore.

WonderCsabo:
@yDelouis i totally agree. I vote for a `block`.

yDelouis:
I prefer a method so that it's clearer in the generated code.

yDelouis:
Fixed by #1585.

PR #1585, opened 2015-10-10
Create blocks for injection and after injection

WonderCsabo:
Related to #1466.

This is incomplete. For some reasons, it generates the following:
```java
private void init_(Bundle savedInstanceState) {
	{
		Resources resources_ = this.getResources();
		injected_string = resources_.getString(R.string.injected_string);
		injectedString = resources_.getString(R.string.injected_string);
		helloHtml = Html.fromHtml(resources_.getString(R.string.hello_html));
		htmlInjected = Html.fromHtml(resources_.getString(R.string.hello_html));
	}
}
```

Note the extra block with braces and indent. I could not figure out why is this happening. Any help is much appreciated.

### DIFF
--- a/AndroidAnnotations/androidannotations-core/androidannotations/src/main/java/org/androidannotations/holder/EActivityHolder.java
+++ b/AndroidAnnotations/androidannotations-core/androidannotations/src/main/java/org/androidannotations/holder/EActivityHolder.java
@@ -443,7 +443,7 @@ public class EActivityHolder extends EComponentWithViewSupportHolder implements 
 		injectExtrasBlock = injectExtrasBody._if(injectExtras.ne(_null()))._then();
 
 		getSetIntent().body().invoke(injectExtrasMethod);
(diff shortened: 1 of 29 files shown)


WonderCsabo:
The problem can be seen in `ContextCompatDrawableTest.activityCompilesOnMinSdkLower21CompileSdkHigher21WithoutContextCompat()`.

dodgex:
i have a similar problem. jcodemodel sets the braces to required as soon as one calls `decl`. see [JBlock.java#L275](https://github.com/phax/jcodemodel/blob/master/src/main/java/com/helger/jcodemodel/JBlock.java#L275)

WonderCsabo:
Thanks! I just realized this myself. See https://github.com/phax/jcodemodel/issues/32.

dodgex:
haha :) good to know. btw we can do some kind of workaround by calling `.bracesRequired(false).indentRequired(false);` but this is... kind of ugly :(

yDelouis:
Great.

Issue #1333, opened 2015-02-28
Feature Request add @Patch to RestClient


bperin:
I found myself needing to send patch requests and it seems like a pretty straight forward thing and dont want to use a different rest client but can't figure out a hack yet. Any suggestions?

Spring Android https://github.com/spring-projects/spring-android has patch support as of 2.0.0-snapshot

WonderCsabo:
We do not support 2.0.0, yet. See #1140.

bperin:
Can you think of a hack? I tried taking my generated apiclient code and making it a regular non generated class and put the method as patch but spring framework still throws an error

03-02 16:23:52.463: W/System.err(3129): org.springframework.web.client.ResourceAccessException: I/O error on PATCH request for "xxx": Unknown method 'PATCH'; must be one of [OPTIONS, GET, HEAD, POST, PUT, DELETE, TRACE]; nested exception is java.net.ProtocolException: Unknown method 'PATCH'; must be one of [OPTIONS, GET, HEAD, POST, PUT, DELETE, TRACE]

I also tried to create my own annotation
@Retention(RetentionPolicy.CLASS)
@Target(ElementType.METHOD)
public @interface Patch {
    String value();
}


but it fails in build

WonderCsabo:
Unfortunetaly that will not work.

I think you got your first error because you are trying to use `PATCH` with Spring for Android 1.0.0 which is not supported there. If you trying to update to 2.0.0, it will fail because we do not generate correct code.

Adding an annotation does nothing, you have to also add a corresponding processor.

WonderCsabo:
This is a valid request, but it is blocked by #1140.

WonderCsabo:
This is no longer blocked. Contributions are welcome!

bperin:
Ill revisit this

WonderCsabo:
BTW, i do not really know PATCH. What is the request type and response type in this case? Just like POST?

WonderCsabo:
Implemented.

PR #1587, opened 2015-10-10
Patch for Rest

simopete:
Implements #1333 

Some test are missing because PATCH is only available in Spring Android Rest Template 2.0


WonderCsabo:
I decided to merge this PR without explicit tests for `@Patch`. Actually that is just like POST, so its functionality is already tested. We could switch our test suite to use Spring Android 2.0, however i am not sure about that, because it is still not released, and we cannot use milestone or snapshot dependencies. @yDelouis WDYT?

WonderCsabo:
@simopete great work, thanks! Can you update the wiki? Please note the Part and Field annotations should be moved to a new section.

WonderCsabo:
Wiki merged, thanks!

PR #1588, opened 2015-10-10
Change default log level to WARN

WonderCsabo:


yDelouis:
I will be able to remove the option from the build.gradle file ! Cool !

PR #1589, opened 2015-10-11
Fix @Patch validation

WonderCsabo:
Follow-up for #1587.

We should have tests for Spring 2. The most coverage would be if we create a new test project which uses Spring 2. Or we can just change the the version for 2, but that means we do not test important things which are differently generated in Spring 1.


yDelouis:
Ok.

Issue #1591, opened 2015-10-12
The order of the ancestor elements are random


shiraji:
While I was working with #1156, I found a strange behavior of AA with ancestors elements.
Here is a test case for ancestors elements.

```java
@EActivity
@OptionsMenu(R.menu.menu_1)
public abstract class TopActivity extends AppCompatActivity {
}

@EActivity
@OptionsMenu(R.menu.menu_2)
public abstract class SubActivity extends TopActivity {
}

@EActivity
@OptionsMenu(R.menu.menu_main)
public class MainActivity extends SubActivity {
}
```

In this case, the generated code expected to be

```java
public final class MainActivity_
    extends MainActivity
    implements HasViews
    ...

    @Override
    public boolean onCreateOptionsMenu(Menu menu) {
        MenuInflater menuInflater = getMenuInflater();
        menuInflater.inflate(com.github.shiraji.myapplication.R.menu.menu_1, menu);
        menuInflater.inflate(com.github.shiraji.myapplication.R.menu.menu_2, menu);
        menuInflater.inflate(com.github.shiraji.myapplication.R.menu.menu_main, menu);
        return super.onCreateOptionsMenu(menu);
    }
    ...
    
```

However, it "sometimes" generate this code.

```java
public final class MainActivity_
    extends MainActivity
    implements HasViews
    ...

    @Override
    public boolean onCreateOptionsMenu(Menu menu) {
        MenuInflater menuInflater = getMenuInflater();
        menuInflater.inflate(com.github.shiraji.myapplication.R.menu.menu_2, menu);
        menuInflater.inflate(com.github.shiraji.myapplication.R.menu.menu_1, menu);
        menuInflater.inflate(com.github.shiraji.myapplication.R.menu.menu_main, menu);
        return super.onCreateOptionsMenu(menu);
    }
    ...
    
```

If there is grand-parents class, the order of the ancestor elements are random. This is because how AA traverses and stores ancestors. Now, I am ready for pull request for this issue. Could you confirm this is a problem?

WonderCsabo:
I agree that we should definitely traverse the inheritance hierarchy correctly.

shiraji:
Thanks for your quick response. I will send pull request for this issue.

WonderCsabo:
Thanks for fixing this!

shiraji:
I’m glad to help. Thanks for merging my pull request.

PR #1592, opened 2015-10-12
Fixed a problem with travasing ancestors.

shiraji:
see #1591

### DIFF
--- a/AndroidAnnotations/androidannotations-core/androidannotations/src/test/java/org/androidannotations/menu/OptionsMenuItemTest.java
+++ b/AndroidAnnotations/androidannotations-core/androidannotations/src/test/java/org/androidannotations/menu/OptionsMenuItemTest.java
@@ -36,4 +36,22 @@ public class OptionsMenuItemTest extends AAProcessorTestHelper {
 		assertCompilationSuccessful(result);
 	}
 
+	@Test
+	public void testDeeplyInheritedMenuNotBreakOrder() throws IOException {
+		CompileResult result = compileFiles(OptionsMenuItemSubSubSubActivity.class);
+		assertCompilationSuccessful(result);
+		// CHECKSTYLE:OFF
+		String[] codeFragment = { //
+				"    @Override", //
+				"    public boolean onCreateOptionsMenu(Menu menu) {", //
+				"        MenuInflater menuInflater = getMenuInflater();", //
+				"        this.menu = menu.findItem(R.id.menu);", //
+				"        this.menu1 = menu.findItem(R.id.menu1);", //
+				"        this.menu2 = menu.findItem(R.id.menu2);", //
+				"        this.menu3 = menu.findItem(R.id.menu3);", //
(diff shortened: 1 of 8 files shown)


AndroidAnnotations/androidannotations-core/androidannotations/src/test/java/org/androidannotations/menu/OptionsMenuItemTest.java
```diff
@@ -36,4 +36,10 @@ public void testWhenFieldNameCouldConflictWithParameterName() throws IOException
 		assertCompilationSuccessful(result);
 	}
 
+	@Test
+	public void testDeeplyInheritedMenuNotBreak() throws IOException {
+		CompileResult result = compileFiles(OptionsMenuItemSubSubSubActivity.class);
+		assertCompilationSuccessful(result);
```
WonderCsabo: Is this test sufficient? You should check actually the code is in order, right?

shiraji: Yes. You are right. I fixed the test case.

shiraji:
mmm...I couldn't reproduce this failure. I ran following commands:

```shell
mvn clean
mvn install -Dandroid.sdk.path=/opt/android-sdk/
```

Could you point how I can fix this problem, @WonderCsabo ?

shiraji:
```java
    @Override
    public boolean onCreateOptionsMenu(Menu menu) {
        MenuInflater menuInflater = getMenuInflater();
        this.menu = menu.findItem(R.id.menu);
        this.menu1 = menu.findItem(R.id.menu1);
        this.menu2 = menu.findItem(R.id.menu2);
        this.menu3 = menu.findItem(R.id.menu3);
        return super.onCreateOptionsMenu(menu);
    }
```

There is `R.id`...mmm, let me create another test cases.

WonderCsabo:
We are using a better way to generate imports to R classes, so we no longer
import the inner R classes, but only R. This is the recommended on Android.
See #1548

shiraji:
@WonderCsabo Could you give me a sample unit test code? After I learn phax/jcodemodel, I still don't understand how I implement this unit test code. Thanks.

WonderCsabo:
@shiraji i just rebased your branch to latest develop use the latest modifications for jcodemodel, and fixed the test case. Check out [this](https://github.com/WonderCsabo/androidannotations/tree/orderAncestors) branch.

WonderCsabo:
BTW @shiraji you do not need to log the generated Java code. You can just check the file under *target/generated-test*.

shiraji:
Thanks, let me fix the issue.

I logged the generated Java code just because I want to know what is the difference between my local generated code and travic ci's generated code.

WonderCsabo:
Please squash the last commit into commit before.

shiraji:
Here you go.

Issue #1578, opened 2015-10-07
Always call #onNewIntent(), even without @AfterExtras method(s)


m-radzikowski:
If there is any method with `@AfterExtras` passing new Intent to activity causes calling `onNewIntent()`, from there `setIntent()` and finally `injectExtras_()` where all new extras are set. And it's great.

On the other hand if there is no any `@AfterExtras` method generated subclass has no `onNewIntent()` method, and in result extra is not updated.

This is unfortunate if extra does not cause any instant action, but is used later - for example in my case there is only one extra that is passed by to next activity, without any operations on it. So currently I had to add blank method with `@AfterExtras` annotation just to make sure that `onNewIntent()` is generated.

WonderCsabo:
The `@Extra`s are injected every time when `setIntent()` is called. I am not sure we should change that to `onNewIntent()`. @dodgex @yDelouis ?

m-radzikowski:
But `setIntent()` is not called if there is no `@AfterExtras` method. On the other hand, if there is any, it is called. So depending on having any `@AfterExtras` methods extras are or are not updated - it's inconsistency for me.

WonderCsabo:
The extras are injected in two places In `setIntent()` and in `onCreate()`. So they will be injected even if there is no `@AfterExtras` method, in `onCreate()`. If the `Activity` is not created, just a new intent reaches the `Activity`, then we only inject the extras if the `setIntent()` method is called. Which seem to be correct to me, because if that is not called, the new `Intent` does not really belong to the `Activity`. The `@AfterExtras` annotation clearly states the the user wants to work with the injected extras, so we call `setIntent()` there.

You are saying that you are just using `@AfterExtras` as dummy now. My question is: which lifecycle method you use to actually work with the newly injected extras?

m-radzikowski:
> You are saying that you are just using @AfterExtras as dummy now. My question is: which lifecycle method you use to actually work with the newly injected extras?

In my (very uncommon I think ;-) ) case I'm only passing this particular extra (if exists) to the next Activity, without interacting with it. The case is this Intent comes from notification click and next Activity must be informed, what was in this notification.

Please consider following two example Activities:

```

```

m-radzikowski:
> You are saying that you are just using @AfterExtras as dummy now. My question is: which lifecycle method you use to actually work with the newly injected extras?

[...]

PR #1607, opened 2015-10-29
do not call setIntent() in onNewIntent()

dodgex:
see https://github.com/excilys/androidannotations/issues/1578


WonderCsabo:
Hmm, we did not have tests for this?

BTW, you are *fast*.

dodgex:
does not seem to be tested. there where no failures

dodgex:
i flash! =)

WonderCsabo:
@dodgex can you update the wiki and add a warning about this breaking change?

dodgex:
hmmm looks like we never mentioned that we override onNewIntent. do you know with which version we introduced the override?

WonderCsabo:
AA 3.2.

dodgex:
https://github.com/dodgex/androidannotations/wiki/Extras#handling-onnewintent

dodgex:
Thank you for the info. :)

WonderCsabo:
Wiki merged, thx.

Issue #1625, opened 2015-11-14
Extend the @DefaultRes annotation to support string arrays


athkalia:
As discussed in https://github.com/excilys/androidannotations/pull/1555#issuecomment-156751925, it would be handy if the @DefaultRes annotation also provided support for a Set of Strings, with a default value of arrays of Strings passed by it's R.array.id value.

WonderCsabo:
Implemented.

athkalia:
damn you are fast lol

WonderCsabo:
Thanks to @dodgex.

PR #1626, opened 2015-11-15
support string arrays in defaultres

dodgex:
see https://github.com/excilys/androidannotations/issues/1625

### DIFF
--- a/AndroidAnnotations/androidannotations-core/androidannotations/src/main/java/org/androidannotations/internal/core/handler/SharedPrefHandler.java
+++ b/AndroidAnnotations/androidannotations-core/androidannotations/src/main/java/org/androidannotations/internal/core/handler/SharedPrefHandler.java
@@ -96,7 +96,7 @@ public class SharedPrefHandler extends CoreBaseGeneratingAnnotationHandler<Share
 			put("int", new DefaultPrefInfo<>(DefaultInt.class, IntPrefField.class, IRClass.Res.INTEGER, 0, "intField"));
 			put("long", new DefaultPrefInfo<>(DefaultLong.class, LongPrefField.class, IRClass.Res.INTEGER, 0L, "longField"));
 			put(CanonicalNameConstants.STRING, new DefaultPrefInfo<>(DefaultString.class, StringPrefField.class, IRClass.Res.STRING, "", "stringField"));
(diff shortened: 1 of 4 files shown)


AndroidAnnotations/androidannotations-core/androidannotations/src/main/java/org/androidannotations/internal/core/handler/SharedPrefHandler.java
```diff
@@ -322,10 +322,21 @@ private IJExpression extractResValue(SharedPrefHolder holder, Element method, IR
...
 			break;
 		}
-		return holder.getContextField().invoke("getResources").invoke(resourceGetMethodName).arg(idRef);
+
+		JInvocation resourceInvocation = holder.getContextField().invoke("getResources").invoke(resourceGetMethodName).arg(idRef);
+
+		if (IRClass.Res.ARRAY.equals(res)) {
+			JInvocation asList = getClasses().ARRAYS.staticInvoke("asList");
+			JInvocation newHashMap = JExpr._new(getClasses().HASH_SET.narrow(getClasses().STRING));
+			resourceInvocation = newHashMap.arg(asList.arg(resourceInvocation));
```
WonderCsabo: We should handle the empty array here as well (I think it is possible to define an empty string array resource or not?)

WonderCsabo:
Can you extend @DefaultRes javadoc and explain which method return value should use which resource type?

AndroidAnnotations/androidannotations-core/androidannotations/src/main/java/org/androidannotations/internal/core/handler/SharedPrefHandler.java
dodgex: emtpy string array resource should not be a prolem. this code generates `new HashSet<>(Arrays.asList(res.getStringArray(R.array.xxx)));` if the resource is empty, the array is emtpy, the list is empty and the set is empty. also i think we have no way to access the number of items of a resource on compile time as the value might be diffrent for e.g. locales or screen sizes.

WonderCsabo: Okay.

dodgex:
@WonderCsabo is the table ok?

WonderCsabo:
I think you are missing `long` in the table.

dodgex:
yeah you are right.

WonderCsabo:
Nice, thanks.

PR #1629, opened 2015-11-17
Use application context to retrieve WifiManager.

### LINKED ISSUE #1628: @SystemService for WifiManager causes a memory leak
Given I have an Activity with a `@SystemService` reference to WifiManager:

```
@SystemService
WifiManager wifiManager;
```

Then AA will generate this code:
```
wifiManager = ((WifiManager) this.getSystemService(Context.WIFI_SERVICE));
```

Which in turn might cause a memory leak (Eclipse MAT):
screen shot 2015-11-17 at 18 50 25

Once I researched and found this page:
https://code.google.com/p/android/issues/detail?id=43945

And then proceeded to change the code in `onCreate` like this:
```
wifiManager = ((WifiManager) getApplicationContext().getSystemService(Context.WIFI_SERVICE));
```

The leak disappeared. Should've known this a few hours ago.

Therefore, I suggest that AA generates the `@SystemService` code for WifiManager as above. (perhaps good to add a comment into the generated code as well).

ened:
Fixes #1628.

### DIFF
--- a/AndroidAnnotations/androidannotations-core/androidannotations/src/main/java/org/androidannotations/internal/core/handler/SystemServiceHandler.java
+++ b/AndroidAnnotations/androidannotations-core/androidannotations/src/main/java/org/androidannotations/internal/core/handler/SystemServiceHandler.java
@@ -66,6 +66,8 @@ public class SystemServiceHandler extends BaseAnnotationHandler<EComponentHolder
 
 		if (CanonicalNameConstants.APP_WIDGET_MANAGER.equals(fieldTypeQualifiedName)) {
 			createSpecialInjection(holder, fieldName, fieldTypeQualifiedName, serviceRef, methodBody, 21, "LOLLIPOP", getClasses().APP_WIDGET_MANAGER, "getInstance", true);
+		} else if (CanonicalNameConstants.WIFI_MANAGER.equals(fieldTypeQualifiedName)) {
+			methodBody.add(createApplicationContextInjection(holder, fieldName, fieldTypeQualifiedName, serviceRef, methodBody));
 		} else {
 			methodBody.add(createNormalInjection(holder, fieldName, fieldTypeQualifiedName, serviceRef, methodBody));
 		}
@@ -73,7 +75,7 @@ public class SystemServiceHandler extends BaseAnnotationHandler<EComponentHolder
 
 	@SuppressWarnings("checkstyle:parameternumber")
 	private void createSpecialInjection(EComponentHolder holder, String fieldName, String fieldTypeQualifiedName, JFieldRef serviceRef, JBlock methodBody, int apiLevel, String apiLevelName,
-										AbstractJClass serviceClass, String injectionMethodName, boolean contextNeeded) {
+			AbstractJClass serviceClass, String injectionMethodName, boolean contextNeeded) {
 		if (getEnvironment().getAndroidManifest().getMinSdkVersion() >= apiLevel) {
 			methodBody.add(createNormalInjection(holder, fieldName, fieldTypeQualifiedName, serviceRef, methodBody));
 		} else {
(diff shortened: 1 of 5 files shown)


ened:
Ready for review once build passes.

AndroidAnnotations/androidannotations-core/androidannotations/src/main/java/org/androidannotations/internal/core/handler/SystemServiceHandler.java
```diff
@@ -73,7 +75,7 @@ public void process(Element element, EComponentHolder holder) {
 
 	@SuppressWarnings("checkstyle:parameternumber")
 	private void createSpecialInjection(EComponentHolder holder, String fieldName, String fieldTypeQualifiedName, JFieldRef serviceRef, JBlock methodBody, int apiLevel, String apiLevelName,
-										AbstractJClass serviceClass, String injectionMethodName, boolean contextNeeded) {
+			AbstractJClass serviceClass, String injectionMethodName, boolean contextNeeded) {
```
WonderCsabo: Is this changed needed?

WonderCsabo:
Can you add a simple compile-time test which verifies this?

AndroidAnnotations/androidannotations-core/androidannotations/src/main/java/org/androidannotations/internal/core/handler/SystemServiceHandler.java
ened: That one came through code formatter.

ened:
Yes, please point me to an example for that.

ened:
Sorry, I'll check it later. Thx.

WonderCsabo:
You already wrote this kind of test here https://github.com/excilys/androidannotations/commit/a5eeb0d7d2e8fa62a432d9f9b18558f9c14feb28 . So just check the generated code line is the expected.

WonderCsabo:
@ened is this ready for review?

ened:
Yes

WonderCsabo:
@ened seems to be fine! Would you mind adding `AudioManager`, too? In that case, we could add a new method in the handler which returns the correct context reference for the service type. And instead of adding a new branch in the process method, we should call that in `createNormalInjection `.

ened:
Happy about your comments, once the CI finishes.

WonderCsabo:
Great, thanks. I am waiting with merging this PR for #1457, so avoid hard merge conflicts on that one.

ened:
@WonderCsabo it is your call, but I'd like to move that "Defect" PRs get pull priority over feature "Enhancement" PRs.

dodgex:
@WonderCsabo @ened i'm ok with merging this one first. as i'm currently to busy to finish the documenation part for the other PR it might be on hold for some time. :( 

@ened the reason why @WonderCsabo did not want to merge this one first is that the last rebase and conflict fix took far more than a hour. but as mentioned i currently don't have much time. so it is ok to merge this and then i have to rebase/fix the other PR once again.

AndroidAnnotations/androidannotations-core/androidannotations/src/main/java/org/androidannotations/internal/core/handler/SystemServiceHandler.java
WonderCsabo: Please revert this change.

ened: Done.

ened:
Hi @WonderCsabo, 

there has been some further research on this. I continuously ran into leaked `Context` objects from `AudioManager`, even the new AA version was applied and the service was retrieved via the application context (see http://stackoverflow.com/questions/6445052/android-context-memory-leak-listview-due-to-audiomanager#9757887 and https://github.com/square/leakcanary/issues/205). 

The suggested fix in the LeakCanary project is to use a extra `ContextWrapper` and override the `getSystemService` method there. I think that solution would be better than modifying AA:

- It's not always AAs concern to "fix" potential Android bugs
- We keep the AA code more lean
- We can not handle all corner cases as demonstrated by AudioManager

Should we remove that commit and add a clear documentation link to the `@SystemService` annotation?

WonderCsabo:
So what do you suggest? We should remove special injection for `WifiManager` or only for `AudioManager`?

ened:
I suggest to remove the special handling all together and let applications handle this (by ContextWrapper) within their "BaseActivity" class or whatever they use.

AA could and should make things easier for writing Android code, fixing platform bugs is not this projects responsibility, I'd say.

dodgex:
> AA could and should make things easier for writing Android code, fixing platform bugs is not this projects responsibility, I'd say.

while i generally think you are right, i'm not sure if that is true for every case. as long as we provide a feature that allows to retrieve the class, we should try to produce working/good code. and if the code we generate produces(or can produce) a memory leak, we should either try to avoid the leak or not provide the injection.

*slightly off-topic*: and even beyond that point, as we now have a plugin api, AA could even be used to fix/workaround other platform bugs. e.g. by generating the required boilerplate for the workaround - but that is the duty of external plugins. ;)

Issue #1568, opened 2015-09-27
@PageChange


ray1984:
Annotation for creation viewpager's OnPageChangeListener

yDelouis:
Why not. Do you want to contribute ?

ray1984:
I don't have contributing experience in AA-core

domsu:
I would like to contribute, I will create pull request this weekend.

WonderCsabo:
@domsu great! If you have any difficulties working with the codebase, just ping us in the [Gitter chat room}(https://gitter.im/excilys/androidannotations).

WonderCsabo:
@domsu any update on this one?

shiraji:
It seems no one is working on this issue. I will take case of this. Please let me know if you guys have any progress. 

Before I am going to implement this, I want to make sure I understand this issue correctly.

Here would be the generate code

```java
    @Override
    public void onViewChanged(HasViews hasViews) {
        {
            final ViewPager view = ((ViewPager) hasViews.findViewById(id.viewpager)); // whatever id that passed by value or resName. the same as TextChange
            if (view!= null) {
                view.addOnPageChangeListener(new ViewPager.OnPageChangeListener() {
                    @Override
                    public void onPageScrolled(int position, float positionOffset, int positionOffsetPixels) {
                        MainActivity_.this.pageScrolled(position, positionOffset, positionOffsetPixels); // method that annotated by @PageScrolled
                    }

                    @Override
                    public void onPageSelected(int position) {
                        MainActivity_.this.pageSelected(position); // method that annotated by @PageSelected
                    }

                    @Override
                    public void onPageScrollStateChanged(int state) {
                        MainActivity_.this.pageScrollStateChanged(state); // method that annotated by @PageScrollStateChanged
                    }
                }
                );
            }
        }
    }
```

* There is no `@PageChange` annotation
* This feature is really close with `TextChangeEvents`
* Parameter handling for `@PageSelected` seems the hardest part but algorithm is mostly the same with `@TextChange`

Is it ok that creating test cases that checks only compile code correctly? (No test case for `TextChangeEvents`...) Could anyone point me if there is test cases for this type of annotation?

WonderCsabo:
Implemented.

PR #1635, opened 2015-11-22
Added page change

shiraji:
see #1568

### DIFF
--- a/AndroidAnnotations/androidannotations-core/androidannotations-api/src/main/java/org/androidannotations/annotations/PageScrolled.java
+++ b/AndroidAnnotations/androidannotations-core/androidannotations-api/src/main/java/org/androidannotations/annotations/PageScrolled.java
@@ -1,0 +1,99 @@
+/**
+ * Copyright (C) 2010-2015 eBusiness Information, Excilys Group
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may not
+ * use this file except in compliance with the License. You may obtain a copy of
+ * the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed To in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+ * License for the specific language governing permissions and limitations under
+ * the License.
+ */
+package org.androidannotations.annotations;
+
+import java.lang.annotation.ElementType;
+import java.lang.annotation.Retention;
+import java.lang.annotation.RetentionPolicy;
+import java.lang.annotation.Target;
+
+/**
+ * <p>
+ * This annotation is intended to be used on methods to receive events defined
+ * by
+ * <code>android.support.v4.view.ViewPager.OnPageChangeListener.OnPageScrolled</code>
+ * when the current page is scrolled, either as part of a programmatically
+ * initiated smooth scroll or a user initiated touch scroll.
+ * </p>
+ * <p>
+ * The annotation value should be one or several R.id.* fields that refers to
+ * ViewPager or subclasses of ViewPager. If not set, the method name will be
+ * used as the R.id.* field name.
+ * </p>
+ * <p>
+ * The method MAY have multiple parameters, but the order must be following :
+ * </p>
+ * <ul>
+ * <li>A ViewPager parameter to know which view has targeted this event.</li>
+ * <li>An int parameter named position to get position index of the first page
+ * currently being displayed. Page position+1 will be visible if positionOffset
+ * is nonzero.</li>
+ * <li>A float parameter is value from [0,1) indicating the offset from the page
+ * at position.</li>
+ * <li>An int parameter named positionOffsetPixels is value in pixels indicating
+ * the offset from position.</li>
+ * </ul>
+ *
+ * <blockquote>
+ *
+ * Examples :
+ *
+ * <pre>
+ * &#064;PageScrolled(<b>R.id.viewpager</b>)
+ * void onPageScrolled(ViewPager view, int position, float positionOffset, int positionOffsetPixels) {
+ * 	// Something Here
+ * }
+ *
+ * &#064;PageScrolled
+ * void viewPager(ViewPager view) {
+ * 	// Something Here
+ * }
+ *
+ * &#064;PageScrolled(<b>{R.id.viewpager, R.id.viewpager2}</b>)
+ * void onPageScrolledOnMultipleViewPager(ViewPager v, int position) {
+ * 	// Something Here
+ * }
+ *
+ * &#064;PageScrolled(<b>R.id.viewpager</b>)
+ * void onPageScrolledNoParam() {
+ * 	// Something Here
+ * }
+ * </pre>
+ *
+ * </blockquote>
+ *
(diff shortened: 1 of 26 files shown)


shiraji:
This is ready for review.

WonderCsabo:
This is very great work. Congrats!

Please note that `addOnPageChangeListener` is a "new" method. Previously only `setOnPageChangeListener` was present. So you have to check that which method is available, and use `setOnPageChangeListener` if `addOnPageChangeListener` is not present.

Also you should validate that `ViewPager` class is present or not. And maybe c

We definitely should add functional test for this. You should check out the `SeekBar` test for example. AFAIK there is not way to access the listener in `ViewPager` directly or though a shadow class in Robolectric, so you should retrieve it by reflection (using FEST-Reflect).

WonderCsabo:
We also did have the feature that finding the parameter by the name. I think we should not use that. Instead, we should define an order of the parameters.

That way you could also use our simple `ValidatorParameterHelper` API. See an example [here](https://github.com/excilys/androidannotations/blob/develop/AndroidAnnotations/androidannotations-core/androidannotations/src/main/java/org/androidannotations/internal/core/handler/ClickHandler.java#L53). You should also use that on all new annotation validations in this PR.

shiraji:
For finding the method exist or not, do you have any sample case for this?
Validating if ViewPager is present or not seems not that hard. It will be like this.

```java
getProcessingEnvironment().getElementUtils().getTypeElement(CanonicalNameConstants.VIEW_PAGER) != null;
```

I'll take a look `SeekBar` for functional tests.

`ValidatorParameterHelper` looks nice! Let me use that.

For `@PageScrolled`, this would be the candidate order of parameters. ViewPager is the first parameter and rest is `onPageScrolled` parameter order.

```java
		validatorHelper.param.inOrder() //
				.type(CanonicalNameConstants.VIEW_PAGER).optional() //
				.primitiveOrWrapper(TypeKind.INT).optional() //
				.primitiveOrWrapper(TypeKind.FLOAT).optional() //
				.primitiveOrWrapper(TypeKind.INT).optional() //
				.validate((ExecutableElement) element, validation);
```

What do you think?

WonderCsabo:
For checking a method is avaible, you can use similar to [this](https://github.com/excilys/androidannotations/blob/d01274d531cf7c29337f5f09d6128118fc2b7575/AndroidAnnotations/androidannotations-core/androidannotations/src/main/java/org/androidannotations/internal/core/handler/ContextCompatAwareResHandler.java#L84).

The rest is okay!

shiraji:
You are fast! Thanks for your help.

shiraji:
Now, it ready for reviewing, again.

I will go to trip a few days where I am not available for the Internet. Until I will come back from there, could you wait for merging? I will squash commits.

Thanks.

AndroidAnnotations/androidannotations-core/androidannotations-test/src/main/java/org/androidannotations/test/PageChangeListenerActivity.java
```diff
@@ -0,0 +1,59 @@
...
+	public void pageScrollStateChanged(ViewPager v, int state) {
+		pageScrollStateChangedViewPager = v;
+		pageScrollStateChangedState = state;
+	}
+
+	@PageSelected(R.id.viewPager1)
+	public void pageSelected(ViewPager v, int position) {
+		pageSelectedViewPager = v;
+		pageSelectedPosition = position;
+	}
```
WonderCsabo: Can you add more test cases (when not all parameters are present)? You can use the same variables because the a new `Activity` is created for each test.

AndroidAnnotations/androidannotations-core/androidannotations-test/src/test/java/org/androidannotations/test/PageChangeListenerActivityTest.java
```diff
@@ -0,0 +1,91 @@
...
+		activity = setupActivity(PageChangeListenerActivity_.class);
+		viewPager = (ViewPager) activity.findViewById(R.id.viewPager1);
+	}
+
+	@Test
+	public void handlePageSelected() {
+		assertThat(activity.pageSelectedViewPager).isNull();
+		assertThat(activity.pageSelectedPosition).isEqualTo(0);
+
+		ViewPager.OnPageChangeListener listener = field("mOnPageChangeListener").ofType(ViewPager.OnPageChangeListener.class).in(viewPager).get();
```
WonderCsabo: Move this common line to `setUp`.

AndroidAnnotations/androidannotations-core/androidannotations/src/main/java/org/androidannotations/internal/core/handler/PageScrollStateChangedHandler.java
```diff
@@ -0,0 +1,103 @@
...
+
+	public PageScrollStateChangedHandler(AndroidAnnotationsEnvironment environment) {
+		super(PageScrollStateChanged.class, environment);
+	}
+
+	@Override
+	public void validate(Element element, ElementValidation validation) {
+		super.validate(element, validation);
+
+		validatorHelper.param.inOrder() //
```
WonderCsabo: Why not `anyOrder()`?

AndroidAnnotations/androidannotations-core/androidannotations/src/main/java/org/androidannotations/internal/core/handler/PageSelectedHandler.java
```diff
@@ -0,0 +1,102 @@
...
+public class PageSelectedHandler extends AbstractPageChangeHandler {
+	public PageSelectedHandler(AndroidAnnotationsEnvironment environment) {
+		super(PageSelected.class, environment);
+	}
+
+	@Override
+	protected void validate(Element element, ElementValidation validation) {
+		super.validate(element, validation);
+
+		validatorHelper.param.inOrder() //
```
WonderCsabo: Why not `anyOrder()`?

AndroidAnnotations/androidannotations-core/androidannotations-api/src/main/java/org/androidannotations/annotations/PageScrolled.java
```diff
@@ -0,0 +1,99 @@
...
+ * when the current page is scrolled, either as part of a programmatically
+ * initiated smooth scroll or a user initiated touch scroll.
+ * </p>
+ * <p>
+ * The annotation value should be one or several R.id.* fields that refers to
+ * ViewPager or subclasses of ViewPager. If not set, the method name will be
+ * used as the R.id.* field name.
+ * </p>
+ * <p>
+ * The method MAY have multiple parameter :
```
WonderCsabo: We should declare that they are all optional, but the order should must be this.

WonderCsabo:
LGTM in general, i added some comments. Thanks!

shiraji:
Thanks! I will fix then as soon as I come back.

shiraji:
Fixed it. It's ready for review, again.

WonderCsabo:
Looks good to me. Please squash the commits and i can merge.

shiraji:
Squashed

WonderCsabo:
Thanks, great PR! Would you mind updating the wiki as well?

shiraji:
No problem! I will let you know as soon as I update wiki.

Issue #1660, opened 2015-12-18
@NonConfigurationInstance code should be generated before @Bean injection


dodgex:
currently this code is generated:

```java
if (cast == null) {
	cast = CastManager_.getInstance_(this);
}
// nonConfigurationInstance = ...
if (nonConfigurationInstance != null) {
	cast = nonConfigurationInstance.cast;
	((CastManager_) cast).rebind(this);
}
```
it would make more sense to change the order as for a fresh activity fields are null (if not otherwise assigned) by default.
```java
// nonConfigurationInstance = ...
if (nonConfigurationInstance != null) {
	cast = nonConfigurationInstance.cast;
	((CastManager_) cast).rebind(this);
}
if (cast == null) {
	cast = CastManager_.getInstance_(this);
}
```

WonderCsabo:
We should move it before `BeanHandler`, shouldn't we?

PR #1662, opened 2015-12-18
1660 move NonConfigurationInstanceHandler before injection handler

dodgex:
see https://github.com/excilys/androidannotations/issues/1660


WonderCsabo:
We should move it before `BeanHandler`, shouldn't we?

dodgex:
it is before `BeanHandler` just that AppHandler` is between. first i had it somewhere around [here](https://github.com/dodgex/androidannotations/blob/da062fddb70d66689112e58dc85e17ba03c592e7/AndroidAnnotations/androidannotations-core/androidannotations/src/main/java/org/androidannotations/internal/core/CorePlugin.java#L142) to have it before all injection handler

Issue #1597, opened 2015-10-20
Add @IgnoreWhenViewDestroyed


WonderCsabo:
Now AndroidAnnotations clears injected view fields in `onDestroyView()` by setting each to `null`. However background tasks can be completed later. To avoid NPE-s, currently the client has to check wether the views are `null` or not. We could create a new annotation just like `@IgnoreWhenDetached` to remove that boilerplate. This was already [suggested](https://github.com/excilys/androidannotations/pull/1165#issuecomment-57061775) by @nbelikov.

WonderCsabo:
@yDelouis @dodgex WDYT?

WonderCsabo:
Or maybe we could generalize the `@IgnoreWhenXXX` annotations, like `@IgnoreWhen(DETACHED)` or `@IgnoreWhen(VIEW_DESTROYED)`. I am not sure.

dodgex:
i'd prefer the `@IgnoreWhen(XXX)` approach.

yDelouis:
What would be the generated code ?

WonderCsabo:
I thought about the following:

```java
boolean viewDestroyed;

View onCreateView() {
  ...
  viewDestroyed = false;
}

void onDestroyView() {
  ...
  viewDestroyed = true;
}

void methodIgnoredWhenViewDestroyed() {
  if (!viewDestroyed) {
    super.methodIgnoredWhenViewDestroyed();
  }
}
```

yDelouis:
Okay. Why not.

WonderCsabo:
So can we rename the `@IgnoreWhenDetached` annotation to `@IgnoreWhen` and add an option to it?

yDelouis:
Yes.

shiraji:
I am taking care of this one. Here is TODO

* Create `@IgnoreWhen` that takes DETACHED or VIEW_DESTROYED (or DESTROY_VIEW?)
* Based on the parameter, viewDestroyed is set to true at `onDestroyView` or `onDetach`
* New annotation is accepted the method that has `@UiThread` annotation.

WonderCsabo:
I vote for `VIEW_DESTROYED`.

`viewDestroyed` is only needed if `VIEW_DESTROYED` is used. For `DETACHED`, we can use the old method (`getActivity() == null`), there is no need to create a variable.

`@UiThread` is not mandatory for this.

shiraji:
Could you check if I understand this issue correctly?

# Generated Code

[...]

PR #1664, opened 2015-12-19
Added @IgnoreWhen

shiraji:
see #1597

### DIFF
--- a/AndroidAnnotations/androidannotations-core/androidannotations-api/src/main/java/org/androidannotations/annotations/IgnoreWhen.java
+++ b/AndroidAnnotations/androidannotations-core/androidannotations-api/src/main/java/org/androidannotations/annotations/IgnoreWhen.java
@@ -25,7 +25,9 @@ import java.lang.annotation.Target;
  * When used standalone in an {@link EFragment} or in conjunction with the
  * {@link UiThread} or {@link Background} annotations, the annotated method will
  * be wrapped in an 'if attached' block such that no code will be executed if
- * the {@link EFragment} is no longer bound to its parent activity.
+ * the {@link EFragment} is no longer bound to its parent activity or
+ * <code>DETACHED</code>the {@link EFragment} views are destroyed
+ * <code>VIEW_DESTROYED</code>.
  * </p>
  * <p>
  * Should be used on method that must meet the following criteria
@@ -47,13 +49,13 @@ import java.lang.annotation.Target;
  * ...
  * 
  * 	&#064;UiThread
- * 	&#064;IgnoreWhen
+ * 	&#064;IgnoreWhen(IgnoreWhen.State.DETACHED)
  * 	void killActivity() {
  * 		getActivity().finish();
  * 	}
  * 
  * 
- * 	&#064;IgnoreWhen
+ * 	&#064;IgnoreWhen(IgnoreWhen.State.VIEW_DESTROYED)
  * 	void updateTitle(String title) {
  * 		getActivity().setTitle(title);
  * 	}
@@ -71,4 +73,27 @@ import java.lang.annotation.Target;
 @Target(ElementType.METHOD)
 public @interface IgnoreWhen {
 
+	/**
+	 * The lifecycle state after the method should not be executed.
+	 *
+	 * @return the state that skips method execution
+	 */
+	State value();
+
+	/**
+	 * The lifecycle state after the method should not be executed.
+	 */
+	enum State {
+
+		/**
+		 * Skip execution if the {@link EFragment} is no longer bound to its
+		 * parent activity.
+		 */
+		DETACHED,
+
+		/**
+		 * Skip execution if the {@link EFragment} views are destroyed.
+		 */
+		VIEW_DESTROYED
+	}
 }
--- a/AndroidAnnotations/androidannotations-core/androidannotations-api/src/main/java/org/androidannotations/annotations/IgnoreWhen.java
+++ b/AndroidAnnotations/androidannotations-core/androidannotations-api/src/main/java/org/androidannotations/annotations/IgnoreWhen.java
@@ -47,13 +47,13 @@ import java.lang.annotation.Target;
  * ...
  * 
  * 	&#064;UiThread
- * 	&#064;IgnoredWhenDetached
+ * 	&#064;IgnoreWhen
  * 	void killActivity() {
  * 		getActivity().finish();
  * 	}
  * 
  * 
- * 	&#064;IgnoredWhenDetached
+ * 	&#064;IgnoreWhen
  * 	void updateTitle(String title) {
  * 		getActivity().setTitle(title);
  * 	}
@@ -69,6 +69,6 @@ import java.lang.annotation.Target;
  */
 @Retention(RetentionPolicy.CLASS)
 @Target(ElementType.METHOD)
-public @interface IgnoredWhenDetached {
+public @interface IgnoreWhen {
 
 }
--- a/AndroidAnnotations/androidannotations-core/androidannotations-test/src/test/java/org/androidannotations/test/efragment/MyListFragmentTest.java
+++ b/AndroidAnnotations/androidannotations-core/androidannotations-test/src/test/java/org/androidannotations/test/efragment/MyListFragmentTest.java
@@ -123,6 +123,46 @@ public class MyListFragmentTest {
 	}
 
 	@Test
+	public void notIgnoredAfterOnCreateView() {
+		assertFalse(myListFragment.didExecute);
+		myListFragment.onCreateView(null, null, null);
+		assertFalse(myListFragment.didExecute);
+		myListFragment.ignoreWhenViewDestroyed();
+		assertTrue(myListFragment.didExecute);
+	}
+
+	@Test
+	public void ignoredWhenViewDestroyedForIgnoredMethod() {
+		assertFalse(myListFragment.didExecute);
+		myListFragment.onDestroyView();
+		myListFragment.ignoreWhenViewDestroyed();
+		assertFalse(myListFragment.didExecute);
+	}
+
+	@Test
+	public void notIgnoredAfterFragmentRecreate() {
+		assertFalse(myListFragment.didExecute);
+		myListFragment.onDestroyView();
+		myListFragment.onCreateView(null, null, null);
+		myListFragment.ignoreWhenViewDestroyed();
+		assertTrue(myListFragment.didExecute);
+	}
+
+	@Test
+	public void notIgnoredBeforeDetached() {
+		assertFalse(myListFragment.didExecute);
+		myListFragment.ignored();
+		assertTrue(myListFragment.didExecute);
+	}
+
+	@Test
+	public void notIgnoredBeforeViewDestroyed() {
+		assertFalse(myListFragment.didExecute);
+		myListFragment.ignoreWhenViewDestroyed();
+		assertTrue(myListFragment.didExecute);
+	}
+
+	@Test
 	public void layoutNotInjectedWithoutForce() {
 		View buttonInInjectedLayout = myListFragment.getView().findViewById(R.id.conventionButton);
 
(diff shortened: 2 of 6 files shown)


AndroidAnnotations/androidannotations-core/androidannotations-api/src/main/java/org/androidannotations/annotations/IgnoreWhen.java
```diff
@@ -69,6 +70,29 @@
  */
 @Retention(RetentionPolicy.CLASS)
 @Target(ElementType.METHOD)
-public @interface IgnoredWhenDetached {
+public @interface IgnoreWhen {
 
+	/**
+	 * the state that skips method execution.
```
WonderCsabo: i would write: the lifecycle state after the method should not be executed

WonderCsabo: Also please use sentences, the first word is capitalized etc.

AndroidAnnotations/androidannotations-core/androidannotations/src/main/java/org/androidannotations/internal/core/handler/IgnoreWhenHandler.java
```diff
@@ -53,6 +59,21 @@ public void process(Element element, EFragmentHolder holder) throws Exception {
 		JMethod delegatingMethod = codeModelHelper.overrideAnnotatedMethod(executableElement, holder);
 		JBlock previousMethodBody = codeModelHelper.removeBody(delegatingMethod);
 
-		delegatingMethod.body()._if(invoke(holder.getGeneratedClass().staticRef("this"), "getActivity").ne(_null()))._then().add(previousMethodBody);
+		IgnoreWhen ignoreWhen = element.getAnnotation(IgnoreWhen.class);
+		switch (ignoreWhen.value()) {
+		case VIEW_DESTROYED:
+			String viewDestroyedFieldName = "viewDestroyed" + generationSuffix();
```
WonderCsabo: We usually do this with the holder. Add a new field the holder class, which is initialised and related code is created on demand (lazily, when we call `holder.getViewDestroyedField()`).

AndroidAnnotations/androidannotations-core/androidannotations/src/main/java/org/androidannotations/internal/core/handler/IgnoreWhenHandler.java
```diff
@@ -53,6 +59,21 @@ public void process(Element element, EFragmentHolder holder) throws Exception {
...
+			JFieldVar viewDestroyed = holder.getGeneratedClass().fields().get(viewDestroyedFieldName);
+			if (viewDestroyed == null) {
+				viewDestroyed = holder.getGeneratedClass().field(PRIVATE | VOLATILE, getCodeModel().BOOLEAN, viewDestroyedFieldName);
+				holder.getSetContentViewBlock().assign(viewDestroyed, FALSE);
+				holder.getOnDestroyViewAfterSuperBlock().assign(viewDestroyed, TRUE);
+			}
+			delegatingMethod.body()._if(viewDestroyed.not())._then().add(previousMethodBody);
```
WonderCsabo: Let's remove code duplication here: only the conditional expression depends on the lifecycle state, the other code is the same in both cases.

AndroidAnnotations/androidannotations-core/androidannotations-api/src/main/java/org/androidannotations/annotations/IgnoreWhen.java
```diff
@@ -25,7 +25,8 @@
  * When used standalone in an {@link EFragment} or in conjunction with the
  * {@link UiThread} or {@link Background} annotations, the annotated method will
  * be wrapped in an 'if attached' block such that no code will be executed if
- * the {@link EFragment} is no longer bound to its parent activity.
+ * the {@link EFragment} is no longer bound to its parent activity (DETACHED) or
```
WonderCsabo: Either add `<code>` around the enum value or `@link` to it.

AndroidAnnotations/androidannotations-core/androidannotations/src/main/java/org/androidannotations/holder/EFragmentHolder.java
```diff
@@ -239,7 +239,7 @@ private void setOnDestroyView() {
 		onDestroyViewAfterSuperBlock = body.blockSimple();
 	}
 
-	private JBlock getOnDestroyViewAfterSuperBlock() {
+	public JBlock getOnDestroyViewAfterSuperBlock() {
```
WonderCsabo: This could be private if you move the field creation to the holder as i suggested.

AndroidAnnotations/androidannotations-core/androidannotations-test/src/test/java/org/androidannotations/test/efragment/MyListFragmentTest.java
```diff
@@ -123,6 +123,28 @@ public void ignoredWhenDetachedWorksForIgnoredMethod() {
...
+		assertTrue(myListFragment.didExecute);
+	}
+
+	@Test
+	public void notIgnoredBeforeViewDestroyed() {
+		assertFalse(myListFragment.didExecute);
+		myListFragment.ignoreWhenViewDestroyed();
+		assertTrue(myListFragment.didExecute);
+	}
+
```
WonderCsabo: We need another test case here: call `onDestroyView()`, then `onCreateView()`. In that case the annotated method should not be ignored.

WonderCsabo:
Thanks for this PR! Looks good in general, please address my comments.

WonderCsabo:
A side note: can you use imperative, present tense (*Renamed IgnoredWhenDetached to IgnoreWhen*) for commit messages?

AndroidAnnotations/androidannotations-core/androidannotations-api/src/main/java/org/androidannotations/annotations/IgnoreWhen.java
shiraji: `@link` did not work at that time. I will use `<code>`, instead.

AndroidAnnotations/androidannotations-core/androidannotations-test/src/test/java/org/androidannotations/test/efragment/MyListFragmentTest.java
shiraji: Make sense. I will add it for `onCreateView()`

shiraji:
Didn't know that...thank for letting know that. I will use present tense from this pull request!

AndroidAnnotations/androidannotations-core/androidannotations/src/main/java/org/androidannotations/internal/core/handler/IgnoreWhenHandler.java
shiraji: OK, I will create `IgnoreWhenHolder` and revert the commit that made `getOnDestroyViewAfterSuperBlock` public.

AndroidAnnotations/androidannotations-core/androidannotations/src/main/java/org/androidannotations/internal/core/handler/IgnoreWhenHandler.java
shiraji: Ah! Good point. I will fix that.

AndroidAnnotations/androidannotations-core/androidannotations/src/main/java/org/androidannotations/internal/core/handler/IgnoreWhenHandler.java
WonderCsabo: No need for a new holder. Just add a field in `EFragmentHolder`

dodgex: no need for IgnoreWhenHolder, put it in EFragmentHolder.

WonderCsabo: Wow. :smile:

shiraji: Nice :+1:

dodgex: :D

shiraji:
There you go.

AndroidAnnotations/androidannotations-core/androidannotations-test/src/test/java/org/androidannotations/test/efragment/MyListFragmentTest.java
```diff
@@ -123,6 +127,40 @@ public void ignoredWhenDetachedWorksForIgnoredMethod() {
 	}
 
 	@Test
+	public void notIgnoredAfterOnCreateView() {
+		assertFalse(myListFragment.didExecute);
+		LayoutInflater inflater = mock(LayoutInflater.class);
+		ViewGroup viewGroup = mock(ViewGroup.class);
+		Bundle bundle = mock(Bundle.class);
+		myListFragment.onCreateView(inflater, viewGroup, bundle);
```
WonderCsabo: We could just pass `null` to every parameter here, there is no need for mocking in this case.

WonderCsabo:
I have just one comment. Please fix then squash.

shiraji:
Sure. I will make them all null and squash commits.

AndroidAnnotations/androidannotations-core/androidannotations-test/src/test/java/org/androidannotations/test/efragment/MyListFragmentTest.java
WonderCsabo: Sorry, but i meant another thing:
destroyView
createView
callMethod
assert didExecute true

This is the case when a `Fragment` is returned from the back stack.
The current test is not that interesting, because basically it is the same as `ignoredWhenViewDestroyedForIgnoredMethod`.

You can just fix it without creating a separate commit.

WonderCsabo: OK, actually not the same as, but it is not testing that `onCreateView` does set the `viewDestroyed` field to `true`.

shiraji:
In case you want to check if I change the test case correctly, chw this link.
https://github.com/shiraji/androidannotations/commit/388aafaac8c9988e16680e3811063a081fc15553#diff-5474a5decd9ffab9908ae97c83444b47R128

AndroidAnnotations/androidannotations-core/androidannotations-test/src/test/java/org/androidannotations/test/efragment/MyListFragmentTest.java
shiraji: Ok, I will add the test case.

I don't think `onCreateView` set `viewDestroyed` field to `true`

WonderCsabo: I meant to false. Sorry.

shiraji: No worry. 

I added more test case and squash. Check this method. https://github.com/excilys/androidannotations/pull/1664/files#diff-5474a5decd9ffab9908ae97c83444b47R143

In case you don't want me to commit & squash, let me know.

shiraji:
Updated it https://github.com/shiraji/androidannotations/commit/f22dfbd24e4e24766e957c01ff00c6dc941ade67#diff-5474a5decd9ffab9908ae97c83444b47R143

WonderCsabo:
Thanks. Can you update the wiki as well?

shiraji:
Sure!

Issue #1646, opened 2015-12-09
OnViewChangedNotifier related code generated but not used?


ened:
Say I have a `@EView` annotated class like this:

```
@EView
public class ExtRecyclerView extends RecyclerView {
    public ExtRecyclerView(Context context) {
        super(context);
    }

    public ExtRecyclerView(Context context, AttributeSet attrs) {
        super(context, attrs);
    }

    public ExtRecyclerView(Context context, AttributeSet attrs, int defStyle) {
        super(context, attrs, defStyle);
    }
}
```

Then AA generates this:

```
public final class ExtRecyclerView_
    extends ExtRecyclerView
    implements HasViews
{
    private boolean alreadyInflated_ = false;
    private final OnViewChangedNotifier onViewChangedNotifier_ = new OnViewChangedNotifier();

    public ExtRecyclerView_(Context context) {
        super(context);
        init_();
    }

    public ExtRecyclerView_(Context context, AttributeSet attrs) {
        super(context, attrs);
        init_();
    }

    public ExtRecyclerView_(Context context, AttributeSet attrs, int defStyle) {
        super(context, attrs, defStyle);
        init_();
    }

    public static ExtRecyclerView build(Context context) {
        ExtRecyclerView_ instance = new ExtRecyclerView_(context);
        instance.onFinishInflate();
        return instance;
    }

    /**
     * The alreadyInflated_ hack is needed because of an Android bug
     * which leads to infinite calls of onFinishInflate()
     * when inflating a layout with a parent and using
     * the <merge /> tag.
     * 
     */
    @Override
    public void onFinishInflate() {
        if (!alreadyInflated_) {
            alreadyInflated_ = true;
            onViewChangedNotifier_.notifyViewChanged(this);
        }
        super.onFinishInflate();
    }

    private void init_() {
        OnViewChangedNotifier previousNotifier = OnViewChangedNotifier.replaceNotifier(onViewChangedNotifier_);
        OnViewChangedNotifier.replaceNotifier(previousNotifier);
    }

    public static ExtRecyclerView build(Context context, AttributeSet attrs) {
        ExtRecyclerView_ instance = new ExtRecyclerView_(context, attrs);
        instance.onFinishInflate();
        return instance;
    }

    public static ExtRecyclerView build(Context context, AttributeSet attrs, int defStyle) {
        ExtRecyclerView_ instance = new ExtRecyclerView_(context, attrs, defStyle);
        instance.onFinishInflate();
        return instance;
    }
[...]

PR #1673, opened 2016-01-04
Lazy code generation for view's OnViewChangedNotifier

shiraji:
see #1646

### DIFF
--- a/AndroidAnnotations/androidannotations-core/androidannotations-test/src/main/java/org/androidannotations/test/eview/CustomView.java
+++ b/AndroidAnnotations/androidannotations-core/androidannotations-test/src/main/java/org/androidannotations/test/eview/CustomView.java
@@ -24,7 +24,7 @@ import android.view.View;
 @EView
 public class CustomView extends View {
 
-	public boolean isOnFinishInflateCall;
+	public boolean onFinishInflateCalled;
 
 	public CustomView(Context context) {
 		super(context);
@@ -42,6 +42,6 @@ public class CustomView extends View {
 	protected void onFinishInflate() {
 		super.onFinishInflate();
 
-		isOnFinishInflateCall = true;
+		onFinishInflateCalled = true;
 	}
 }
--- a/AndroidAnnotations/androidannotations-core/androidannotations-test/src/main/java/org/androidannotations/test/eview/CustomView.java
+++ b/AndroidAnnotations/androidannotations-core/androidannotations-test/src/main/java/org/androidannotations/test/eview/CustomView.java
@@ -1,0 +1,47 @@
+/**
+ * Copyright (C) 2010-2015 eBusiness Information, Excilys Group
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may not
+ * use this file except in compliance with the License. You may obtain a copy of
+ * the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed To in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+ * License for the specific language governing permissions and limitations under
+ * the License.
+ */
+package org.androidannotations.test.eview;
+
+import org.androidannotations.annotations.EView;
+
+import android.content.Context;
+import android.util.AttributeSet;
+import android.view.View;
+
+@EView
+public class CustomView extends View {
+
+	public boolean isOnFinishInflateCall;
+
+	public CustomView(Context context) {
+		super(context);
+	}
+
+	public CustomView(Context context, AttributeSet attrs) {
+		super(context, attrs);
+	}
+
+	public CustomView(Context context, AttributeSet attrs, int defStyle) {
+		super(context, attrs, defStyle);
+	}
+
+	@Override
+	protected void onFinishInflate() {
+		super.onFinishInflate();
+
+		isOnFinishInflateCall = true;
+	}
+}
--- a/AndroidAnnotations/androidannotations-core/androidannotations/src/main/java/org/androidannotations/holder/EViewHolder.java
+++ b/AndroidAnnotations/androidannotations-core/androidannotations/src/main/java/org/androidannotations/holder/EViewHolder.java
@@ -169,7 +169,7 @@ public class EViewHolder extends EComponentWithViewSupportHolder {
 		if (onViewChangedBodyAfterInjectionBlock == null) {
 			onViewChangedBodyAfterInjectionBlock = super.getOnViewChangedBodyAfterInjectionBlock();
 			for (OnFinishInflateCallBlock callBlock : onFinishInflateCallBlocks) {
-				callBlock.buildAfterNewInstanceBlock.invoke(callBlock.newInsanceVar, getOnFinishInflate());
+				callBlock.buildAfterNewInstanceBlock.invoke(callBlock.newInstanceVar, getOnFinishInflate());
 				callBlock.copyConstructorBodyBlock.invoke(getInit());
 			}
 		}
@@ -179,12 +179,12 @@ public class EViewHolder extends EComponentWithViewSupportHolder {
 	static class OnFinishInflateCallBlock {
 		JBlock buildAfterNewInstanceBlock;
 		JBlock copyConstructorBodyBlock;
-		JVar newInsanceVar;
+		JVar newInstanceVar;
 
-		OnFinishInflateCallBlock(JBlock buildAfterNewInstanceBlock, JBlock copyConstructorBodyBlock, JVar newInsanceVar) {
+		OnFinishInflateCallBlock(JBlock buildAfterNewInstanceBlock, JBlock copyConstructorBodyBlock, JVar newInstanceVar) {
 			this.buildAfterNewInstanceBlock = buildAfterNewInstanceBlock;
 			this.copyConstructorBodyBlock = copyConstructorBodyBlock;
-			this.newInsanceVar = newInsanceVar;
+			this.newInstanceVar = newInstanceVar;
 		}
 	}
 }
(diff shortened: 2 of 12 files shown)


shiraji:
It's ready for review.

AndroidAnnotations/androidannotations-core/androidannotations/src/test/resources/org/androidannotations/eview/AndroidManifest.xml
```diff
@@ -0,0 +1,30 @@
...
+    the License.
+
+-->
+<manifest package="org.androidannotations.testprocessor"
+          xmlns:android="http://schemas.android.com/apk/res/android"
+          android:versionCode="1"
+          android:versionName="1.0">
+
+    <application>
+        <activity android:name="org.androidannotations.eview.SomeView_"/>
```
dodgex: Why do you add Views as Activies here?

dodgex: an empty `<application/>` tag should be enough.

WonderCsabo:
I do not think this is a good approach. It is not enough to check if the current class has `@AfterViews` in i, for example there can be `@AfterView`s in the superclass.
We should decouple this somehow from `@AfterView`s... Or if we cannot, maybe we should generate this code by triggering explicitly from the `AfterViewsHandler`? I am not sure.

AndroidAnnotations/androidannotations-core/androidannotations/src/test/resources/org/androidannotations/eview/AndroidManifest.xml
shiraji: Ahh...thanks I will replace that.

dodgex:
i think the generation of this code should be triggered by AfterViewsHandler (and View(s)ByIdHandler if necessary?) as soon as it is required.

read: do not let the holder decide if it needs this code, but let it wait until it got told that it needs it.

shiraji:
OK, I will change to trigger AfterViewsHandler.

Are you guys talking about the following code?

```java
			if (hasAfterViews) {
				JVar newCall = staticHelper.body().decl(narrowedGeneratedClass, "instance", newInvocation);
				staticHelper.body().invoke(newCall, getOnFinishInflate());
				staticHelper.body()._return(newCall);
				body.invoke(getInit());
			} else {
				staticHelper.body()._return(newInvocation);
			}
```

Or whole generating process in constructor of EViewHolder?

dodgex:
about the specific code that is used to generate the `OnViewChangedNotifier` related code.

shiraji:
OK, I think I am not in right track.

Here is what I think

* Stop generating OnViewChangedNotifier and its related in `EViewHolder` constructor
* Generate OnViewChangedNotifier and its related using `EViewHolder` in `AfterViewsHandler#process`
* If no class generate `EView`'s constructor and build methods, `EViewHander` generate them.

WonderCsabo:
Why should `EViewHandler` generate this?

Also, the generation logic should still be in the holder, but there should be a call from the handler to the holder to generate the needed code (if it was not generated before). Like we all do with the lazy getter methods.

shiraji:
`EViewHolder` generate this build methods

```java
    public static ExtRecyclerView build(Context context, AttributeSet attrs) {
        ExtRecyclerView_ instance = new ExtRecyclerView_(context, attrs);
        return instance;
    }

    public static ExtRecyclerView build(Context context, AttributeSet attrs, int defStyle) {
        ExtRecyclerView_ instance = new ExtRecyclerView_(context, attrs, defStyle);
        return instance;
    }
```

`EViewHolder` will have `private List<JBlock> buildAfterNewInstanceBlocks` that help insert `instance.onFinishInflate();` This insertion will be called from  `AfterViewsHandler`

`AfterViewsHandler` will call `EViewHolder#getOnFinishInflate` and `EViewHolder#getInit()`.

WonderCsabo:
Seems to be a good plan.

WonderCsabo:
I wonder maybe we should override `getOnViewChangedBodyAfterInjectionBlock()`?

shiraji:
That's good idea, @WonderCsabo . I used your idea to change the code so that now `AfterViewsHandler` triggers to generate `OnFinishInflate`

AndroidAnnotations/androidannotations-core/androidannotations-test/src/main/java/org/androidannotations/test/eview/CustomView.java
```diff
@@ -0,0 +1,47 @@
...
+import org.androidannotations.annotations.EView;
+
+import android.content.Context;
+import android.util.AttributeSet;
+import android.view.View;
+
+@EView
+public class CustomView extends View {
+
+	public boolean isOnFinishInflateCall;
```
WonderCsabo: Let's rename this to `onFinishInfalteCalled`.

AndroidAnnotations/androidannotations-core/androidannotations/src/main/java/org/androidannotations/holder/EViewHolder.java
```diff
@@ -58,8 +58,11 @@
 	protected JMethod onFinishInflate;
 	protected JFieldVar alreadyInflated;
 
+	private List<OnFinishInflateCallBlock> onFinishInflateCallBlock;
+
 	public EViewHolder(AndroidAnnotationsEnvironment environment, TypeElement annotatedElement) throws Exception {
 		super(environment, annotatedElement);
+		onFinishInflateCallBlock = new ArrayList<>();
```
WonderCsabo: Let's move the initialisation to the declaration.

AndroidAnnotations/androidannotations-core/androidannotations/src/main/java/org/androidannotations/holder/EViewHolder.java
```diff
@@ -58,8 +58,11 @@
 	protected JMethod onFinishInflate;
 	protected JFieldVar alreadyInflated;
 
+	private List<OnFinishInflateCallBlock> onFinishInflateCallBlock;
```
WonderCsabo: This should use a plural name.

AndroidAnnotations/androidannotations-core/androidannotations/src/main/java/org/androidannotations/holder/EViewHolder.java
```diff
@@ -99,9 +102,9 @@ private void createConstructorAndBuilder() {
 			}
 
 			JVar newCall = staticHelper.body().decl(narrowedGeneratedClass, "instance", newInvocation);
-			staticHelper.body().invoke(newCall, getOnFinishInflate());
+			OnFinishInflateCallBlock callBlock = new OnFinishInflateCallBlock(staticHelper.body().blockSimple(), body.blockSimple(), newCall);
```
WonderCsabo: I think we should use `blockVirtual()` here to avoid unnecessary braces around the blocks.

AndroidAnnotations/androidannotations-core/androidannotations/src/main/java/org/androidannotations/holder/EViewHolder.java
```diff
@@ -159,4 +162,28 @@ public JFieldVar getAlreadyInflated() {
 	private void setAlreadyInflated() {
 		alreadyInflated = generatedClass.field(PRIVATE, getCodeModel().BOOLEAN, "alreadyInflated" + generationSuffix(), JExpr.FALSE);
 	}
+
+	@Override
+	public JBlock getOnViewChangedBodyAfterInjectionBlock() {
+		JBlock onViewChangedBodyAfterInjectionBlock = super.getOnViewChangedBodyAfterInjectionBlock();
+
+		for (OnFinishInflateCallBlock callBlock : onFinishInflateCallBlock) {
```
WonderCsabo: This should only happen at the first call to `getOnViewChangedBodyAfterInjectionBlock()`.

AndroidAnnotations/androidannotations-core/androidannotations/src/main/java/org/androidannotations/holder/EViewHolder.java
```diff
@@ -159,4 +162,28 @@ public JFieldVar getAlreadyInflated() {
...
+			callBlock.copyConstructorBodyBlock.invoke(getInit());
+		}
+
+		return onViewChangedBodyAfterInjectionBlock;
+	}
+
+	static class OnFinishInflateCallBlock {
+		JBlock buildAfterNewInstanceBlock;
+		JBlock copyConstructorBodyBlock;
+		JVar newInsanceVar;
```
WonderCsabo: Typo.

WonderCsabo:
I think, we should generate the calls to `init_()` in the constructors in `setInit()` as well. Currently these calls will be never added if `@AfterViews` is not used in the class, but for example `@StringRes` is.

shiraji:
Is `init_()` empty method if there is no `@AfterViews`?

AndroidAnnotations/androidannotations-core/androidannotations/src/main/java/org/androidannotations/holder/EViewHolder.java
shiraji: I fixed it for both of parameter. Didn't notice there is `blockVirtual()`.

WonderCsabo:
OK, actually not just `@AfterViews` needs this, but also `@ViewById` and co. I think we should override `setOnViewChanged()` instead.

shiraji:
This is ready for review. I should've mention this after push the change.

WonderCsabo:
The code is getting more complicated. :( Why do you have to change `ViewNotifierHelper`?

shiraji:
The changes at `ViewNotifierHelper` is for `init_()` method. They are not related to your [last comment](https://github.com/excilys/androidannotations/pull/1673#issuecomment-169946179).

PR #1686, opened 2016-01-27
Fix bug with calculation of serial remaining delay

### LINKED ISSUE #1685: @Background: BackgroundExecutor delay is calculated wrongly. Task is never called.
There seems to be a bug in `BackgroundExecutor` class. The `remainingDelay` of the next task (same serial) is set wrongly.

**To reproduce:**

```java
@EActivity
public class MainActivity extends AppCompatActivity {

    private static final String TAG = "APP";

    @Override
    protected void onCreate(Bundle savedInstanceState) {
        super.onCreate(savedInstanceState);
        setContentView(R.layout.activity_main);

        firstTaskNoDelay();
        secondTaskWithDelay();
    }

    @Background(serial = "serial1")
    void firstTaskNoDelay() {
        Log.d(TAG, "Executing task 1");
        try {
            Thread.sleep(6000);
        } catch (InterruptedException e) {
            e.printStackTrace();
        }
    }

    @Background(serial = "serial1", delay = 1000)
    void secondTaskWithDelay() {
        Log.d(TAG, "Executing task 2"); // <-- this is not called (delayed by a huge time)
    }
}
```

There are two tasks with same serial. First one with no delay, second one with delay 1000. The calculation of `next.remainingDelay` is incorrect which causes the second task to be delayed by a huge time.

**Buggy method:**
There is issue when casting long to int.
```java
private void postExecute() {
		/* nothing to do */
		return;
	}
	CURRENT_SERIAL.set(null);
		/* execution complete */
		TASKS.remove(this);

		if (serial != null) {
			Task next = take(serial);
			if (next != null) {
					// =============  BUG HERE  ============= 
					next.remainingDelay = Math.max(0, (int) (targetTimeMillis - System.currentTimeMillis()));
					/* 
					 * targetTimeMillis => 0 
				     * System.currentTimeMillis() => 1453870701164l
				     * (int) (targetTimeMillis - System.currentTimeMillis()) => ((int) (0 - 1453870701164l)) => 2123212180
				     * next.remainingDelay =>  2123212180  <-- WRONG
				     */
				}
			}
		}
	}
}
```

**Proposed Fix:**
```java
next.remainingDelay = Math.max(0l, targetTimeMillis - System.currentTimeMillis());
```

**Update**:
Interestingly this bug will start to happen for all users from today (Tue Jan 26 2016 22:13:49 UTC time) onwards as the `System.currentTimeMillis()` value exceeds `1453846429696l`. 
```
(int)(0 - 1453846429696l) = -2147483648 // <-- negative (so no bug)
(int)(0 - 1453846429697l) = 2147483647  // <-- positive (bug starts to happen)
```

Let me know if I have misunderstood anything.

Regards
-Amulya

WonderCsabo:
Fixes #1685.

### DIFF
--- a/AndroidAnnotations/androidannotations-core/androidannotations-api/src/main/java/org/androidannotations/annotations/Background.java
+++ b/AndroidAnnotations/androidannotations-core/androidannotations-api/src/main/java/org/androidannotations/annotations/Background.java
@@ -191,7 +191,7 @@ public @interface Background {
 	 *
 	 * @return the delay of the execution
 	 */
-	int delay() default 0;
+	long delay() default 0;
 
 	/**
 	 * Serial execution group.
--- a/AndroidAnnotations/androidannotations-core/androidannotations-api/src/main/java/org/androidannotations/api/BackgroundExecutor.java
+++ b/AndroidAnnotations/androidannotations-core/androidannotations-api/src/main/java/org/androidannotations/api/BackgroundExecutor.java
@@ -93,7 +93,7 @@ public final class BackgroundExecutor {
 	 *             if the current executor set by {@link #setExecutor(Executor)}
 	 *             does not support scheduling
 	 */
-	private static Future<?> directExecute(Runnable runnable, int delay) {
+	private static Future<?> directExecute(Runnable runnable, long delay) {
 		Future<?> future = null;
 		if (delay > 0) {
 			/* no serial, but a delay: schedule the task */
@@ -360,7 +360,7 @@ public final class BackgroundExecutor {
 	public static abstract class Task implements Runnable {
 
 		private String id;
-		private int remainingDelay;
+		private long remainingDelay;
 		private long targetTimeMillis; /* since epoch */
 		private String serial;
 		private boolean executionAsked;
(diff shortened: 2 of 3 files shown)


yDelouis:
Code is okay IMHO.

dodgex:
maybe use long for all delay related places to be consistent? ( e.g. [BackgroundExecutor.java#L383](https://github.com/WonderCsabo/androidannotations/blob/1685_serialDelayBug/AndroidAnnotations/androidannotations-core/androidannotations-api/src/main/java/org/androidannotations/api/BackgroundExecutor.java#L383) and possibly the annotation itself?)

otherwise it looks okay for me too. :)

WonderCsabo:
@dodgex good point, i also thought about that.

WonderCsabo:
@dodgex updated.

dodgex:
you missed the [handler](https://github.com/excilys/androidannotations/blob/4e2bccfca60d3875dd22d157f5ff88e84a2dfe8a/AndroidAnnotations/androidannotations-core/androidannotations/src/main/java/org/androidannotations/internal/core/handler/BackgroundHandler.java#L73)

WonderCsabo:
Yep, you are right. :( Updated again.

dodgex:
i think i should be okay now. good work :+1:

dodgex:
@WonderCsabo @yDelouis i think this is something that we should hotfix for 3.x too. as @amulyakhare has pointed out that this issue is affecting all apps that use this combination and 4.0 is a beaking change so there is no drop-in update possible (at least not recommendable).

PR #1690, opened 2016-02-04
fix possible deadlock in serialized background

### LINKED ISSUE #1689: BackgroundExecutor may cause deadlock for serial tasks
First of all, this issue only happens when running robolectric test because I override the Executor, but it can happen in production if tasks are executed too fast.

I have the following executor, that simple execute tasks in serial for test purpose:

```
public class UIThreadExecutor implements Executor{
    
    @Override
    public void execute(@NonNull Runnable command) {
        try {
            command.run();
        }catch (Exception ignored){
        }
    }
}
```   

And I have the follow test:
```

    @Test
    public void testSerial(){
        
        BackgroundExecutor.setExecutor(new UIThreadExecutor());

        for (int i = 0; i < 10; i++){
            activity.testSerial();
        }
        await().atMost(10, TimeUnit.SECONDS).until(new Callable<Boolean>() {
            @Override
            public Boolean call() throws Exception {
                return  activity.count == 10;
            }
        });
    }
```

The activity testSerial implementation is: 

```
    public static int count = 0;
    @Background(serial = "test")
    public void testSerial(){
        count++;
        System.out.println("count = " + count);
    }
```

If you run this test, it will cause dead lock. Digging in the code I was able find out why.

Look at this code:

```
public static synchronized void execute(Task task) {
		Future<?> future = null;
		if (task.serial == null || !hasSerialRunning(task.serial)) {
			task.executionAsked = true;
			future = directExecute(task, task.remainingDelay);
		}
		if (task.id != null || task.serial != null) {
			/* keep task */
			task.future = future;
			TASKS.add(task);
		}
	}
```

As you see the task is add after the directExecute is called. Let's take a look at the Task clean post executed code:

```
	private void postExecute() {
				/* nothing to do */
				return;
			}
			CURRENT_SERIAL.set(null);
				/* execution complete */
				TASKS.remove(this);

				if (serial != null) {
					Task next = take(serial);
					if (next != null) {
							next.remainingDelay = Math.max(0L, targetTimeMillis - System.currentTimeMillis());
						}
					}
				}
			}
		}
```

As you see in this code we remove the task **after the execution**, but my execution always finish before the Task is added to the stack. Causing a dead lock, because the task is added to to list and never removed.

One possible solution would be simple check if the task is already finished (managed) before add it to the list:   `if ( (task.id != null || task.serial != null) && !task.managed.get()) `

guitcastro:
Fixes #1689.

### DIFF
--- a/AndroidAnnotations/androidannotations-core/androidannotations-api/src/main/java/org/androidannotations/api/BackgroundExecutor.java
+++ b/AndroidAnnotations/androidannotations-core/androidannotations-api/src/main/java/org/androidannotations/api/BackgroundExecutor.java
@@ -133,7 +133,7 @@ public final class BackgroundExecutor {
 			task.executionAsked = true;
 			future = directExecute(task, task.remainingDelay);
 		}
-		if (task.id != null || task.serial != null) {
+		if ((task.id != null || task.serial != null) && !task.managed.get()) {
 			/* keep task */
 			task.future = future;
 			TASKS.add(task);


WonderCsabo:
Really nice catch, thanks!

Issue #1697, opened 2016-02-16
Add instantRun processing parameter


WonderCsabo:
When using *Instant Run*, it changes the user-provided application class to a special one in the full manifest, which application can instrument the bytecode to enable *Instant Run*.

For use this means we cannot validate the application class with `@EApplication`. The current workaround in #1639 is to turn of all validations for components in manifests. However this is not a good solution, because it can lead to more developer errors.

Maybe we could introduce a new processing parameter which only turns off validation for `@EApplication`.

@dodgex @yDelouis WDYT?

dodgex:
maybe it is possible to disable the warning when the Application class for *Instant Run* builds is replaced. at least if it is possible to be sure that it is the instant run class (i.e. if registred application is `com.android.instantrun.Application` do not warn/error)

WonderCsabo:
Hmm, so you are suggesting if the application in the manifest is `com.android.instantrun.Application` then we should just skip the error? That is great, so in that case we will do not have to add the processing option.

dodgex:
`com.android.instantrun.Application` was just an example for a possible name of the class. we first would need to check if it is a name where we can assume that it is actually the Instant Run `Application` class.

but yeah, if that is possible i'd suggest that as it is the less intrusive solution. for normal builds we still validate and for instant run we can ignore this specific error.

WonderCsabo:
Oh, i see. I will try out Instant Run and check the real application class name. I wanted to try that anyway. :)

WonderCsabo:
OK, the real classname is `com.android.tools.fd.runtime.BootstrapApplication`.

WonderCsabo:
Good news: it seems the full manifest still contains the original application class name:
```xml
<application
        name="org.androidannotations.sample.MyApp_"
        android:name="com.android.tools.fd.runtime.BootstrapApplication"
        android:icon="@drawable/icon"
        android:label="@string/app_name" >
```

So we can still validate even with instant run!

dodgex:
nice. so they copy the original `andorid:name` attribute to `name` and then set thier own... most probably so that they can read it in thier `BootstrapApplication`. should be an easy fix now?

WonderCsabo:
Yep, I already have a fix in my mind, hopefully I can implement it tonight.

PR #1700, opened 2016-02-21
Support Instant Run application class

WonderCsabo:
Instant Run swaps the application class name in the manifest to its own
bootstrap application, which can instrument the app to use instant Run.
It also adds a new "name" attribute to the manifest, which contains the
original application. When Instant Run is used, we should retrieve this
attribute as the application class name.

Related to #1697.


WonderCsabo:
@dodgex can you check this? I added two tests, and also ran this with Instant Run in Android Studio Beta.

dodgex:
I'll try to test it this week. Wanted to test Instant run anyway. :)

dodgex:
from changes it looks ok. still messing with AS + Instant Run to test it with my app...

dodgex:
while I'm not ok with callling it "Instant Run" as it is incredibly slow (a new color in a layout and a tiny code change take each 40 seconds to launch)... I'm ok with this change.

What i did:
1) enabled instant run -> fail due to missing application
2) added `library 'true'` -> success due to disabled checks
3) removed `library 'true'` -> again failing
4) installed SNAPSHOT build of this PR -> success

PR #1704, opened 2016-02-25
Fix @InstanceState calling order

### LINKED ISSUE #1703: @InstanceState injects after @AfterInject methods
This example code:
```java
@InstanceState
int savedField = -1;

@AfterInject
void afterInject() {

}
```

Can result in this generated code:

```java
private void init_(Bundle savedInstanceState) {
    afterInject();
    restoreSavedInstanceState_(savedInstanceState);
}
```

Note that the `afterInject()` method is called before the instance state injection happens. The order should be the opposite, since we need the injected values in `@AfterInject` methods.

WonderCsabo:
In some scenarios, state restoration was called after `@AfterInject`
methods are called. This commit fixes this issue by moving the restore
call to the dedicated injection block.

Fixes #1703.


dodgex:
code looks good, thank you for the fast fix :-)

Issue #1717, opened 2016-03-03
@Receiver.Extra of type Intent is not considered as a Parcelable extra of the broadcast Intent


tlevavasseur-innovantic:
```
@Receiver(actions = "BROADCAST_NETWORK_ERROR", local = true)
protected void onNetworkError(@Receiver.Extra("EXTRA_RETRY_INTENT") Intent retry) {
```
will consider the Intent param "retry" as the broadcast intent and not the "EXTRA_RETRY_INTENT" Parcelable extra of the broadcast intent.

Doing this don't solve the issue : 
```
@Receiver(actions = "BROADCAST_NETWORK_ERROR", local = true)
protected void onNetworkError(Intent broadcast, @Receiver.Extra("EXTRA_RETRY_INTENT") Intent retry) {
```
and the broadcast intent is passed as param twice.

WonderCsabo:
Yeah, this is an edge case. Please note the `@Receiver.Extra` annotation is only a marker, so you can pass the key as a String. But otherwise the behaviour is just the same without using those annotations.

This is problematic, because currently we the parameters of type `Intent` as the original `Intent`. This is because in 99% of the cases an `Intent` is not used as an extra, and if people adding an `Intent` parameter, they expect the original `Intent`.

We could say that if the `Intent` parameter is marked with the `@Extra` annotation, then we interpret it as an extra, however this would break the consistency between the annotated and not annotated usage, and also complicate our generation code.

@dodgex @yDelouis any thoughts?

dodgex:
I think this should be a doable fix.

WonderCsabo:
@dodgex so do you mean if the `Intent` param has extra annotation, then it is considered as an extra, otherwise the origianal `Intent`?

dodgex:
yeah.

WonderCsabo:
It turned out `@OnActivityResult` already has this behaviour. I will change `@Receiver` and `@ReceiverAction`.

WonderCsabo:
@tlevavasseur-innovantic done!

PR #1718, opened 2016-03-04
Interpret @Extra annotated params as extras with @Receiver{Action}

WonderCsabo:
Implements #1717 .

### DIFF
--- a/AndroidAnnotations/androidannotations-core/androidannotations-test/src/main/java/org/androidannotations/test/receiver/ActivityWithReceiver.java
+++ b/AndroidAnnotations/androidannotations-core/androidannotations-test/src/main/java/org/androidannotations/test/receiver/ActivityWithReceiver.java
@@ -33,6 +33,9 @@ public class ActivityWithReceiver extends Activity {
 
 	public String wifiSsid = null;
 
+	Intent originalIntent;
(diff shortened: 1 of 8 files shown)


AndroidAnnotations/androidannotations-core/androidannotations-test/src/main/java/org/androidannotations/test/receiver/ActivityWithReceiver.java
```diff
@@ -58,4 +61,10 @@ protected void onBroadcastWithTwoActions(Intent intent) {
 			action2Fired = true;
 		}
 	}
+
+	@Receiver(actions =  "org.androidannotations.ACTION_1")
```
dodgex: here is an extra space :)

dodgex: also a string instead of a constant as in ReceiverWithActions. i think you did that to not polute this PR with extracting the other strings?

dodgex:
I think this looks ok just a little extra space that could be removed.

WonderCsabo:
@dodgex thanks, good points. Updated.

Issue #1640, opened 2015-11-25
Android 6.0 release removes support for the Apache HTTP client.


han361769045:
Android 6.0 release removes support for the Apache HTTP client。use the HttpURLConnection class instead。
I think it is necessary to add Androidannotations @HttpUrlConnection instead of @HttpsClient

han361769045:
https://developer.android.com/intl/zh-cn/about/versions/marshmallow/android-6.0-changes.html#behavior-apache-http-client

WonderCsabo:
There is no way to inject `HttpUrlConnection` as it is can be done with `HttpClient`, because `HttpUrlConnection` is obtained via calling `URL.openConnection()` on a specific `URL` instance.

dodgex:
@WonderCsabo but maybe we want to emit a warning or even error when using those with android 6 target?

WonderCsabo:
Yeah, we could add a warning when `targetSdkVersion >= 23`.

WonderCsabo:
And we should issue an error if the `HttpClient` class is not on the compile classpath.

... or we could just remove `@HttpsClient`...

dodgex:
+1 for removing it.

WonderCsabo:
@yDelouis should we remove this annotation?

dodgex:
@WonderCsabo @yDelouis if we decide to remove this, we should do it before 4.0 to not have a post-4.0 breaking change due to removing an annotation

PR #1719, opened 2016-03-04
Deprecate @HttpsClient

WonderCsabo:
Related to #1640.

### DIFF
--- a/AndroidAnnotations/androidannotations-core/androidannotations-api/src/main/java/org/androidannotations/annotations/HttpsClient.java
+++ b/AndroidAnnotations/androidannotations-core/androidannotations-api/src/main/java/org/androidannotations/annotations/HttpsClient.java
@@ -85,9 +85,12 @@ import java.lang.annotation.Target;
  * </pre>
  * 
  * </blockquote>
+ *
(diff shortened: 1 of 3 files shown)


AndroidAnnotations/androidannotations-core/androidannotations/src/main/java/org/androidannotations/internal/core/helper/CoreValidatorHelper.java
```diff
@@ -467,4 +467,10 @@ private boolean methodIsAvailableIn(TypeElement element, String methodName) {
 		}
 		return false;
 	}
+
+	public void isApacheHttpClientClassPresent(Element element, ElementValidation validation) {
+		if (!isClassPresent(CanonicalNameConstants.DEFAULT_HTTP_CLIENT)) {
+			validation.addError(element, "The Apache HTTP library is not found on the classpath!");
```
dodgex: have you tested this with a gradle project targeting sdk 23+?

dodgex:
looks good. when you verified that it works as expected its okay.

PR #1727, opened 2016-03-16
Add Kotlin example

WonderCsabo:


dodgex:
having no deeper of kotlin yet, i think this looks ok. :)

PR #1728, opened 2016-03-16
Make members in Kotlin example protected

WonderCsabo:
In Kotlin, the default visibility (no modifier) is public.
There is no equivalent modifier of package-private, so we must use protected.


dodgex:
:+1:

Issue #1735, opened 2016-03-25
AA doesn't generate ParameterizedTypeReference code when using ResponseEntity<T> as return type


alexthec0d3r:
my build.gradle
```gradle
dependencies {
    //annotations
    apt "org.androidannotations:androidannotations:4.0.0"
    compile "org.androidannotations:androidannotations-api:4.0.0"
    compile "org.androidannotations:rest-spring-api:4.0.0"
    apt "org.androidannotations:rest-spring:4.0.0"
}
```

#1440 is good idea, but If you create interface like:

```java
public interface OperatorsService {
    @Get("/")
    ResponseEntity<List<OperatorInfo>> getOperators();
}
```

then in generated class we still get

```java
public final class OperatorsService_ {
....

    @Override
    public ResponseEntity<List<OperatorInfo>> getOperators() {
        return restTemplate.exchange(rootUrl.concat("/"), HttpMethod.GET, null, List_OperatorInfo.class);
    }

... 
}
```

and Android Studio will show an error: 
```
Error:(42, 37) error: incompatible types
required: ResponseEntity<List<OperatorInfo>>
found:    ResponseEntity<List_OperatorInfo>
```

WonderCsabo:
@alexthec0d3r what is the expected generated code in this case?

alexthec0d3r:
The only one working workaround is to use wrapper class like

```java
public class Operators extends ArrayList<OperatorInfo> {
}
```

I expected from AA that generated code will be 

```java
@Override
public ResponseEntity<List<Operator>> getOperators() {
    return restTemplate.exchange(rootUrl.concat("/"), HttpMethod.GET, null, new ParameterizedTypeReference<List<Operator>>() {});
}
```

but not

```java
@Override
public ResponseEntity<List<Operator>> getOperators() {
    return restTemplate.exchange(rootUrl.concat("/"), HttpMethod.GET, null, List_Operator.class);
}
```

WonderCsabo:
@alexthec0d3r thanks for filing out the bug. Wanna contribute to fix it? :wink:

PR #1736, opened 2016-03-27
Fixes RestTemplate generation with ResponseEntity as return type

simopete:
Implements #1735 

Some test are missing because ParameterizedTypeReference is only available in Spring Android Rest Template 2.0.
I tested it locally.


WonderCsabo:
@alexthec0d3r can you test this branch?

PR #1739, opened 2016-03-30
Log tags need to be shorter than 24 chars according to android docume…

delta2wye:
…ntation

http://developer.android.com/reference/android/util/Log.html#isLoggable(java.lang.String,%20int)


WonderCsabo:
Thanks for this PR! We have the same problem in `TransactionalHandler` and `OrmLiteDaoHandler`? Can you fix those as well? And maybe you can extract this into a helper class as well for example `LogHelper`.

delta2wye:
@WonderCsabo bump

WonderCsabo:
Thanks! Can i ask you to:

- fix the header in the new class (the build is failed because of that)
- please reword your commit messages per the contributor guide in the wiki
- add an overload to the log helper method, which accepts a `GeneratedClassHolder` (so we can remove more code duplication)
- squash the last commits

WonderCsabo:
@delta2wye can you update the PR?

Issue #1746, opened 2016-04-11
Android annotation Encoding Problem


CrazyClownSola:
in case of Android annotation 3.2

if I write like this 

MyActivity.class
```

/**
 * 获取布局
 */
@Extra("menu")
int menu_id;

```

then in MyActivity_.class 
```

public MyActivity_.IntentBuilder_ menu_id(int menu_id) {
       return super.extra(MENU_ID_EXTRA, menu_id);
}

```

But in Android annotation 4.0

the same code

but in MyActivity_.class 
```
  /**
         * 
         * ���ݴ�����
         * @param menu_id
         *     the value for this extra
         * @return
         *     the IntentBuilder to chain calls
         */
        public DetailListActivity_.IntentBuilder_ menu_id(int menu_id) {
            return super.extra(MENU_ID_EXTRA, menu_id);
        }

```
and this cause some error in Android Studio 
like this

```
Error:(188, 12) 错误: 编码UTF-8的不可映射字符
Error:(188, 13) 错误: 编码UTF-8的不可映射字符
Error:(188, 14) 错误: 编码UTF-8的不可映射字符
Error:(188, 16) 错误: 编码UTF-8的不可映射字符
Error:(188, 17) 错误: 编码UTF-8的不可映射字符
Error:(188, 18) 错误: 编码UTF-8的不可映射字符
Error:(188, 19) 错误: 编码UTF-8的不可映射字符
Error:(188, 20) 错误: 编码UTF-8的不可映射字符
8 个错误
:main:compileRetrolambdaDebug
Error:Execution failed for task ':main:compileDebugJavaWithJavac'.
> Compilation failed; see the compiler error output for details.
```

Could I do something to deal with this problem?
Thanks

dodgex:
looks like Android Annotations does not generate UTF-8 files or has some other problems with your UTF-8 comments.

as a workarounnd it should be able to fix the issue when you change your comment to
```
/*
 * 获取布局
 */
@Extra("menu")
int menu_id;
```
note that there is only one star in the first line of the comment. so it is just a multiline comment and no javadoc comment and AA should not try to copy the value to the generated class.


@WonderCsabo any idea how we could fix that?

WonderCsabo:
First, @CrazyClownSola you should try the following:
```groovy
android {
    ...
    compileOptions {
        encoding "UTF-8"
        ...
    }
}
```

CrazyClownSola:
@WonderCsabo  

```
android {
[...]

PR #1749, opened 2016-04-12
Better generated source encoding support

WonderCsabo:
The default encoding is now UTF-8, and the encoding can be configured with the `encoding` annotation processing option.

Implements #1746 .


dodgex:
don't we need to add the new option [here](https://github.com/excilys/androidannotations/blob/develop/AndroidAnnotations/androidannotations-core/androidannotations/src/main/java/org/androidannotations/internal/Options.java#L40)?

WonderCsabo:
You are right!

WonderCsabo:
@dodgex updated.

dodgex:
I think it looks good now. :)

PR #1755, opened 2016-04-17
Fix enabling/disabling annotations

### LINKED ISSUE #1750: AA 4.0.0 @Trace code is not generated
Is @Trace Annotation working in 4.0.0? Somehow I see not effect
this is original class
```
@EBean
class DbConcurrentAccessWorker
{
	
	@Trace(tag = PlaygroundApplication.TAG, level = AndroidConstants.LOG_WARN)
	@Background(id = "worker")
	public void doWork()
	{
		while (Thread.currentThread().isAlive())
		{
			if (random.nextBoolean())
			{
				writeToDb();
			}
			else
			{
				readFromDb();
			}
		}
		Ln.w("Stopping");
	}

	@Trace(tag = PlaygroundApplication.TAG, level = AndroidConstants.LOG_WARN)
	void writeToDb()
	{
		
	}

	@Trace(tag = PlaygroundApplication.TAG, level = AndroidConstants.LOG_WARN)
	void readFromDb()
	{
		
	}
}
```
and this is what I get after aa processing
```
public final class DbConcurrentAccessWorkerImpl
    extends DbConcurrentAccessWorker
{
    private Context context_;

    private DbConcurrentAccessWorkerImpl(Context context) {
        context_ = context;
        init_();
    }

    public static DbConcurrentAccessWorkerImpl getInstance_(Context context) {
        return new DbConcurrentAccessWorkerImpl(context);
    }

    private void init_() {
    }

    public void rebind(Context context) {
        context_ = context;
        init_();
    }

    @Trace(tag = "DE.TOMCOM.FRAMEWORK.PLAYGROUND", level = 5)
    @Override
    public void doWork() {
        BackgroundExecutor.execute(new BackgroundExecutor.Task("worker", 0L, "") {

            @Override
            public void execute() {
                try {
                    DbConcurrentAccessWorkerImpl.super.doWork();
                } catch (final Throwable e) {
                    Thread.getDefaultUncaughtExceptionHandler().uncaughtException(Thread.currentThread(), e);
                }
            }
        }
        );
    }
}
```

the first strange thing is why @Trace is copied to target class. the second - there is no tracing at all.

WonderCsabo:
Fixes #1750 .

@dodgex can you take a look? I wonder, we may want to introduce a new method in the handler interface for this (`isEnabled()`), which has a `true` default implementation in the base classes.

### DIFF
--- a/AndroidAnnotations/androidannotations-core/androidannotations/src/main/java/org/androidannotations/internal/core/CorePlugin.java
+++ b/AndroidAnnotations/androidannotations-core/androidannotations/src/main/java/org/androidannotations/internal/core/CorePlugin.java
@@ -219,9 +219,7 @@ public class CorePlugin extends AndroidAnnotationsPlugin {
 		annotationHandlers.add(new PreferenceClickHandler(androidAnnotationEnv));
 		annotationHandlers.add(new AfterPreferencesHandler(androidAnnotationEnv));
 
-		if (androidAnnotationEnv.getOptionBooleanValue(OPTION_TRACE)) {
-			annotationHandlers.add(new TraceHandler(androidAnnotationEnv));
-		}
+		annotationHandlers.add(new TraceHandler(androidAnnotationEnv, androidAnnotationEnv.getOptionBooleanValue(OPTION_TRACE)));
 
 		/*
 		 * WakeLockHandler must be after TraceHandler but before UiThreadHandler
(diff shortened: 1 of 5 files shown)


dodgex:
i think i'd prefer to have an `isEnabled()` in base annotation handler that by default returns `true` and can be overridden by each handler to return what ever it needs to return. also I'd say to have the option check in the `isEnabled()` and not passing a boolean in the constructor.

```
@Override
public boolean isEnabled() {
    return environment.getOptionBooleanValue(OPTION_TRACE);
}
```

also the check if a processor should run (`if(isEnabled())`) should be at the place where the methods are called. 

```
if(processor.isEnabled()) {
  processor.validate(...);
  processor.process(...);
}
```

or if you do not want to have the check wherever validate and/or process are called add a new method that is called and that one does the enabled check and call the concrete validation/processing. (but i think that is a bit to much)


if you want to keep the current implementation just remove one of the two empty lines in [`TraceHandler`](https://github.com/excilys/androidannotations/pull/1755/files#diff-c05e05ca6fbcf7b549eb1c31b268fbadR63) :P

WonderCsabo:
Thanks @dodgex! I will refactor it for the `isEnabled()` method. I think it is not a breaking change because the default implementation will retain the old behaviour.

WonderCsabo:
Superseded by #1782 .

Issue #1797, opened 2016-06-13
@SystemService FingerprintManager not compiling


nishal:
Using Android Studio with Gradle and get this error when I try to @SystemService inject the FingerprintManager service.

Error:(18, 5) error: Unknown service type: android.hardware.fingerprint.FingerprintManager
Warning:(19, 24) Element fingerprintManager invalidated by SystemServiceHandler

Any help appreciated. TIA

dodgex:
currently `@SystemService` does not support the FingerprintManager. you have to use [`getSystemService()`](https://developer.android.com/reference/android/content/Context.html#getSystemService(java.lang.String)) directly

WonderCsabo:
@nishal API 23 services are not added, yet, but i will add them shorty.

WonderCsabo:
Actually it is a bug, too, we define `android.service.fingerprint.FingerprintManager` instead of `android.hardware.fingerprint.FingerprintManager`.

WonderCsabo:
The bug is due to renaming the hidden service class and exposing it under a new package in API 23 (we supported the old one as a hidden service). The same happened with `NetworkStatsManager`.

PR #1799, opened 2016-06-13
Add new system services from API 23

WonderCsabo:
Related to #1797 .

### DIFF
--- a/AndroidAnnotations/androidannotations-core/androidannotations-test/src/main/java/org/androidannotations/test/ActivityWithServices.java
+++ b/AndroidAnnotations/androidannotations-core/androidannotations-test/src/main/java/org/androidannotations/test/ActivityWithServices.java
@@ -97,6 +97,10 @@ public class ActivityWithServices extends Activity {
 	// @SystemService
 	// CaptioningManager captioningManager;
 
+	// TODO API 23
+	// @SystemService
+	// CarrierConfigManager carrierConfigManager;
+
 	@SystemService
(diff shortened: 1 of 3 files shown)


dodgex:
Have you checked if there are new services for API 24? As the API is final and a release should not be too far from now, we should add them too.

WonderCsabo:
Nope, i did not. Usually i just loo like the GitHub mirror `android_frameworks_base` repo. I think we should add those only if API 24 is released.

WonderCsabo:
BTW @dodgex, did you look at the code? Should i merge this? :)

dodgex:
While I can understand that, I think in should be no harm to have them before. As you can already publish Apps for API24 to play store.

WonderCsabo:
@dodgex i did not know that. I can add those if you point me to the relevant repository. :)

dodgex:
[here](http://android-developers.blogspot.de/2016/06/android-n-apis-are-now-final.html) is the related Android Developers Blog.

i checked from docs and those are the ones i found added with API 24.
[SYSTEM_HEALTH_SERVICE](https://developer.android.com/reference/android/content/Context.html#SYSTEM_HEALTH_SERVICE)
[HARDWARE_PROPERTIES_SERVICE](https://developer.android.com/reference/android/content/Context.html#HARDWARE_PROPERTIES_SERVICE)


about the changes you already made: i think it looks okay.

WonderCsabo:
@dodgex added the new services.

PR #1803, opened 2016-06-15
Fix @Background serial execution

### LINKED ISSUE #1689: BackgroundExecutor may cause deadlock for serial tasks

### LINKED ISSUE #1775: Background with serial not working
The **Background** annotation with **serial** does not work always as expected, in the case, when it's **called from another background thread**.

It is expected that methods annotated with the @Background(serial = X) annotation (provided X is the same for all of them) will run in the order that they were called.

Example
```java
package org.androidannotations.sample;

import org.androidannotations.annotations.Background;
import org.androidannotations.annotations.Click;
import org.androidannotations.annotations.EActivity;

import android.app.Activity;
import android.util.Log;

@EActivity(R.layout.my_activity)
public class MyActivity extends Activity {

    @Click
    void myButtonClicked() {
        bgTaskCallingOthers();
    }

    @Background
    void bgTaskCallingOthers() {
        // we expect that first the long finishes, then the short.
        backgroundSerialLong();
        backgroundSerialShort();
    }

    @Background(serial = "same")
    void backgroundSerialShort() {
        Log.w("MyActivity", "Background serial short finished");
    }

    @Background(serial = "same")
    void backgroundSerialLong() {
        try {
            Thread.sleep(5000);
        } catch (InterruptedException ex) {
            // ignore
        } finally {
            Log.w("MyActivity", "Background serial long finished");
        }
    }
}

```

In the above example, we expect that first the *backgroundSerialLong* method then the second *backgroundSerialShort* method prints it's log message.

But sometimes it's the other way around, resulting in the following log:
```
05-19 08:24:49.480 15669-15839/org.androidannotations.sample W/MyActivity: Background serial short finished
05-19 08:24:54.480 15669-15765/org.androidannotations.sample W/MyActivity: Background serial long finished
```

My first thought, is that sometimes the executor starts to run the task immediately, which in it's first line calls ```managed.getAndSet(true)```, so in *BackgroundExecutor.execute(Task task)*, in the second if, the task is not added to the *TASKS* field.

Because of this, the second background task, doesn't know about the first one.

**UPDATE**
I've tested this by calling the aforementioned two methods from the *myButtonClicked* method (which runs on the main thread) and the same thing happens. Sometimes it works, sometimes it doesn't. Around 2 out of 5 times it fails.

rom1v:
Revert the commit provided to fix the issue https://github.com/excilys/androidannotations/issues/1689 because it broke the serial execution of background tasks. This fixes #1775.

Then provide another commit, that fixes #1689.

See commit messages for details.

### DIFF
--- a/AndroidAnnotations/androidannotations-core/androidannotations-api/src/main/java/org/androidannotations/api/BackgroundExecutor.java
+++ b/AndroidAnnotations/androidannotations-core/androidannotations-api/src/main/java/org/androidannotations/api/BackgroundExecutor.java
@@ -133,7 +133,7 @@ public final class BackgroundExecutor {
 			task.executionAsked = true;
 			future = directExecute(task, task.remainingDelay);
 		}
-		if ((task.id != null || task.serial != null) && !task.managed.get()) {
+		if (task.id != null || task.serial != null) {
 			/* keep task */
 			task.future = future;
 			TASKS.add(task);
--- a/AndroidAnnotations/androidannotations-core/androidannotations-api/src/main/java/org/androidannotations/api/BackgroundExecutor.java
+++ b/AndroidAnnotations/androidannotations-core/androidannotations-api/src/main/java/org/androidannotations/api/BackgroundExecutor.java
@@ -128,15 +128,13 @@ public final class BackgroundExecutor {
 	 *             executor)
 	 */
 	public static synchronized void execute(Task task) {
-		Future<?> future = null;
-		if (task.serial == null || !hasSerialRunning(task.serial)) {
-			task.executionAsked = true;
-			future = directExecute(task, task.remainingDelay);
-		}
 		if (task.id != null || task.serial != null) {
 			/* keep task */
-			task.future = future;
 			TASKS.add(task);
+		}
+		if (task.serial == null || !hasSerialRunning(task.serial)) {
+			task.executionAsked = true;
+			task.future = directExecute(task, task.remainingDelay);
 		}
 	}
 
--- a/AndroidAnnotations/androidannotations-core/androidannotations-test/src/test/java/org/androidannotations/test/ThreadActivityTest.java
+++ b/AndroidAnnotations/androidannotations-core/androidannotations-test/src/test/java/org/androidannotations/test/ThreadActivityTest.java
@@ -163,6 +163,31 @@ public class ThreadActivityTest {
 		}
 	}
 
+	@Test
+	public void serializedBackgroundTasksAsynchronous() throws InterruptedException {
+		/* set an executor with 4 threads */
+		BackgroundExecutor.setExecutor(Executors.newFixedThreadPool(4));
+
+		testSerializedBackgroundTasks();
+	}
+
+	@Test
+	public void serializedBackgroundTasksSynchronous() throws InterruptedException {
+		/* set a synchronous executor */
+		BackgroundExecutor.setExecutor(new Executor() {
+			@Override
+			public void execute(Runnable command) {
+				try {
+					command.run();
+				} catch (RuntimeException ignored) {
+					// ignored
+				}
+			}
+		});
+
+		testSerializedBackgroundTasks();
+	}
+
 	/**
 	 * Verify that serialized background tasks are correctly serialized.
 	 *
@@ -173,13 +198,9 @@ public class ThreadActivityTest {
 	 * Once all tasks have completed execution, verify that the items in the
 	 * list are ordered.
 	 */
-	@Test
-	public void serializedBackgroundTasks() {
+	private void testSerializedBackgroundTasks() {
 		/* number of items to add to the list */
 		final int NB_ADD = 10;
-
-		/* set an executor with 4 threads */
-		BackgroundExecutor.setExecutor(Executors.newFixedThreadPool(4));
 
 		/*
 		 * the calls are serialized, but not necessarily on the same thread, so


LouisCAD:
@rom1v 
Does it means you choose serial over deadlock? I mean, how likely is issue #1689 to happen on a program using multiple task serials with the `@Background` annotation?

I'm a bit afraid as my app needs to be as fail proof as possible to not fail at saving lives.

rom1v:
@LouisCAD 

> Does it means you choose serial over deadlock?

I reverted the fix that broke serial when fixing deadlock, then I provided a fix to fix the deadlock without breaking serial.

> I'm a bit afraid as my app needs to be as fail proof as possible to not fail **at saving lives**.

Then don't use AA or even Android at all.

LouisCAD:
@rom1v 

> fix to fix the deadlock without breaking serial

You mean that only testing is remaining for this PR?

> Then don't use AA or even Android at all.

Android is not that bad at this when you need a [device like this one](http://www.kyoceramobilephone.com/en/s701/) that needs to access network, take pictures, give sensors informations (GPS, accelerometer, etc) and so on. Doing something like this without Android would be reinventing the wheel.

LouisCAD:
@rom1v Thanks for your fix.
Does anyone has an idea about when we can see it released?

BTW, Android is the best option for the product I'm working on, and it's not a medical app, but rather an emergency app for workers, which needs access to network, camera, sensors, and other things found in a smartphone.

WonderCsabo:
Sorry guys, obviously it is my job to merge this, and eventually create a hotfix. Unfortunately i have very busy days (i am working onsite for several weeks), so i cannot guarantee any time. @LouisCAD if this is very important to you, you can build this PR locally and use that in your project.

But i will review this PR and merge this as soon as i find some time.

LouisCAD:
Yes, this is very important for me, I'll follow your suggestion and build this PR locally. Thanks to you @rom1v and @WonderCsabo 

BTW, if you need to make additional tests to prevent such issue in the future, be sure to check [my comment here](https://github.com/excilys/androidannotations/issues/1775#issuecomment-227446540) which includes a simple sample reproducing the issue reliably.

WonderCsabo:
@rom1v would you mind taking look at https://github.com/excilys/androidannotations/issues/1775#issuecomment-227446540 and include as a test case in this PR?

LouisCAD:
I have bad news for everyone there... :disappointed: 
I checked out @rom1v fork, setup maven on my computer, built AA from these sources, updated my `build.gradle` in [my sample project](https://drive.google.com/open?id=0B59c-Qkht7tYcG1GS0wwZ3RrUFE) to use AA 4.1-SNAPSHOT from the local maven like this:
```groovy
repositories {
    mavenLocal()
}

ext {
    AAVersion = '4.1.0-SNAPSHOT'
    supportLibsVersion = '24.0.0'
}
```
I run the sample, and after I touched the FAB to launch the background tasks (which have the same serial if you look at the very simple source code), I saw that the behavior didn't change. I double checked the sources, and yes, the two commits are in the sources I built, so they don't fix the serial not really serial bug... :worried: 

Note that **I didn't updated my sample**, so **don't forget to replace the `ext {...}` portion** with the snippet above before testing against it.

I think that a test like I did will be mandatory. It would need a way to stop the recursive background task I put in the MainActivity though.

LouisCAD:
@rom1v Tell me if you want me to investigate the issue too.

rom1v:
> Tell me if you want me to investigate the issue too.

@LouisCAD yes, please, because currently I have no time.

I think this PR fixes both #1775 and #1689, could you (or anyone) confirm please?
It seems you detected another bug related to *serial* (or is it the same?).

LouisCAD:
@rom1v I think it's the same bug. The original reporter of #1775 had an issue with `serial` not being really serial when called from another `@Background` annotated method. This means that another `@Background` method was running when the others `@Background` methods with `serial` were started, and in my sample, this is exactly what I do, I call an `@Background` method with serial while another `@Background method is running. So **real** issue is that any `@Background` running method **will** make `@Background` with `serial` method calls _not_ serial.

rom1v:
@LouisCAD I simplified your sample to get only 1 activity and remove unrelated stuff, like "Timber":

~~~java
import android.app.Activity;
import android.os.Bundle;
import android.os.Handler;
import android.os.SystemClock;
import android.util.Log;

import org.androidannotations.annotations.Background;
import org.androidannotations.annotations.EActivity;

@EActivity(R.layout.my_activity)
public class MainActivity extends Activity {

    @Override
    protected void onCreate(Bundle savedInstanceState) {
        super.onCreate(savedInstanceState);
        doDummyWork();
        new Handler().postDelayed(new Runnable() {
            @Override
            public void run() {
                for (int i = 0; i < 10; ++i) {
                    concurrencyTest();
                }
            }
        }, 4000);
    }

    @Background
    void doDummyWork() {
        SystemClock.sleep(500);
        doDummyWork();
    }

    @Background(serial = "concurrency")
    void concurrencyTest() {
        long threadId = Thread.currentThread().getId();
        Log.d("ConcurrencyTest", "Hello from " + threadId);
        SystemClock.sleep(2000);
        Log.d("ConcurrencyTest", "Goodbye from " + threadId);
    }
}
~~~

Indeed, on the current AA version, the result is:

~~~
D/ConcurrencyTest(26023): Hello from 241
D/ConcurrencyTest(26023): Hello from 232
D/ConcurrencyTest(26023): Hello from 242
D/ConcurrencyTest(26023): Hello from 240
D/ConcurrencyTest(26023): Hello from 239
D/ConcurrencyTest(26023): Hello from 244
D/ConcurrencyTest(26023): Hello from 243
D/ConcurrencyTest(26023): Hello from 236
D/ConcurrencyTest(26023): Goodbye from 241
D/ConcurrencyTest(26023): Goodbye from 232
D/ConcurrencyTest(26023): Hello from 232
D/ConcurrencyTest(26023): Goodbye from 242
D/ConcurrencyTest(26023): Goodbye from 240
D/ConcurrencyTest(26023): Goodbye from 239
D/ConcurrencyTest(26023): Hello from 239
D/ConcurrencyTest(26023): Goodbye from 244
D/ConcurrencyTest(26023): Goodbye from 243
D/ConcurrencyTest(26023): Goodbye from 236
D/ConcurrencyTest(26023): Goodbye from 232
D/ConcurrencyTest(26023): Goodbye from 239
~~~

But with my fix, the result is:

~~~
D/ConcurrencyTest(27721): Hello from 246
D/ConcurrencyTest(27721): Goodbye from 246
D/ConcurrencyTest(27721): Hello from 244
D/ConcurrencyTest(27721): Goodbye from 244
D/ConcurrencyTest(27721): Hello from 248
D/ConcurrencyTest(27721): Goodbye from 248
D/ConcurrencyTest(27721): Hello from 248
D/ConcurrencyTest(27721): Goodbye from 248
D/ConcurrencyTest(27721): Hello from 241
D/ConcurrencyTest(27721): Goodbye from 241
D/ConcurrencyTest(27721): Hello from 241
D/ConcurrencyTest(27721): Goodbye from 241
D/ConcurrencyTest(27721): Hello from 249
D/ConcurrencyTest(27721): Goodbye from 249
D/ConcurrencyTest(27721): Hello from 246
D/ConcurrencyTest(27721): Goodbye from 246
D/ConcurrencyTest(27721): Hello from 246
D/ConcurrencyTest(27721): Goodbye from 246
D/ConcurrencyTest(27721): Hello from 245
D/ConcurrencyTest(27721): Goodbye from 245
~~~

Did you execute `mvn install` from the `AndroidAnnotations` directory after having checked out this PR?

LouisCAD:
@rom1v Yes, I double checked. Did you try with my original version, just editing the `build.gradle` to use your version?

WonderCsabo:
@LouisCAD to make sure you are using the right version, you can rename the AA version and install that. Sometimes i used AA, because maven not resolved the right same snapshot version. Also double check you check out and build the right branch.
```sh
mvn versions:set -DnewVersion=4.1.0-SOMETHING
```

LouisCAD:
@rom1v I already double checked I checkout the right branch yesterday, in both GitHub desktop and in command line. I tried replacing with the code you ( @rom1v ) just posted, and had the same output as you. I retried with my original code, touched the FAB, and I didn't have the issue though! Sounds good for the fix, but bad for my mental integrity lol! I really don't understand why It's working today while it was not yesterday. I didn't rebuilt or edited anything apart from trying @rom1v snippet and reverting back to retry. My hypothesis are that either I didn't redeployed the app properly, or maybe an Instant Run issue which may not have invalidated some cached stuff. I'll check try on my bigger non sample project with much more `@Background` usages, and tell you if it's working properly.

LouisCAD:
So, the issue is likely to be related to Instant Run. I replaced the dependency in my big project to use the SNAPSHOT from this PR, and the issue still appeared after I:

- Performed a gradle sync
- Pushed the run button to run the app on the device.

I hit `Clean an Rerun 'app'` item from the `Run` menu in Android Studio, it got redeployed to the device, and the issue is gone. I tried multiple times, and I can confirm that there was an issue that this PR solves, and there's in issue in Instant Run or something in Android Studio which doesn't redeploy updated libraries.

So although a test to check that serial is really serial would be welcome, I can say that this PR seems safe to merge and deployed as a hot fix as @WonderCsabo told earlier.

Sorry for the wrong report, and thanks a lot to you @rom1v!

WonderCsabo:
Thanks for cross-testing @LouisCAD !

Yep, Instant Run can be tricky... Actually turned it off, because it not deployed the code changes correctly multiple times, and i debugged for half hours, after realised that the old code is used...

noinnion:
is the bug fixed in PR, could somebody merge it?

WonderCsabo:
Hopefully i can do a release in the upcoming week. However this is an open-source project, so it could be built by anyone from this branch.
Now it is merged, so the snapshot can be used if needed.

LouisCAD:
Fine! I guess this will solve @noinnions problem

PR #1822, opened 2016-07-25
Update docs to include @Path annotation

lukesleeman:
As part of 4.0.0 you now need to include the @Path annotation.  This change updates the @Get docs to include that


WonderCsabo:
Nice catch, thanks very much!

I know you just used the GH editor, but please next time try to use our branch convention with manual branching.

PR #1823, opened 2016-07-25
Update docs to include @Path annotation

lukesleeman:
As part of 4.0.0 all rest method params now need annotations.  This adds the @Path annotation to the docs for @Post


WonderCsabo:
Nice catch, thanks!

Issue #1860, opened 2016-10-02
Reading the Meta-Data information from AndroidManifest


smaugho:
For the creation of some plugins, it is very useful to be able to read the Meta-Data information from the AndroidManifest file.

For instance, in the implementation of a plugin that integrates with Facebook SDK, it is very useful to validate the correct configuration of the Application:

```java

		final String FACEBOOK_APP_ID_NAME = "com.facebook.sdk.ApplicationId";
		final String facebookAppIdValue = getEnvironment().getAndroidManifest().getMetaDataQualifiedNames().get(FACEBOOK_APP_ID_NAME);
		if (facebookAppIdValue == null) {
			valid.addError( 
					"Your application must include the meta-data named " + FACEBOOK_APP_ID_NAME + "." + 
					"\nEx. <meta-data android:name=\"com.facebook.sdk.ApplicationId\"" +
					"\n				android:value=\"<Your Facebook App Id>\"/>"
				);
		}
		
		final String FACEBOOK_LOGIN_ACTIVITY_NAME = "com.facebook.LoginActivity";
		if (!getEnvironment().getAndroidManifest().getComponentQualifiedNames().contains(FACEBOOK_LOGIN_ACTIVITY_NAME)) {
			valid.addError( 
					"Your application must declare " + FACEBOOK_LOGIN_ACTIVITY_NAME + " activity." + 
					"\nEx. <activity android:name=\"com.facebook.LoginActivity\"/>"
				);
		}

```

I did already an implementation for this feature, adding a getMetaDataQualifiedNames method to the AndroidManifest  https://github.com/excilys/androidannotations/commit/86f6f18dfcbdf0407460f2296c72c925ddc5d742

There's many 3rd parties libraries that check for Meta-Data information, this would facilitate the integration and validations with such libraries.

WonderCsabo:
Feel free to open a pull request with this feature. Thanks.

PR #1861, opened 2016-10-02
Reading metadata from manifest

smaugho:
#1860 .

### DIFF
--- a/AndroidAnnotations/androidannotations-core/androidannotations/src/main/java/org/androidannotations/helper/AndroidManifest.java
+++ b/AndroidAnnotations/androidannotations-core/androidannotations/src/main/java/org/androidannotations/helper/AndroidManifest.java
@@ -16,12 +16,15 @@
 package org.androidannotations.helper;
 
 import java.util.Collections;
+import java.util.HashMap;
 import java.util.List;
+import java.util.Map;
 
 public final class AndroidManifest {
 
 	private final String applicationPackage;
 	private final List<String> componentQualifiedNames;
+	private final Map<String, String> metaDataQualifiedNames;
 	private final List<String> permissionQualifiedNames;
 	private final String applicationClassName;
 	private final boolean libraryProject;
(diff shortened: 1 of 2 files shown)


AndroidAnnotations/androidannotations-core/androidannotations/src/main/java/org/androidannotations/helper/AndroidManifest.java
```diff
@@ -71,6 +74,10 @@ public String getApplicationPackage() {
 		return Collections.unmodifiableList(componentQualifiedNames);
 	}
 
+	public Map<String, String> getMetaDataQualifiedNames() {
+		return metaDataQualifiedNames;
```
WonderCsabo: Please use `Collections. unmodifiableMap(metaDataQualifiedNames)` here.

AndroidAnnotations/androidannotations-core/androidannotations/src/main/java/org/androidannotations/internal/helper/AndroidManifestFinder.java
```diff
@@ -291,13 +293,31 @@ private AndroidManifest parse(File androidManifestFile, boolean libraryProject)
...
 		componentQualifiedNames.addAll(providerQualifiedNames);
 
+		Map<String, String> metaDataQualifiedNames = new HashMap<String, String>();
+		NodeList metaDataNodes = documentElement.getElementsByTagName("meta-data");
+		for (int i = 0; i < metaDataNodes.getLength(); i++) {
+			Node node = metaDataNodes.item(i);
+			Node nameAttribute = node.getAttributes().getNamedItem("android:name");
+			Node valueAttribute = node.getAttributes().getNamedItem("android:value");
+
+			if (nameAttribute == null || valueAttribute == null) {
```
WonderCsabo: Can you flatten this `if`?
```java
if (nameAttribute == null && valueAttribute == null) {
  // malformed
} else if (nameAttribute != null) {
  // malformed with name
} else {
  // add it
}
```

smaugho: Hi, I will make a review, and I will check directly the docs https://developer.android.com/guide/topics/manifest/meta-data-element.html, in General, I solved it to the needs of collecting the <meta-data> tags from within the manifest, and following the general structure:

```xml
<meta-data android:name="zoo" android:value="@string/kangaroo" />
```

I used also a Key-Value (a Map) to store and read all the elements. Checking in the docs, there's also a 3rd attribute that can be written in the Manifest, android:resource, never use it before to be honest, but in the general case should be included.

Maybe return a `Map<String, SomeObject>`, with SomeObject being a pair Value, Resource?.

On the other hands, I'm collecting the Metadata at the root level (Application), but it can be applied to:

```xml
    <activity>
    <activity-alias>
    <application>
    <provider>
    <receiver>
    <service>
```

In a general scenario, a reference to the parent tag should be also stored.

Any suggestion about this?
.

smaugho: Checking now the code, I see that also a specific method is called to collect other tasks, I will do the same, to follow the same "rules". This was something I did like 1year ago, so probably the structure of the file was different for then, I will review this too.

smaugho: Ah, about "flatting the if", now checking, I followed same structure for collecting other tags.. Do I flag them all?

```java
private List<String> extractComponentNames(String applicationPackage, NodeList componentNodes) {
		List<String> componentQualifiedNames = new ArrayList<>();

		for (int i = 0; i < componentNodes.getLength(); i++) {
			Node activityNode = componentNodes.item(i);
			Node nameAttribute = activityNode.getAttributes().getNamedItem("android:name");

			String qualifiedName = manifestNameToValidQualifiedName(applicationPackage, nameAttribute);

			if (qualifiedName != null) {
				componentQualifiedNames.add(qualifiedName);
			} else {
				if (nameAttribute != null) {
					LOGGER.warn("A class activity declared in the AndroidManifest.xml cannot be found in the compile path: [{}]", nameAttribute.getNodeValue());
				} else {
					LOGGER.warn("The {} activity node in the AndroidManifest.xml has no android:name attribute", i);
				}
			}
		}
		return componentQualifiedNames;
	}
```

WonderCsabo: OOps, i missed the already existing `if`s. In this case, i think you did the right thing, this way the code is consistent. We can refactor all `if`s later.

WonderCsabo: i think `documentElement.getElementsByTagName()` will traverse this recursively and return all `<meta-data>` in components as well, doesn't it?

smaugho:
Hi, I updated the requested changes, and merged all into a single commit.

AndroidAnnotations/androidannotations-core/androidannotations/src/main/java/org/androidannotations/internal/helper/AndroidManifestFinder.java
smaugho: Yes here it returns all...

WonderCsabo:
Thanks for your work!

PR #1878, opened 2016-10-16
Clarified javadocs in @SharedPrefs with backticks

ironjan:
```
Previously, "<p><b>Defaults to</b>: </p>" was generated for string fields with
an empty default value. This commit inserts back-ticks so that the default
value in the javadoc is easier to be identified:

 * "<p><b>Defaults to</b>: ""</p>" for empty default strings
 * "<p><b>Defaults to</b>: "value"</p>" for @DefaultString("value")
```

I had to wrap this in code tags because the b-tags made the message unreadable

### DIFF
--- a/AndroidAnnotations/androidannotations-core/androidannotations/src/main/java/org/androidannotations/holder/SharedPrefHolder.java
+++ b/AndroidAnnotations/androidannotations-core/androidannotations/src/main/java/org/androidannotations/holder/SharedPrefHolder.java
@@ -1,5 +1,6 @@
 /**
  * Copyright (C) 2010-2016 eBusiness Information, Excilys Group
+ * Copyright (C) 2016 the AndroidAnnotations project
  *
  * Licensed under the Apache License, Version 2.0 (the "License"); you may not
  * use this file except in compliance with the License. You may obtain a copy of
@@ -35,6 +36,7 @@ import org.androidannotations.api.sharedpreferences.IntPrefEditorField;
 import org.androidannotations.api.sharedpreferences.LongPrefEditorField;
 import org.androidannotations.api.sharedpreferences.SharedPreferencesHelper;
 import org.androidannotations.api.sharedpreferences.StringPrefEditorField;
+import org.androidannotations.api.sharedpreferences.StringPrefField;
 import org.androidannotations.api.sharedpreferences.StringSetPrefEditorField;
 import org.androidannotations.helper.CanonicalNameConstants;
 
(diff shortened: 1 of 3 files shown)


dodgex:
Hello @ironjan,

thank you for this PR.

I have some points:
  - could you please add `Copyright (C) 2016 the AndroidAnnotations project` to the copyright header (as shown [here](https://github.com/dodgex/androidannotations/blob/47e7785d9aa1260ae9dd65e1de1760c6835bfedc/AndroidAnnotations/androidannotations-core/androidannotations-api/src/main/java/org/androidannotations/api/bundle/BundleHelper.java#L3)) to all your modified files? 
  - could you please check if this is the same for JavaDoc generation of `@Extra` and `@FragmentArg`? (added in [this](https://github.com/androidannotations/androidannotations/pull/1526) PR)


thanks

dodgex:
okay. ignroe the 2nd point. I just realized that we are not adding a value in those javadoc comments.

WonderCsabo:
Thanks for the PR! IMHO, it would be much cleaner, if you would add quotes, and only around string default values.

ironjan:
Thanks for the feedback. I'll add the copyright and adapt this to add quotes only around string default values.

ironjan:
I'm not sure if I chose the correct way to check for StringFields. If the changes look good, I'll rebase this PR.

dodgex:
@ironjan could you please add the update copyright in SharedPrefWithJavaDoc.java`?

dodgex:
@ironjan also it would be good to rebase your branch with the latest changes from our develop as it has some changes to the `pom.xml` that allow the build pass on travis due to the license header changes.

ironjan:
Copyright notice added and branch rebases.

AndroidAnnotations/androidannotations-core/androidannotations/src/main/java/org/androidannotations/holder/SharedPrefHolder.java
```diff
@@ -118,8 +119,14 @@ private void createEditMethod() {
 	public void createFieldMethod(Class<?> prefFieldHelperClass, IJExpression keyExpression, String fieldName, String fieldHelperMethodName, IJExpression defaultValue,
 			String docComment, String defaultValueStr) {
 		JMethod fieldMethod = generatedClass.method(PUBLIC, prefFieldHelperClass, fieldName);
+
 		if (defaultValueStr != null) {
-			fieldMethod.javadoc().append("<p><b>Defaults to</b>: " + defaultValueStr + "</p>\n");
+			boolean isStringPrefField = "stringField".equals(fieldHelperMethodName);
```
WonderCsabo: Maybe a little bit nicer check would be `prefFieldHelperClass == StringSetPrefField.class`. So we at least do not rely on magic string constants.

AndroidAnnotations/androidannotations-core/androidannotations/src/main/java/org/androidannotations/holder/SharedPrefHolder.java
```diff
@@ -118,8 +119,14 @@ private void createEditMethod() {
...
+			if (isStringPrefField) {
+				fieldMethod.javadoc().append("<p><b>Defaults to</b>: \"" + defaultValueStr + "\"</p>\n");
```
WonderCsabo: Can you avoid the code duplication? Extract the only part which is different, to a variable.

For example:
```java
String defaultValueJavaDoc;

if (isStringPrefField) {
  defaultValueJavaDoc = "\"" + defaultValueStr + "\"";
} else {
  ...
}

fieldMethod.javadoc()...
```

AndroidAnnotations/androidannotations-core/androidannotations/src/main/java/org/androidannotations/internal/core/handler/SharedPrefHandler.java
```diff
@@ -1,5 +1,6 @@
 /**
  * Copyright (C) 2010-2016 eBusiness Information, Excilys Group
+ * Copyright (C) 2016 the AndroidAnnotations project
```
WonderCsabo: Remove this line, as you did not edited changed file.

AndroidAnnotations/androidannotations-core/androidannotations/src/main/java/org/androidannotations/holder/SharedPrefHolder.java
dodgex: Shouldnt it be `StringPrefField.class` instead of `StringSetPrefField.class`

WonderCsabo: Of course!

ironjan:
@WonderCsabo I applied the suggestions and squashed everything together.

dodgex:
Thank you! :)

PR #1884, opened 2016-10-26
Update licenses for new ownership

WonderCsabo:
This does the following:

* `maven license:check` passes for all combinations
* `maven license:format` will add the license with only AA project as copyright holder (this should be fine, as mostly new files are formatted for license headers, existing files already have a header. New files should only have the AA project in the header).
* Adding the AA line to the header for changed files is still not enforced by the plugin, it must be enforced manually when reviewing PRs 😢

### DIFF
--- a/AndroidAnnotations/pom.xml
+++ b/AndroidAnnotations/pom.xml
@@ -316,11 +316,14 @@
 				<groupId>com.mycila</groupId>
 				<artifactId>license-maven-plugin</artifactId>
 				<configuration>
-					<skip>true</skip>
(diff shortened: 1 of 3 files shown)


dodgex:
Looks good i think. :)

I just wonder what we do on 2017? until now we updated all header to contain 2017. But that would no longer be correct for the excilys copyright, or am i wrong?

do we add (c) AA or do we just not update those files until we change them? :/

WonderCsabo:
Yes, we should not update the year on excilys part, ever.

dodgex:
Thank you! :)

Issue #1891, opened 2016-11-08
Using @TextChange and @Trace causes compile error.


ytakahashi1981:
Hello.
I found a very very small problem on AA 4.1.0.

```java
@Trace
@TextChange(R.id.text)
void onTextChange(int start) {
   // do something
}
```

This code causes compile error because AA makes this code, the variable named 'start' already declared.

```java
@Override
void onTextChange(final int start) {
  :
    long start = System.currentTimeMillis();
  :
}
```

And I couldn't change the argument name because AA processor says

`Unrecognized parameter name. You can only have start, before, or count parameter name. Try to pick a parameter from the android.text.TextWatcher.onTextChanged() method.`

So I can't use `@TextChange` and `@Trace` on the same method.

Thank you.

dodgex:
Thank you for the report. This one should be easy to fix on our side! :)

dodgex:
I just openend PR #1893 to fix this issue.

WonderCsabo:
Fixed. Thanks @dodgex !

PR #1893, opened 2016-11-12
Rename variables used for @Trace to avoid name collisions

dodgex:
- prefix variables with trace
- postifx variables with underscore

see #1891

### DIFF
--- a/AndroidAnnotations/androidannotations-core/androidannotations/src/main/java/org/androidannotations/internal/core/handler/TraceHandler.java
+++ b/AndroidAnnotations/androidannotations-core/androidannotations/src/main/java/org/androidannotations/internal/core/handler/TraceHandler.java
@@ -1,5 +1,6 @@
 /**
  * Copyright (C) 2010-2016 eBusiness Information, Excilys Group
+ * Copyright (C) 2016 the AndroidAnnotations project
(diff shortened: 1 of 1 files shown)


AndroidAnnotations/androidannotations-core/androidannotations/src/main/java/org/androidannotations/internal/core/handler/TraceHandler.java
```diff
@@ -98,7 +99,7 @@ public void process(Element element, EComponentHolder holder) throws Exception {
 
 		logEnterInvoke.arg(getEnterMessage(method, executableElement));
 		thenBody.add(logEnterInvoke);
-		JVar startDeclaration = thenBody.decl(getCodeModel().LONG, "start", currentTimeInvoke);
+		JVar startDeclaration = thenBody.decl(getCodeModel().LONG, "traceStart_", currentTimeInvoke);
```
WonderCsabo: I think we should use `GENERATION_SUFFIX` instead of hardcoding the underscore into every string.

dodgex:
updated

dodgex:
I think we should not merge until we know what happened with #1905

PR #1911, opened 2016-12-10
Support instance state for views

dodgex:
This PR is based on the [work](https://github.com/simopete/androidannotations/commit/67b55ac1b6592e7326e0ef4a16a786d42c4f9fdb) of @simopete. 

#### What is the goal?
Adding support of `@InstanceState` to custom views.

### What is changed from the original work?
- Update copyright header
- Added a test case

### DIFF
--- a/AndroidAnnotations/androidannotations-core/androidannotations-test/src/main/java/org/androidannotations/test/eview/ViewWithInstanceState.java
+++ b/AndroidAnnotations/androidannotations-core/androidannotations-test/src/main/java/org/androidannotations/test/eview/ViewWithInstanceState.java
@@ -1,0 +1,54 @@
+/**
+ * Copyright (C) 2016 the AndroidAnnotations project
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may not
+ * use this file except in compliance with the License. You may obtain a copy of
+ * the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed To in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+ * License for the specific language governing permissions and limitations under
+ * the License.
+ */
+package org.androidannotations.test.eview;
+
+import org.androidannotations.annotations.EView;
+import org.androidannotations.annotations.InstanceState;
+import org.androidannotations.test.instancestate.MyParcelableBean;
+
+import android.content.Context;
+import android.os.Bundle;
+import android.os.Parcelable;
+import android.view.View;
+
+@EView
+public class ViewWithInstanceState extends View {
+
+	@InstanceState
+	String stateTest = "does it work?";
+
+	String stringState = "does it work?";
+	MyParcelableBean beanState = new MyParcelableBean(0);
+
+	public ViewWithInstanceState(Context context) {
+		super(context);
+	}
+
+	@Override
+	public Parcelable onSaveInstanceState() {
+		Bundle bundle = new Bundle();
+		bundle.putString("stringState", stringState);
+		bundle.putParcelable("beanState", beanState);
+		return bundle;
+	}
+
+	@Override
+	public void onRestoreInstanceState(Parcelable state) {
+		Bundle bundle = (Bundle) state;
+		stringState = bundle.getString("stringState");
+		beanState = bundle.getParcelable("beanState");
+	}
+}
--- a/AndroidAnnotations/androidannotations-core/androidannotations-test/src/test/java/org/androidannotations/test/eview/ViewInstanceStateTest.java
+++ b/AndroidAnnotations/androidannotations-core/androidannotations-test/src/test/java/org/androidannotations/test/eview/ViewInstanceStateTest.java
@@ -1,0 +1,57 @@
+/**
+ * Copyright (C) 2016 the AndroidAnnotations project
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may not
+ * use this file except in compliance with the License. You may obtain a copy of
+ * the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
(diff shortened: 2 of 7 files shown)


AndroidAnnotations/androidannotations-core/androidannotations-test/src/test/java/org/androidannotations/test/eview/ViewInstanceStateTest.java
```diff
@@ -0,0 +1,48 @@
...
+	public void viewInstanceStateSaveTest() {
+		Context context = Robolectric.buildActivity(EmptyActivityWithoutLayout_.class).create().get();
+		ViewWithInstanceState_ savingView = (ViewWithInstanceState_) ViewWithInstanceState_.build(context);
+
+		assertThat(savingView.stateTest).isEqualTo("does it work?");
+
+		savingView.stateTest = "it works!";
+		Parcelable savedInstanceState = savingView.onSaveInstanceState();
+
+		ViewWithInstanceState_ resotringView = (ViewWithInstanceState_) ViewWithInstanceState_.build(context);
```
WonderCsabo: typo

AndroidAnnotations/androidannotations-core/androidannotations-test/src/main/java/org/androidannotations/test/eview/ViewWithInstanceState.java
```diff
@@ -0,0 +1,38 @@
...
+
+import org.androidannotations.annotations.EView;
+import org.androidannotations.annotations.InstanceState;
+
+import android.content.Context;
+import android.util.AttributeSet;
+import android.view.View;
+
+@EView
+public class ViewWithInstanceState extends View {
```
WonderCsabo: I think we should have another test, which asserts we correctly retain the saved instance information from the super class. So there should be a non-AA Android custom View class, which have `onSaveInstanceState()` and `onRestoreInstanceState()`, then an enhanced class should subclass it, and add `@InstanceState`. We should assert that the state restoration in parent works as well.

AndroidAnnotations/androidannotations-core/androidannotations-test/src/main/java/org/androidannotations/test/eview/ViewWithInstanceState.java
```diff
@@ -0,0 +1,38 @@
...
+
+	public ViewWithInstanceState(Context context) {
+		super(context);
+	}
+
+	public ViewWithInstanceState(Context context, AttributeSet attrs) {
```
WonderCsabo: This constructor is not needed.

AndroidAnnotations/androidannotations-core/androidannotations-test/src/main/java/org/androidannotations/test/eview/ViewWithInstanceState.java
dodgex: do we need an extra non aa custom view? shouldn't it be enough to have custom save/restore in this view class?

WonderCsabo: It is enough, if you can check if the state in the superclass is properly saved, however i think you cannot do that. Maybe you can extend from another `View` class, which saves state we can assert against, but the most clean would be writing a non-AA custom `View`.

AndroidAnnotations/androidannotations-core/androidannotations-test/src/test/java/org/androidannotations/test/eview/ViewInstanceStateTest.java
dodgex: fixed

AndroidAnnotations/androidannotations-core/androidannotations-test/src/main/java/org/androidannotations/test/eview/ViewWithInstanceState.java
```diff
@@ -0,0 +1,38 @@
...
+		super(context, attrs);
+	}
+
+	@InstanceState
+	String stateTest = "does it work?";
```
dodgex: fixed

AndroidAnnotations/androidannotations-core/androidannotations-test/src/main/java/org/androidannotations/test/eview/ViewWithInstanceState.java
dodgex: fixed

dodgex:
@Zhuinden not sure what exactly you mean?  could you explain the issue you have and maybe suggest what should happen instead ? Thanks!


> Did you guys ever fix this, or has it been broken for 2.5 years?

If there is an issue with that, I'm sure that it is not fixed yet, as i never heard of it before.

WonderCsabo:
@Zhuinden i think you are not correct. The parent class' state is saved and is put into the bundle, later extracted and restored. It should work well.
```java
    @Override
    public Parcelable onSaveInstanceState() {
        Parcelable instanceState = super.onSaveInstanceState();
        Bundle bundle_ = new Bundle();
        bundle_.putParcelable(INSTANCE_STATE_KEY, instanceState);
        saveInstanceState(bundle_);
        return bundle_;
    }

    private void saveInstanceState(Bundle bundle) {
        bundle.putString("something", something);
    }

    @Override
    public void onRestoreInstanceState(Parcelable state) {
        Bundle bundle_ = ((Bundle) state);
        Parcelable instanceState = bundle_.getParcelable(INSTANCE_STATE_KEY);
        something = bundle_.getString("something");
        super.onRestoreInstanceState(instanceState);
    }
```

Zhuinden:
Ok in that case it should most likely work, albeit there is a chance that extending RecyclerView could have problem if `bundle.setClassLoader(RecyclerView.class.getClassLoader())` isn't called 🤔 

I'd have to test it to know if that happens or not.

However, as you ***do*** save out the super-state, my initial claim was incorrect, thanks

PR #1928, opened 2016-12-30
Updated kotlin example

WeaponMan:
I updated to last eap version of kotlin, because 1.0.5-3 with kapt2 is unable to find AndroidManifest.
I added shared preferences example that is not working with kotlin.

### DIFF
--- a/examples/kotlin/build.gradle
+++ b/examples/kotlin/build.gradle
@@ -1,10 +1,12 @@
 buildscript {
+    ext.kotlin_version = '1.0.6'
+    
     repositories {
         jcenter()
     }
     dependencies {
-        classpath 'com.android.tools.build:gradle:2.1.0'
-        classpath 'org.jetbrains.kotlin:kotlin-gradle-plugin:1.0.2'
+        classpath 'com.android.tools.build:gradle:2.2.3'
+        classpath "org.jetbrains.kotlin:kotlin-gradle-plugin:$kotlin_version"
     }
 }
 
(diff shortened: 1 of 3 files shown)


WonderCsabo:
@WeaponMan thanks for updating the sample. Can you remove the non-working Pref from this PR? I want to merge this, but we should not commit non-working example code. 

We want to send the non-working code to kotlin devs ofc.

dodgex:
also i'm not sure if we want to use kotlin EAP in the sample project?

WonderCsabo:
I agree, we should only use stable versions. Why did you update to EAP?
Again, it is ok in the test branch which we will show to Kotlin devs.

WeaponMan:
I updated to EAP, because kapt2 was unable to find AndroidManifest file.
EAP version containts new kapt3 code that works mostly ok.
But it seams that generated code is not added to classpath of kapt gradle task.
I tried this without plugin kotlin-kapt which brings back the first version of kapt.
The build errors at generating stubs. (Unable to compile Cl class generated by kapt).

I will revert non working stuff and remove eap version from this PR tomorrow.

WeaponMan:
Updated to 1.0.6 which was released on Dec 26. Totally missed that.

examples/kotlin/build.gradle
```diff
@@ -1,10 +1,11 @@
 buildscript {
+    ext.kotlin_version = '1.0.6'
```
WonderCsabo: Add a new line after this.

examples/kotlin/src/main/kotlin/org/androidannotations/gradle/activity/HelloAndroidActivity.kt
```diff
@@ -12,7 +12,7 @@ import org.androidannotations.gradle.R
 import java.util.Date
 
 @EActivity(R.layout.main)
-open public class HelloAndroidActivity : Activity() {
+open class HelloAndroidActivity : Activity() {
```
WonderCsabo: Why did you remove `public`?

WonderCsabo:
@WeaponMan pls next time use a topic branch, not just `develop` as a target branch.

examples/kotlin/src/main/kotlin/org/androidannotations/gradle/activity/HelloAndroidActivity.kt
WeaponMan: Classes with no access modifier are always public in kotlin. There is no need to type public.

WonderCsabo: Good to know, thx.

WeaponMan:
Squashed into one commit.

examples/kotlin/build.gradle
```diff
@@ -1,10 +1,12 @@
...
+    
     repositories {
         jcenter()
     }
     dependencies {
-        classpath 'com.android.tools.build:gradle:2.1.0'
-        classpath 'org.jetbrains.kotlin:kotlin-gradle-plugin:1.0.2'
+        classpath 'com.android.tools.build:gradle:2.2.3'
```
dodgex: when updating versions in this sample we should also update the versions in the gradle sample (and if necessary in the maven samples too).

dodgex:
Thank you for your contribution @WeaponMan .

I added some additional comments. Unfortunately the PR is already merged.

@WonderCsabo could you have a look at those comments? i think at least for the updated versions in the other samples we should have another PR.

examples/kotlin/build.gradle
```diff
@@ -1,10 +1,12 @@
...
+        classpath "org.jetbrains.kotlin:kotlin-gradle-plugin:$kotlin_version"
```
WonderCsabo: This is needed because of string interpolation.

dodgex: okay. do we want to be consistent and update the other lines too? if it is no noticeable performance drawback?

WonderCsabo:
Yes, we should update all samples (include the maven one), to use latest and greatest plugin and dependency versions.

examples/kotlin/build.gradle
WonderCsabo: I think we should use double quotes in all Gradle scripts for strings, yes. It looks better for consistency.

dodgex:
Added new issue for updating the samples https://github.com/androidannotations/androidannotations/issues/1932

PR #1937, opened 2017-01-17
Support for RadioGroup OnCheckedChangeListener in @CheckedChange annotation

gauravsak:
According to the discussion with @WonderCsabo, I have implemented the logic to check if the method annotated with `@CheckedChange` has the first parameter of type `RadioGroup`, then the second parameter (if present) should be of type `int/Integer` to represent `checkedId` else if first parameter is a `CompoundButton`, then the second parameter (if present) should be of type `boolean/Boolean` to represent `isChecked`.

### DIFF
--- a/AndroidAnnotations/androidannotations-core/androidannotations/src/main/java/org/androidannotations/helper/ValidatorParameterHelper.java
+++ b/AndroidAnnotations/androidannotations-core/androidannotations/src/main/java/org/androidannotations/helper/ValidatorParameterHelper.java
@@ -120,7 +120,7 @@ public class ValidatorParameterHelper {
 		}
 
 		public V extendsAnyOfTypes(String... types) {
-			return  param(new ExtendsAnyOfTypesParameterRequirement(types));
+			return param(new ExtendsAnyOfTypesParameterRequirement(types));
 		}
 
 		public V anyType() {
@@ -133,6 +133,10 @@ public class ValidatorParameterHelper {
 
 		public V primitiveOrWrapper(TypeKind primitive) {
 			return param(new PrimitiveOrWrapperParameterRequirement(primitive));
+		}
+
+		public V anyOfPrimitiveOrWrapper(TypeKind... primitives) {
+			return param(new AnyOfPrimitiveOrWrapperParameterRequirement(primitives));
 		}
 
 		public V anyOfTypes(String... types) {
@@ -168,7 +172,6 @@ public class ValidatorParameterHelper {
 		protected void invalidate(ExecutableElement element, ElementValidation validation) {
 			validation.addError("%s can only have the following parameters: " + createMessage(element));
 		}
-
 
 		protected String createMessage(ExecutableElement element) {
 			StringBuilder builder = new StringBuilder();
@@ -442,6 +445,62 @@ public class ValidatorParameterHelper {
 		}
 
 		private String getWrapperType() {
+			switch (type) {
+			case BOOLEAN:
+				return CanonicalNameConstants.BOOLEAN;
+			case INT:
+				return CanonicalNameConstants.INTEGER;
+			case BYTE:
+				return CanonicalNameConstants.BYTE;
+			case SHORT:
+				return CanonicalNameConstants.SHORT;
+			case LONG:
+				return CanonicalNameConstants.LONG;
+			case CHAR:
+				return CanonicalNameConstants.CHAR;
+			case FLOAT:
+				return CanonicalNameConstants.FLOAT;
+			case DOUBLE:
+				return CanonicalNameConstants.DOUBLE;
+			default:
+				throw new IllegalArgumentException("The TypeKind passed does not represent a primitive");
+			}
+		}
+	}
+
+	public class AnyOfPrimitiveOrWrapperParameterRequirement extends BaseParameterRequirement {
+
+		private List<TypeKind> types;
+
+		public AnyOfPrimitiveOrWrapperParameterRequirement(TypeKind... types) {
+			this.types = Arrays.asList(types);
+		}
+
+		@Override
+		protected String description() {
+			StringBuilder builder = new StringBuilder();
+			builder.append("Any of the primitives: [");
+			for (int i = 0; i < types.size() - 1; ++i) {
+				builder.append(getWrapperType(types.get(i))).append(", ");
+			}
+			builder.append(getWrapperType(types.get(types.size() - 1))).append(" ]");
+
+			return builder.toString();
+		}
+
+		@Override
+		public boolean isSatisfied(VariableElement parameter) {
+			TypeMirror elementType = parameter.asType();
+
+			for (TypeKind type : types) {
+				if (elementType.getKind() == type || elementType.toString().equals(getWrapperType(type))) {
+					return true;
+				}
+			}
+			return false;
+		}
+
+		private String getWrapperType(TypeKind type) {
 			switch (type) {
 			case BOOLEAN:
 				return CanonicalNameConstants.BOOLEAN;
--- a/AndroidAnnotations/androidannotations-core/androidannotations/src/main/java/org/androidannotations/internal/core/handler/CheckedChangeHandler.java
+++ b/AndroidAnnotations/androidannotations-core/androidannotations/src/main/java/org/androidannotations/internal/core/handler/CheckedChangeHandler.java
@@ -50,9 +50,9 @@ public class CheckedChangeHandler extends AbstractViewListenerHandler {
 		ExecutableElement executableElement = (ExecutableElement) element;
 		validatorHelper.returnTypeIsVoid(executableElement, validation);
 
-		validatorHelper.param.anyOrder() //
-				.extendsType(CanonicalNameConstants.COMPOUND_BUTTON).optional() //
-				.primitiveOrWrapper(TypeKind.BOOLEAN).optional() //
+		validatorHelper.param.inOrder() //
+				.extendsAnyOfTypes(CanonicalNameConstants.COMPOUND_BUTTON, CanonicalNameConstants.RADIO_GROUP).optional() //
+				.anyOfPrimitiveOrWrapper(TypeKind.BOOLEAN, TypeKind.INT).optional() //
 				.validate(executableElement, validation);
 	}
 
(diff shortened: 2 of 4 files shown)


AndroidAnnotations/androidannotations-core/androidannotations/src/main/java/org/androidannotations/internal/core/handler/CheckedChangeHandler.java
```diff
@@ -36,10 +36,13 @@
...
 import com.helger.jcodemodel.JMod;
 import com.helger.jcodemodel.JVar;
+import com.helger.jcodemodel.AbstractJType;
 
 public class CheckedChangeHandler extends AbstractViewListenerHandler {
 
-	public CheckedChangeHandler(AndroidAnnotationsEnvironment environment) {
+	private String viewElement;
+
+	public CheckedChangeHandler( AndroidAnnotationsEnvironment environment) {
```
dodgex: please remove the whitespace here

AndroidAnnotations/androidannotations-core/androidannotations/src/main/java/org/androidannotations/internal/core/handler/CheckedChangeHandler.java
```diff
@@ -63,16 +66,26 @@ protected void makeCall(JBlock listenerMethodBody, JInvocation call, TypeMirror
...
-
 		for (VariableElement parameter : parameters) {
-			String parameterType = parameter.asType().toString();
 			if (isTypeOrSubclass(CanonicalNameConstants.COMPOUND_BUTTON, parameter)) {
-				call.arg(castArgumentIfNecessary(holder, CanonicalNameConstants.COMPOUND_BUTTON, btnParam, parameter));
-			} else if (parameterType.equals(CanonicalNameConstants.BOOLEAN) || parameter.asType().getKind() == TypeKind.BOOLEAN) {
-				call.arg(isCheckedParam);
-			}
+				viewElement = CanonicalNameConstants.COMPOUND_BUTTON;
+        JVar btnParam = listenerMethod.param(getClasses().COMPOUND_BUTTON, "buttonView");
```
dodgex: please fix the indentation here and the lines below

AndroidAnnotations/androidannotations-core/androidannotations/src/main/java/org/androidannotations/internal/core/handler/CheckedChangeHandler.java
```diff
@@ -63,16 +66,26 @@ protected void makeCall(JBlock listenerMethodBody, JInvocation call, TypeMirror
...
+        call.arg(castArgumentIfNecessary(holder, CanonicalNameConstants.COMPOUND_BUTTON, btnParam, parameter));
+        if (listenerMethod.hasSignature( new AbstractJType[] { getClasses().COMPOUND_BUTTON, getCodeModel().BOOLEAN} )) {
+          JVar isCheckedParam = listenerMethod.param(getCodeModel().BOOLEAN, "isChecked");
+          call.arg( isCheckedParam );
```
dodgex: no extra spaces around the paramter

AndroidAnnotations/androidannotations-core/androidannotations/src/main/java/org/androidannotations/internal/core/handler/CheckedChangeHandler.java
```diff
@@ -88,11 +101,17 @@ protected String getSetterName() {
 
 	@Override
 	protected AbstractJClass getListenerClass(EComponentWithViewSupportHolder holder) {
+		if ( CanonicalNameConstants.RADIO_GROUP.equals( viewElement ) ) {
```
dodgex: no extra spaces

AndroidAnnotations/androidannotations-core/androidannotations/src/main/java/org/androidannotations/internal/core/handler/CheckedChangeHandler.java
```diff
@@ -88,11 +101,17 @@ protected String getSetterName() {
...
+			return getClasses().RADIO_GROUP_ON_CHECKED_CHANGE_LISTENER;
+		}
 		return getClasses().COMPOUND_BUTTON_ON_CHECKED_CHANGE_LISTENER;
 	}
 
 	@Override
 	protected AbstractJClass getListenerTargetClass(EComponentWithViewSupportHolder holder) {
+		if ( CanonicalNameConstants.RADIO_GROUP.equals( viewElement ) ) {
```
dodgex: no extra spaces

AndroidAnnotations/androidannotations-core/androidannotations/src/main/java/org/androidannotations/helper/ValidatorParameterHelper.java
```diff
@@ -465,6 +469,61 @@ private String getWrapperType() {
...
+
+			for ( TypeKind type: types ) {
+				if ( elementType.getKind() == type || elementType.toString().equals( getWrapperType(type) ) )
+				return true;
+			}
+			return false;
+		}
+
+		private String getWrapperType(TypeKind type) {
+			switch ( type ) {
```
WonderCsabo: This is duplicated code.

AndroidAnnotations/androidannotations-core/androidannotations/src/main/java/org/androidannotations/helper/ValidatorParameterHelper.java
```diff
@@ -465,6 +469,61 @@ private String getWrapperType() {
...
+
+		public AnyOfPrimitiveOrWrapperParameterRequirement(TypeKind... types ) {
+			this.types = Arrays.asList(types);
+		}
+
+		@Override
+		protected String description() {
+			StringBuilder builder = new StringBuilder();
+			builder.append("Any of the primitives: [");
+			for (int i = 0; i < types.size() - 1; ++i) {
```
WonderCsabo: Add new line before `for`.

AndroidAnnotations/androidannotations-core/androidannotations/src/main/java/org/androidannotations/helper/ValidatorParameterHelper.java
```diff
@@ -465,6 +469,61 @@ private String getWrapperType() {
...
+				builder.append(getWrapperType(types.get(i))).append(", ");
+			}
+			builder.append(getWrapperType(types.get(types.size() - 1))).append(" ]");
```
WonderCsabo: I think it would be better to move this to do `for` loop and check with an `if` for the last comma. And just add `]` outside the loop.

AndroidAnnotations/androidannotations-core/androidannotations/src/main/java/org/androidannotations/internal/core/handler/CheckedChangeHandler.java
```diff
@@ -50,9 +53,9 @@ public void validate(Element element, ElementValidation validation) {
 		ExecutableElement executableElement = (ExecutableElement) element;
 		validatorHelper.returnTypeIsVoid(executableElement, validation);
 
-		validatorHelper.param.anyOrder() //
-				.extendsType(CanonicalNameConstants.COMPOUND_BUTTON).optional() //
-				.primitiveOrWrapper(TypeKind.BOOLEAN).optional() //
+		validatorHelper.param.inOrder() //
+				.extendsAnyOfTypes(CanonicalNameConstants.COMPOUND_BUTTON, CanonicalNameConstants.RADIO_GROUP).optional() //
+				.anyOfPrimitiveOrWrapper(TypeKind.BOOLEAN, TypeKind.INT).optional() //
```
WonderCsabo: Do not we have more sophisticated validation here? For example can we have a second int parameter for `CompoundButton` first param?

AndroidAnnotations/androidannotations-core/androidannotations/src/main/java/org/androidannotations/internal/core/handler/CheckedChangeHandler.java
```diff
@@ -63,16 +66,26 @@ protected void makeCall(JBlock listenerMethodBody, JInvocation call, TypeMirror
...
-			}
+				viewElement = CanonicalNameConstants.COMPOUND_BUTTON;
+        JVar btnParam = listenerMethod.param(getClasses().COMPOUND_BUTTON, "buttonView");
+        call.arg(castArgumentIfNecessary(holder, CanonicalNameConstants.COMPOUND_BUTTON, btnParam, parameter));
+        if (listenerMethod.hasSignature( new AbstractJType[] { getClasses().COMPOUND_BUTTON, getCodeModel().BOOLEAN} )) {
+          JVar isCheckedParam = listenerMethod.param(getCodeModel().BOOLEAN, "isChecked");
+          call.arg( isCheckedParam );
+          break;
+        }
+			} else if (isTypeOrSubclass(CanonicalNameConstants.RADIO_GROUP, parameter)) {
```
WonderCsabo: It would be nice to extract the different generations to methods.

AndroidAnnotations/androidannotations-core/androidannotations/src/main/java/org/androidannotations/internal/core/handler/CheckedChangeHandler.java
dodgex: For cases like this it would be awesome if our `ValidatorParameterHelper` would support some kind of `or`.

e.g.
```
validatorHelper.param
	.anyOrder()
	.extendsType(CanonicalNameConstants.COMPOUND_BUTTON).optional()
	.primitiveOrWrapper(TypeKind.BOOLEAN).optional()

	or()

	.anyOrder()
	.extendsType(CanonicalNameConstants.RADIO_GROUP).optional()
	.primitiveOrWrapper(TypeKind.INT).optional()

	.validate(executableElement, validation);
```

dodgex:
I tested a bit with this PR but it looks like there is no way to pick the correct callback for target methods with no arguments - what is currently supported with `@CheckedChange`. I'd say that we should create a new annotation for handling `RadioGroup` checked change. 

Also a good validation is hard to achiev if we want to support both `CompoundButton` and `RadioGroup` with one annotation.

WonderCsabo:
Considering these things, I agree, we should go with the new annotation.
Maybe we can create an abstract handler to save some code duplication.

dodgex:
Thank you @gauravsak for the attempt but after working on this PR we decided that it is better to create a new annotation instead of extending this one. I'll close this PR.

wmailn:
Are there any new annotation now?

Issue #1929, opened 2017-01-04
False negative warning for valid meta-data


WonderCsabo:
AA cannot parse a valid meta-data element from the manifest, like this:
```xml
<meta-data
                android:name="com.commonsware.cwac.provider.STREAM_PROVIDER_PATHS"
                android:resource="@xml/file_paths"/>
```

It emits the following warning: 
```
warning: A malformed <meta-data> has been found in the manifest with name com.commonsware.cwac.provider.STREAM_PROVIDER_PATHS
```
This is a valid meta-data, so we should be able to parse it.

dodgex:
also: the warning is ... bad :D a malformed WHAT has been found? :D

WonderCsabo:
It is okay, we print the name of the meta-data element (which is this weird thing in this case).

dodgex:
yeah but still i think the warning should contain the element type in some way.

WonderCsabo:
Oh, the message is correct, just Github removed `<meta-data>`. :)

This is the full message
```
warning: A malformed <meta-data> has been found in the manifest with name com.commonsware.cwac.provider.STREAM_PROVIDER_PATHS
```

dodgex:
Oh. Okay. :)

smaugho:
The problem is that I assumed that always it should have a "value" field, and this is not true. This is the code:

```java
private Map<String, String> extractMetaDataQualifiedNames(NodeList metaDataNodes) {
		Map<String, String> metaDataQualifiedNames = new HashMap<String, String>();
		
		for (int i = 0; i < metaDataNodes.getLength(); i++) {
			Node node = metaDataNodes.item(i);
			Node nameAttribute = node.getAttributes().getNamedItem("android:name");
			Node valueAttribute = node.getAttributes().getNamedItem("android:value");

			if (nameAttribute == null || valueAttribute == null) {
				if (nameAttribute != null) {
					LOGGER.warn("A malformed <meta-data> has been found in the manifest with name {}", nameAttribute.getNodeValue());
				} else {
					LOGGER.warn("A malformed <meta-data> has been found in the manifest");
				}
			} else {
				metaDataQualifiedNames.put(nameAttribute.getNodeValue(), valueAttribute.getNodeValue());
			}
		}
		
		return metaDataQualifiedNames;
	}
```

I will try to make it generic.. Now following the correct syntax for the meta-data tag:

```xml

    <meta-data android:name="string"
               android:resource="resource specification"
               android:value="string" />

[...]

PR #1956, opened 2017-02-19
Fix AndroidManifest meta-data parsing

simopete:
Implements #1929


AndroidAnnotations/androidannotations-core/androidannotations/src/main/java/org/androidannotations/helper/AndroidManifest.java
```diff
@@ -106,4 +106,33 @@ public int getTargetSdkVersion() {
 		return targetSdkVersion;
 	}
 
+	public static final class MetaDataInfo {
+
+		private String name;
```
WonderCsabo: Can you make these `final`?

dodgex:
Thank you!

PR #1962, opened 2017-03-05
Inflate menu before calling method injection

dodgex:
Fix issue reported in https://github.com/androidannotations/androidannotations/issues/1961

### DIFF
--- a/AndroidAnnotations/androidannotations-core/androidannotations-test/src/test/java/org/androidannotations/test/menu/InjectMenuActivityTest.java
+++ b/AndroidAnnotations/androidannotations-core/androidannotations-test/src/test/java/org/androidannotations/test/menu/InjectMenuActivityTest.java
@@ -16,11 +16,16 @@
 package org.androidannotations.test.menu;
 
 import static org.fest.assertions.api.Assertions.assertThat;
+import static org.mockito.ArgumentMatchers.any;
+import static org.mockito.ArgumentMatchers.anyInt;
+import static org.mockito.Mockito.doAnswer;
 import static org.mockito.Mockito.mock;
 
 import org.junit.Before;
 import org.junit.Test;
 import org.junit.runner.RunWith;
+import org.mockito.invocation.InvocationOnMock;
+import org.mockito.stubbing.Answer;
 import org.robolectric.Robolectric;
 import org.robolectric.RobolectricTestRunner;
 
@@ -34,6 +39,7 @@ public class InjectMenuActivityTest {
 	@Before
 	public void setUp() {
 		injectMenuActivity = Robolectric.buildActivity(InjectMenuActivity_.class).create().get();
+		injectMenuActivity.mockMenuInflater = createMenuInflater();
 	}
 
 	@Test
@@ -49,6 +55,14 @@ public class InjectMenuActivityTest {
 	}
 
 	@Test
+	public void methodInjectionComesAfterInflation() {
+		Menu menu = mock(Menu.class);
+		assertThat(injectMenuActivity.menuIsInflated).isFalse();
+		injectMenuActivity.onCreateOptionsMenu(menu);
+		assertThat(injectMenuActivity.menuIsInflated).isTrue();
+	}
+
+	@Test
 	public void methodInjectedMenu() {
 		Menu menu = mock(Menu.class);
 		injectMenuActivity.onCreateOptionsMenu(menu);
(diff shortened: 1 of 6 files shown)


dodgex:
I just realized that it should be enough to move `InjectMenuHandler` below `OptionsMenuHandler` or `OptionsMenuItemHandler` (see [CorePlugin.java#L177](https://github.com/androidannotations/androidannotations/blob/develop/AndroidAnnotations/androidannotations-core/androidannotations/src/main/java/org/androidannotations/internal/core/CorePlugin.java#L177))

@WonderCsabo what do you think should we prefer?

WonderCsabo:
@dodgex i think this fix is better, because this way we do not rely on the order of handlers.

dodgex:
Not sure how to test this as it seems that robolectrict does [not support menu inflation](https://github.com/androidannotations/androidannotations/blob/b1e1903307d143b22f985d61c8b8e552d9c1c291/AndroidAnnotations/androidannotations-core/androidannotations-test/src/test/java/org/androidannotations/test/menu/OptionsMenuActivityTest.java#L98)?

WonderCsabo:
@dodgex maybe you could create a *spy* of the `Activity`?
```java
TestActivity activity = Robolectric.buildActivity ...
TestActivity activitySpy = Mockito.spy(activity);

Mockito.when(activity.getMenuInflater()).thenReturn(...); // create a mock menu inflater, and stub its inflate() method which alters the `Menu` object in some way you can assert on it 

activity.onCreateOptionsMenu(menu);
assertThat(menu).is...
```

dodgex:
added a test similar to what you suggested.

dodgex:
Yay.... to get the spy working i needed the generated class to be non-final... 

> AbstractActivityTest.activityShouldBeFinal:33 expected:<[tru]e> but was:<[fals]e>

but there is a test that checks if a generated class is final...

WonderCsabo:
Then we can just override getMenuInflater in the test activity which
returns the mock, and check if it was called correctly with some flag.

dodgex:
Build should go green now.

AndroidAnnotations/androidannotations-core/androidannotations-test/src/test/java/org/androidannotations/test/menu/InjectMenuActivityTest.java
```diff
@@ -51,14 +58,34 @@ public void testMenuInjectedFromOnCreateOptionsMenu() {
...
+		assertThat(injectMenuActivity.menuIsInflated).isTrue();
+
 		assertThat(injectMenuActivity.multiInjectedMenu).isSameAs(menu);
 	}
+
+	private InjectMenuActivity.MockMenuInflater createMenuInflater() {
+		final InjectMenuActivity.MockMenuInflater menuInflater = Mockito.mock(InjectMenuActivity.MockMenuInflater.class);
+		doAnswer(new Answer<Void>() {
+			public Void answer(InvocationOnMock invocation) {
+				menuInflater.menuInflated = true;
```
WonderCsabo: Why is this needed? You can just you Mockito verification, since you already using mocking.
```java
Mockito.verify(menuInflater).inflate(anyInt(), any(Menu.class));
```

This way you can get rid of the inner class and the boolean field as well.

Another thing: most of the times, we static import methods like `mock` and `verify`.

dodgex: We then would need to call `verify()` in the InjectMenuActivity. after the call to `injectMenuActivity.onCreateOptionsMenu(menu);` the inflater is awlays called. the interesting part is that it has to be called BEFORE a method with `@InjectMenu` is called.

WonderCsabo: Yes, you are right.

dodgex:
I reverted the changes to the existing tests and added a new test case for this

AndroidAnnotations/androidannotations-core/androidannotations-test/src/test/java/org/androidannotations/test/menu/InjectMenuActivityTest.java
dodgex: So do you have another Idea to change this or should we keep it?

WonderCsabo: Let's keep this one.

AndroidAnnotations/androidannotations-core/androidannotations-test/src/test/java/org/androidannotations/test/menu/InjectMenuActivityTest.java
```diff
@@ -49,6 +55,17 @@ public void testMenuInjectedFromOnCreateOptionsMenu() {
...
 
 	@Test
+	public void methodInjectionComesAfterInflation() {
+		Menu menu = mock(Menu.class);
+
+		assertThat(injectMenuActivity.menuIsInflated).isFalse();
+		injectMenuActivity.onCreateOptionsMenu(menu);
+		assertThat(injectMenuActivity.menuIsInflated).isTrue();
+
+		assertThat(injectMenuActivity.methodInjectedMenu).isSameAs(menu);
```
WonderCsabo: I think this `assert` is not needed, it is already checked in other tests.

dodgex: removed

Issue #1966, opened 2017-03-07
Method based @ViewsById isEmpty()


miszmaniac:
I've tried to used method based @ViewsById method, but found out that method parameter is empty List.
This happens beacuse of wrong order of calls in generated code.
I've also noticed that test of this part of code is incorrect. 
![zrzut ekranu 2017-03-07 11 32 31](https://cloud.githubusercontent.com/assets/1630300/23652595/d2b45d60-0329-11e7-8bcc-a29f5f7aea47.png)

**AndroidAnnotations version**:
4.2.0
**Android compile SDK version**:
25
**Annotated code**:
```java
 @ViewsById({R.id.divider_1, R.id.divider_2, R.id.divider_3, R.id.short_cut_container})
    void init(List<View> views) {
       ...
    }
```

**Expected generated code**:
```java
@Override
    public void onViewChanged(HasViews hasViews) {
        ArrayList<View> list_init_views = new ArrayList<View>();
        View view_divider_1 = hasViews.findViewById(R.id.divider_1);
        View view_divider_2 = hasViews.findViewById(R.id.divider_2);
        View view_divider_3 = hasViews.findViewById(R.id.divider_3);
        View view_short_cut_container = hasViews.findViewById(R.id.short_cut_container);

        if (view_divider_1 != null) {
            list_init_views.add(view_divider_1);
        }
        if (view_divider_2 != null) {
            list_init_views.add(view_divider_2);
        }
        if (view_divider_3 != null) {
            list_init_views.add(view_divider_3);
        }
        if (view_short_cut_container!= null) {
            list_init_views.add(view_short_cut_container);
        }
    {
            List<View> views = null;
            views = list_init_views;
            init(views);
        }
    }
```

**Actual generated code**:
```java
@Override
    public void onViewChanged(HasViews hasViews) {
        ArrayList<View> list_init_views = new ArrayList<View>();
        View view_divider_1 = hasViews.findViewById(R.id.divider_1);
        View view_divider_2 = hasViews.findViewById(R.id.divider_2);
        View view_divider_3 = hasViews.findViewById(R.id.divider_3);
        View view_short_cut_container = hasViews.findViewById(R.id.short_cut_container);

        {
            List<View> views = null;
            views = list_init_views;
            init(views);
        }
        if (view_divider_1 != null) {
            list_init_views.add(view_divider_1);
        }
        if (view_divider_2 != null) {
            list_init_views.add(view_divider_2);
        }
[...]

PR #1967, opened 2017-03-07
Use correct block for method injection of ViewById and ViewsById

dodgex:
fix https://github.com/androidannotations/androidannotations/issues/1966


WonderCsabo:
@dodgex i do not understand this. The current code is using the injection block, and these are injection annotations.

dodgex:
There is an issue with the order of injections. If you use method injection for a list of views there is a chance that you get an empty or only partially filled list in the injecting method. With this change the method injection is ensured to be called after all views have been resolved. you can see an example of generated code in the issue #1966

WonderCsabo:
@dodgex but with your change, the view injections end up in the same block as `@AfterViews`. And in that case, in case of wrong handler order, `@AfterViews` generated code could be added before the view injections... We created the blocks to add code in correct order, and this change would revert that, making the blocks redundant.

dodgex:
Oh damn... yeah. I forgot about `@AfterViews`. I'll see if i can find a better solution. :)

dodgex:
PR Updated. I added a new block for viewholder lookups between "before injection" and "injection" blocks.

PR #1968, opened 2017-03-07
Move super call onRestoreInstanceState below state restoration

dodgex:
see https://github.com/androidannotations/androidannotations/issues/1959


dodgex:
@WonderCsabo added test & rebased

PR #1971, opened 2017-03-11
Update versions in example projects

dodgex:
Updates versions as mentioned in https://github.com/androidannotations/androidannotations/issues/1932
and updates APT configuration for latest android plugin version. related to https://github.com/androidannotations/androidannotations/issues/1970


dodgex:
@WonderCsabo the new gradle plugin seems to require java 8 to build. is there a way to let travis only run certain steps on jdk7 and the gradle builds only on jdk8?

WonderCsabo:
@dodgex we can drop the JDK 7 build, JDK 7 already reached EOL as far as i know.

dodgex:
According to [this](https://java.com/en/download/faq/java_7.xml) Java 7 has reached his EOL in April 2015. I'll remove java7 from travis build then

WonderCsabo:
Thanks, great work!

Issue #1976, opened 2017-03-14
Generate classes even if there was a validation error


smaugho:
Right now AA skips the generation of all the Enhanced Classes  when there was a validation error in any place in the project. This causes a huge list of errors to be displayed and it is almost impossible to find the error that is causing the problem

 A good option, would be to generate all the classes, except the one causing the error... This would be the simplest solution. 

Even though, in most of the scenarios would be useful even generate the class with the validation error as described [here](https://github.com/androidannotations/androidannotations/issues/1868#issuecomment-269757610), but this would require probably a revision of every annotation handler in the framework.

dodgex:
I'm currently investigating this. At the moment it looks to be a fairly easy task. I already got a build where it only produces the validation errors but generates classes (at least if it is not an error on a class level annotation). But I want to do some more tests. :)

dodgex:
@smaugho could you checkout and build a local snapshot of https://github.com/androidannotations/androidannotations/pull/1977 and see how it behaves for your project?

smaugho:
Yes I can, I'll be on this now...

smaugho:
Well, I'm making some tests with it... but right now it is failing

![im](https://cloud.githubusercontent.com/assets/6622086/23921942/c5371952-0900-11e7-8150-a780c3b34846.png)

The error in the LogFile:

The error: ERROR:74 - Resource name not found in R.id: modalEditExpenseTEST

Please note that I don't have any only-AA here, I use DecleX, but this is similar to having a plugin applied no difference in behavior should be noted, all the features from the framework should keep the same (but better that you are ware for any comparison ;) )... I'm investigating here what happens after there's one invalidation, and if it gets to the generate method... I will let you know the results. 

This project I'm testing is not that complex, I have some complex one where I can test after, but well, it should pass this one..

smaugho:
Ok, in general, it generated the classes, but for some reason they were not detected by "Java" I had this problem before, just trying to remember what was it :).. I will try with something simpler. 


```log
21:50:12.495 [Daemon worker Thread 3] INFO :340 - Number of files generated by AndroidAnnotations: 27
21:50:12.497 [Daemon worker Thread 3] DEBUG:57 - Generating class: com.dspot.declex.example.expenses.auth.Auth_
21:50:12.584 [Daemon worker Thread 3] DEBUG:57 - Generating class: com.dspot.declex.example.expenses.fragment.ExpenseDetailsFragmentActionHolder
[...]

PR #1978, opened 2017-03-14
Generate classes even after validation errors

smaugho:
This PR includes changes applies by the PR [1977](https://github.com/androidannotations/androidannotations/pull/1977), plus some improvements. 

see #1976


smaugho:
Hi, it seems the reason for the build fail it is that there's some Tests that requires that the process failed. Right now as explained in #1978 the process is failing in the same round... it seems this is not being detected... 

I've run several tests in projects and it seams to be working fine.

Could you anyway please check why they are failing exactly?.. maybe it is something else. 

Thanks

dodgex:
Thank you for testing and enhancing my PR. I'll try to test this on my project and check whats wrong with the tests as soon as possible. But currently I have a lot of projects for my job and in my free time. I hope I can get into it this weekend! :)

dodgex:
I just closed my version of this PR as we should not have two PR for the same issue. Please ensure that you have checked the `Allow edits from maintainers.` checkbox on the bottom of the right menu in this PR. so we can directly contribute to your PR to fix it.

smaugho:
Yes, the option is checked.. Didn't know it existed, sorry maybe I was able to work in your PR.

dodgex:
I think this option is fairly new. but you would not have been able to work in my PR as it requires write access to the androidannoations repo to get permission to write on PR branches when this option is checked.

dodgex:
It looks like the problems with the test are related to the `Messager` used in our `MessagerAppender`.

When its `printMessage` is called during processing it internally has a ` JavaFileObject` as source that is used to set the required source field in the `Diagnostic` that is checked in `ProcessorTestHelper` to check if the error is present in the correct source file.

I'll see if i can find a way to keep this source state for the deffered adding of the erros that you implemented. But I'm afraid that this might limit our compiler support as the implementation of the `Messager` interface is part of the compiler used (in my test case it is `com.sun.tools.javac.processing.JavacMessager` of the `javac` compiler).

smaugho:
Ok, I guess sacrificing  the tests is not an option neither... I simply didn't find a way to keep the compiler working at least that the report was done in that last round... not sure if making it in the prev-last round would make difference, but that would be hard to detect from within the framework. 

I'm not sure of this, but isn't there a way to send information to the test from within the annotation processor?... of course that would require to change those tests...

dodgex:
The problem is that we not only break the tests with this behaviour but also e.g. eclipse would no longer be able to show the error on the annotation as it currently is as it is missing the information about the source file that has the error.

smaugho:
Making it like this for Android Studio only would work? The real problem that we are solving with this PR is exactly there.

smaugho:
Well, I guess if it is applied to Android Studio, it would be to IntellyJ in general.. not sure how to detect that exactly, but well, between eclipse and IntellyJ the biggest difference is the folder structure..

smaugho:
Mmm.. well, in that moment, this fixed is not needed anymore... right now IMHO, this is something serious, to be no able to detect the bugs, or have to make that process of watching the console, or going to the log file (I have it opened all the time in NotePad++ for instance) it is really annoying for programming... I really wouldn't like to omit this fix.. of course, maybe there's some other way to make it work..

smaugho:
Maybe not completely good idea, and it sounds more like a "patch" to this issue... but well, it can be created a Flag to pass as parameter to the framework which activates this behavior..

dodgex:
It is even an issue in CLI builds as the javac does not know the files where the error is added.

See this screenshot as an example. the blue line should be part of the `MainActivity.java` block. building on CLI looks the same.
image

dodgex:
> Mmm.. well, in that moment, this fixed is not needed anymore... 

well it would still be required as on bigger projects not generating source might again turn the whole project red and makes the source of the issue hard to find.

smaugho:
Yes that's right.. but this was something that I didn't understand, but now that you mention it, maybe there's a simple solution to it.. in the code, what was done, it is that the error is reported, by a mechanism in which I don't see the Element.. 

This is the original code (which I didn't modify):

```java
for (Element annotatedElement : annotatedElements) {
	ElementValidation elementValidation = annotationHandler.validate(annotatedElement);

	AnnotationMirror annotationMirror = elementValidation.getAnnotationMirror();
	for (ElementValidation.Error error : elementValidation.getErrors()) {
		LOGGER.error(error.getMessage(), error.getElement(), annotationMirror);
	}

	for (String warning : elementValidation.getWarnings()) {
		LOGGER.warn(warning, elementValidation.getElement(), elementValidation.getAnnotationMirror());
	}

	if (elementValidation.isValid()) {
		validatedAnnotatedElements.add(annotatedElement);
	} else {
		LOGGER.warn("Element {} invalidated by {}", annotatedElement, annotationName);
	}
}
```

Maybe it would be enough if we instead of reporting all the errors separated and the warning separated, we could create a single String and reported in the last step with the major Level (so if there's Info and Warns, a Warn would be reported, if there's Error, then this is what would be reported).

I will try this I will let you know what happens then.

smaugho:
Mmm.. no, I see they are almost the same... the main difference is that the last one uses de annotated alement and a Warn, which is reported correctly...

smaugho:
Ok, I got it completely now.. there's no easy work around.. since the JavaFileObject exists only in that round, probably you cannot reference it in a different round... 

But in general JavaFileObjects can be created and passed to other methods.. not sure if that could be a possible solution, I would have to see more of your research about it... 

Well, a different approach it is to report all as a Warning and fail in the last round (this will solve what you point of the place the Error is being generate).. 

But of course, it still doesn't solve what the Tests do....

smaugho:
Please note that I simply passed a reference to the Element and Annotation Mirror obtained in the first round, to the last one.... I'm just thinking if trying to create an an element in the last round pointing to the same place would make any difference...

dodgex:
I do not think that creating a Element or something is a solution as we are in java internal api where the relevant stuff happens... :(

dodgex:
> I would have to see more of your research about it...

well there is not much to know. I found that the Messager is [`com.sun.tools.javac.processing.JavacMessager`](http://grepcode.com/file/repository.grepcode.com/java/root/jdk/openjdk/8u40-b25/com/sun/tools/javac/processing/JavacMessager.java?av=f) and it sets the related source in lines 98-105. and this source is required by ides and the test to know the file related to the warning/error.

dodgex:
unfortunately i haven't found a solution for this problem yet. :(

smaugho:
I got here some good result with something, I will let you know if I can get more improvements ;)

screenshot_1

dodgex:
What have you done to reach this? maybe I'll can try too

smaugho:
I'm still testing this... what I did was to get the elements from the TypeElement:

I use getElement to send the message

```java
private class Message {
		Kind kind;
		String message;
		AnnotationMirror annotationMirror;
		
		List <String> elements = new LinkedList<>();

		Message(Kind kind, String message, Element element,
				AnnotationMirror annotationMirror) {
			super();
			this.kind = kind;
			this.message = message;
			this.annotationMirror = annotationMirror;
			
			Element enclosingElement = element;
			do { 
				elements.add(0, enclosingElement.toString());
				enclosingElement = enclosingElement.getEnclosingElement();
			} while (!enclosingElement.getKind().equals(ElementKind.PACKAGE));
			
			System.out.println("ELEMENTS: " + elements);
		}

		public Element getElement() {
			Element element = processingEnv.getElementUtils().getTypeElement(elements.get(0));
			elements.remove(0);
			
			while (elements.size() > 0) {
				for (Element elem : element.getEnclosedElements()) {
					if (elem.toString().equals(elements.get(0))) {
						System.out.println("FOUND: " + elem);
						elements.remove(0);
						element = elem; 
						break;
					}
				}
			}
			
			return element;
		}

	}
```

smaugho:
The test is getting stuck.. I'm checking this:

ss 2017-03-16 at 09 21 19

To be honest I'm not sure why is needed the AnnotationMirror... well, I will try to generated also from the element

dodgex:
Not sure what your problem is there?

smaugho:
I'm not sure neither.. it simply gets stuck there... I will try to print more information in console (not sure if you can debug there directly the annotation processor, I can't, read some articles about setting this in IntelliJ, but I'm using Eclipse to program in the framework, hehehe, have to migrate).

WonderCsabo:
Guys, for faster messaging I suggest Gitter. ☺️

On Mar 16, 2017 21:27, "Kay-Uwe Janssen" <notifications@github.com> wrote:

Not sure what your problem is there?

—
You are receiving this because you are subscribed to this thread.
Reply to this email directly, view it on GitHub
<https://github.com/androidannotations/androidannotations/pull/1978#issuecomment-287181103>,
or mute the thread

dodgex:
good idea, lets continue here: https://gitter.im/androidannotations/PR1978

dodgex:
Well with latest changes i just pushed the build of this PR should go green. :)

smaugho:
Hi @dodgex I got kind of an infinite loop with this method... 

```log
weekExpenses(WeekExpense_) in [com.dspot.declex.example.expenses.fragment.StatisticsFragment, weekExpenses(WeekExpense_)]
   Check StatisticsFragment()
   Check weekExpenses
   Check weekExpenses(com.dspot.declex.example.expenses.model.WeekExpense_)
   Check onBackPressedEvent()
```

Basically it happens because when the error occurred in the validation, WeekExpense_ is not defined. At the end of the process the class WeekExpense_ is generated and the method toString is different than the first one...  the infinite loops occurs here:

```java
while (localElements.size() > 0) {
		System.out.println(localElements.get(0) + " in " + elements);
		Thread.sleep(1);
		if (element instanceof ExecutableElement) {
			ExecutableElement method = (ExecutableElement) element;
			for (VariableElement param : method.getParameters()) {
				if (param.toString().equals(localElements.get(0))) {
					localElements.remove(0);
					element = param;
					break;
				}
			}
		} else {
			for (Element elem : element.getEnclosedElements()) {
				System.out.println("   Check " + elem);
				if (elem.toString().equals(localElements.get(0))) {
					localElements.remove(0);
					element = elem;
					break;
				}
			}
		}
	}
```

In MessagerAppender.. 

I reproduced got this in DecleX, where it injects parameters (this is not present in AndroidAnnotation).. I'm thinking in some situation in which this could happen in AA...

smaugho:
Well the idea is, to declare an EBean, then let's say in an activity use something like this:

```java
@AfterViews
void init(MyBean_ bean) {

}
```

Here @AfterViews should fail because it doesn't accept a parameter... the String for that element would be:

`init(MyBean_)`

at the end of the process (last round) when it tries to get that element again, since in that moment the class MyBean_ was successfully generated, then it should get:

`init(my.package.MyBean_)`

And the infinite loop situation would be reproduced.

smaugho:
This could be patched, detecting the infinite loop and then in the methods removing the "package" part of the parameters and check in that way... but I'm not sure if this would be good enough...

smaugho:
I'm not sure if updating the PR is good idea after you made the other modifications.. so I let here a fix to this problem:

```java
private Element getElement() {
	if (elements.isEmpty()) {
		return null;
	}

	boolean ignorePackage = false;
	
	List<String> localElements = new LinkedList<>(elements);
	Element element = processingEnv.getElementUtils().getTypeElement(localElements.remove(0));
	while (localElements.size() > 0) {
		int prevSize = localElements.size();
		
		if (element instanceof ExecutableElement) {
			ExecutableElement method = (ExecutableElement) element;
			for (VariableElement param : method.getParameters()) {						
				if (param.toString().equals(localElements.get(0))) {
					localElements.remove(0);
					element = param;
					break;
				}
			}
		} else {
			for (Element elem : element.getEnclosedElements()) {
				
				String elemStringValue = elem.toString();
				
				if (ignorePackage) {
					elemStringValue = elemStringValue.replaceAll("([a-zA-Z_$][a-zA-Z_$0-9]\\.)+", "");
				}
				
				if (elemStringValue.equals(localElements.get(0))) {
					localElements.remove(0);
					element = elem;
					break;
				}
			}
		}
		
		if (prevSize == localElements.size()) {
			ignorePackage = true;
		}
	}

	return element;
}
```

dodgex:
@smaugho could you provide a small sample class that triggers the issue you mentioned? Thanks!

smaugho:
Hi.. now I'm trying to build this PR, but I get Failures all the time, in the tests, in 

obraz

Well basically the full idea that I proposed above was:

Bean:

```java

@EBean
public class MyBean {
    
}

```
Activity:

```java
@EActivity(R.layout.main_activity)
public class MainActivity extends Activity {

    @AfterViews
    void init(MyBean_ myBean) {  //Here you should get an infinite cycle, the compiler simply continues forever
    }

}
```

dodgex:
@smaugho I can't reproduce this issue with your sample.for simplicity i added the parameter to the `AfterViewsActivity` from androidannotations-test.

```
package org.androidannotations.test.afterviews;

import org.androidannotations.annotations.AfterViews;
import org.androidannotations.annotations.Bean;
import org.androidannotations.annotations.EActivity;
import org.androidannotations.test.R;

import android.app.Activity;

@EActivity(R.layout.main)
public class AfterViewsActivity extends Activity {
	
	@Bean
	AfterViewBean afterViewBean;
	
	public boolean afterViewBeanCalledBefore = false;

	@AfterViews
	void afterViews(AfterViewBean_ y) {
		afterViewBeanCalledBefore = afterViewBean.afterViewsCalled;
	}
}
```
and as result I get
> AfterViewsActivity.java:[33,9] org.androidannotations.annotations.AfterViews cannot have any parameters

can you provide a small project that can reproduce this infinite loop issue? If necessary a project using declex.

dodgex:
when i remove the `@Bean AfterViewBean afterViewBean;` get the following error in intelliJ and the same as above in CLI build using maven.
```
Error:(34, 25) java: cannot find symbol
  symbol:   class AfterViewBean_
  location: class org.androidannotations.test.afterviews.AfterViewsActivity
```

smaugho:
Hi, I made local build normally (had to build without the tests, because it is really not passing the tests at least locally in this PR), and I got with that example the infinite loop:

I uploaded the Android Studio project to my GitLab repo: http://repo.ciberos.com/root/infinite_loop.git

AndroidStudio still trying to build, more than one minute:

obraz

Not sure if building with some other tool would me the things different.. I'm using Android Studio 2.3, and well, he Gradle Wrapper and Android Gradle Plugin are the one declared in the project files ;).

dodgex:
Thank you, I can reproduce the issue with this project.

dodgex:
I added your change with a tiny refactor. I added an extra check to avoid the infinite loop when the remove package round also did not found a result. In that case I return the current element, so we have the error on the nearest known element.

Could you test this please? :)

smaugho:
Yeap, I think is good idea, even if we think this should not happen to return the element if nothing is found :).. I will test with same project. The solution I did is applied to DecleX and I used it already in some big projects without any problem, I found this one, and with the work-around with the packages it was solved.

smaugho:
@dodgex it is working fine, I merged this code with DecleX "develop" branch and compiled 2 big projects, everything is working fine, errors are reported nicely.

dodgex:
@smaugho @WonderCsabo Had you guys any chance to do some more tests/builds with this changes? I unfortunately am currently not working on an Android App so i had no chance to test this in a real project.

If you guys found no additional issues I'd re-request the Review of @WonderCsabo ;)

smaugho:
@dodgex I've tested this enough integrated to Declex. I've applied it to 4 apps, 2 simpler and 2 bigger. I really suggest a last review of @WonderCsabo ;)

WonderCsabo:
@smaugho @dodgex i finally checked this PR. However, after rebasing this to develop, i got this test failure:
```
Failed tests: 
  KeyEventTest.compileTestWithTypeImplementsKeyEventCallback:34->ProcessorTestHelper.assertCompilationSuccessful:130 Expected no errors, found error: Could not find the AndroidManifest.xml file, using  generation folder [/var/folders/lz/67px51zs5xv5v18s52tcc71c0000gn/T/AA8335127508239890971/build/generated/source/apt/debug])
```

Maybe the errors are not cleared correctly?

smaugho:
@WonderCsabo, maybe conflicting with some of the many changes done in AA on last months?... I will try to run it locally as well. The error itself looks like not related to the issue that was solving this PR.

WonderCsabo:
Nothing was changed related to this. The error is definitely caused by this PR. :(

WonderCsabo:
Thanks for checking @smaugho .

smaugho:
NP, got some free time now :), I'm on it...

WonderCsabo:
Yeah, it should be definitely something like this. Before `KeyEventTest`, `AndroidManifestFinderTest` runs, which creates 5 errors. But when running `KeyEventTest`, those five errors are still in the appender. Please note these errors are special, because they are not belong to any element (element is `null`).

WonderCsabo:
@smaugho i think i found a fix. I removed the `static` keyword from the `MessagerAppender.errors` field. The same processor instance is used between the rounds (and `init` is only called before the first round), so we do not have to use `static` to retain data between them. I think you can even remove the `errors.clear()` expression.

smaugho:
Nice thanks!!.. I was having some problems to run the project here. Please if this is a complete solution and all the checks are passed please commit and merge!!...

smaugho:
mm... but "errors" is static in the "develop" branch as well.... :/

WonderCsabo:
That field was introduced by you, it is [not on develop](https://github.com/androidannotations/androidannotations/blob/develop/AndroidAnnotations/androidannotations-core/androidannotations/src/main/java/org/androidannotations/logger/appender/MessagerAppender.java#L32).

smaugho:
sorry!!!... yeap, that's probably I got an error.. have to clean my "develop"... it is ahead of androidannotations develop branch!!

WonderCsabo:
Unfortunately  this is not a perfect solution. :( 

Let's have the following scenario:

1. You have an `@EBean`, which is invalid, and will not be generated.
2. You have a totally valid `@EActivity`, where you inject that bean.

In this case, we generated the code for the activity, because there are no validation errors on it, and we will generate not compilable code, because we are referring to the not generated EBean class in it!

smaugho:
@WonderCsabo yes this is true, but anyway, the project cannot run anyway due to the invalid EBean.. and yes, this will generate 2 errores, in the EBean and the Activity will have a compilation error (because the Bean genreated class doesn't exist).. but this is nothing compared with hundred of errors when something fails right now in AA...

dodgex:
I have to agree with @smaugho

smaugho:
It is good idea, but I'm not sure to be honest if this will solve the initial problem better... I mean, if you make this validation this means that the activity will be not generated due to its dependency (a Bean) will not be generated. You could be using that Activity (or Fragment, or Service, or... ) from many other places, which will generate as well a chain of errors (this is what this PR was trying to reduce, the huge amount of errors generated while compiling)... 

I'm not saying it is incorrect, just that it would be nice to weight what is more important here... once again, the PR was to reduce the amount of errors and let the programmer to focus in a few of them...

WonderCsabo:
@smaugho the Activity itself will be generated. The code which injects the bean into the activity will not be generated in it.

smaugho:
In my particular case, I use this solution in a day by day basis, and it helps me to find easily where the error is... I can ignore easily any "MyBean_ not found" in an "Activity_" class, but it is very hard to find an error when the amount of these start growing..

WonderCsabo:
@smaugho again, if we introduce these checks, we do not leak error to the user, because the Activity will be generated. Only we do not generated the bean injection code inside. So there will be no `MyActivity_` not found error.

smaugho:
yeap yeap, got it :).. sounds nice, just didn't see the first message... so basically the @Bean will do nothing if the generating handler for the @EBean failed...

WonderCsabo:
Yep. I am wondering if there are other cases where on annotation is depending on the other?

smaugho:
From AA, I don't remember another... 

@WonderCsabo are you implementing this already? 

It would be good to  invalidate @Bean when the "linked" @EBean validation failed... but I don't find a way to know this at priori just in the validation phase...  at least that you pass the "validatedAnnotatedElements" as parameter to the validations to the Annotation Handler... 

Any better idea?

WonderCsabo:
I am experimenting with this, yeah.

smaugho:
Ok, I guess you have it cracked :)... probably enough getting the validated elements from the AndroidAnnotationEnvironment..

WonderCsabo:
@smaugho i think i managed to do it. Here is my [branch](https://github.com/WonderCsabo/androidannotations/tree/bean-dep-validation). Can you reset your branch to mine?

Also @dodgex @smaugho please review my changes.

dodgex:
@WonderCsabo you should have write access to  the branch of this PR.

dodgex:
@WonderCsabo Does your code work with incremental compilation?

WonderCsabo:
@dodgex thanks, i did not know that, it is cool!

WonderCsabo:
@dodgex yes it works, i actually checked it, that is why i introduced the `extractedModel` into the `AndroidAnnotationsEnvironment`. (If you compile using IntelliJ "make" and not maven, it will use incremental).

AndroidAnnotations/androidannotations-core/androidannotations/src/main/java/org/androidannotations/logger/appender/MessagerAppender.java
```diff
@@ -66,4 +85,110 @@ private Kind resolveKind(Level level) {
 		return Kind.OTHER;
 	}
 
+	private class Message {
```
WonderCsabo: This class could be `static`.

AndroidAnnotations/androidannotations-core/androidannotations/src/main/java/org/androidannotations/logger/appender/MessagerAppender.java
```diff
@@ -66,4 +85,110 @@ private Kind resolveKind(Level level) {
...
+			return elemStringValue.replaceAll("([a-zA-Z_$][a-zA-Z_$0-9]*\\.)+", "");
+		}
+
+		private ElementDetails getElementDetails() {
+			Element element = getElement();
+			return new ElementDetails(element, getAnnotationMirror(element));
+		}
+	}
+
+	private class ElementDetails {
```
WonderCsabo: This class could be `static`.

AndroidAnnotations/androidannotations-core/androidannotations/src/main/java/org/androidannotations/logger/appender/MessagerAppender.java
```diff
@@ -66,4 +85,110 @@ private Kind resolveKind(Level level) {
...
+
+			for (AnnotationMirror mirror : element.getAnnotationMirrors()) {
+				if (mirror.toString().equals(annotationMirrorString)) {
+					return mirror;
+				}
+			}
+			return null;
+		}
+
+		private Element getElement() {
```
WonderCsabo: This is a rather complicated method, should not we cache the result? Or this is called only once?

WonderCsabo: I think we should move this to an utility method, this logic is too much for a `getter`.

dodgex: As far as i can see it is only called once when the appender is closed in the last round.

dodgex: what do you mean by `move this to an utility method`? If it is to much logic for a "getter" i'd just rename the method to something like `findRelatedElement` or so. :D

AndroidAnnotations/androidannotations-core/androidannotations/src/main/java/org/androidannotations/logger/appender/MessagerAppender.java
dodgex: only if we pass the processingEnv as it is used in getElement()

WonderCsabo: Okay, leave it.

WonderCsabo:
Did we update all licences? :)

dodgex:
I think all licenses are up to date. but I can check that when updating the PR. how to go with the `getElement`?

dodgex:
You missed to update the licenses but i fixed that. Also I made the ElementDetails static.

AndroidAnnotations/androidannotations-core/androidannotations/src/main/java/org/androidannotations/logger/appender/MessagerAppender.java
WonderCsabo: Okay, we can leave this here.

WonderCsabo:
Finally merged. Let's hope if will not introduce problems. Thanks for your work on this, @smaugho !

smaugho:
You're are welcome! Would like to be able to contribute much more to AA... lacking of time :'( ...

PR #2008, opened 2017-05-26
Add support for new method signature of findViewById in Android O

dodgex:
This PR allows to build against Android-O and fixes https://github.com/androidannotations/androidannotations/issues/2001 by adding an updated version of the view changed notfication code.

### DIFF
--- a/AndroidAnnotations/androidannotations-core/androidannotations/src/main/java/org/androidannotations/holder/EComponentWithViewSupportHolder.java
+++ b/AndroidAnnotations/androidannotations-core/androidannotations/src/main/java/org/androidannotations/holder/EComponentWithViewSupportHolder.java
@@ -71,8 +71,12 @@ public abstract class EComponentWithViewSupportHolder extends EComponentHolder i
 
 	public EComponentWithViewSupportHolder(AndroidAnnotationsEnvironment environment, TypeElement annotatedElement) throws Exception {
 		super(environment, annotatedElement);
-		viewNotifierHelper = new ViewNotifierHelper(this);
+		viewNotifierHelper = new ViewNotifierHelper(this, environment);
 		keyEventCallbackMethodsDelegate = new KeyEventCallbackMethodsDelegate<>(this);
+	}
+
+	public IJExpression getFindViewByIdExpression(JVar idParam) {
+		return _null();
 	}
 
 	public JBlock getOnViewChangedBody() {
@@ -132,7 +136,7 @@ public abstract class EComponentWithViewSupportHolder extends EComponentHolder i
 	}
 
 	public JInvocation findViewById(JFieldRef idRef) {
-		JInvocation findViewById = invoke(getOnViewChangedHasViewsParam(), "findViewById");
+		JInvocation findViewById = invoke(getOnViewChangedHasViewsParam(), "internalFindViewById");
 		findViewById.arg(idRef);
 		return findViewById;
 	}
--- a/AndroidAnnotations/androidannotations-core/androidannotations/src/main/java/org/androidannotations/holder/EComponentWithViewSupportHolder.java
+++ b/AndroidAnnotations/androidannotations-core/androidannotations/src/main/java/org/androidannotations/holder/EComponentWithViewSupportHolder.java
@@ -161,8 +161,6 @@ public abstract class EComponentWithViewSupportHolder extends EComponentHolder i
 
 		if (viewClass == null) {
 			viewClass = getClasses().VIEW;
-		} else if (viewClass != getClasses().VIEW) {
-			findViewExpression = cast(viewClass, findViewExpression);
 		}
 
 		IJAssignmentTarget foundView = fieldRef;
--- a/AndroidAnnotations/androidannotations-core/androidannotations/src/main/java/org/androidannotations/internal/helper/ViewNotifierHelper.java
+++ b/AndroidAnnotations/androidannotations-core/androidannotations/src/main/java/org/androidannotations/internal/helper/ViewNotifierHelper.java
@@ -1,5 +1,6 @@
 /**
  * Copyright (C) 2010-2016 eBusiness Information, Excilys Group
+ * Copyright (C) 2016-2017 the AndroidAnnotations project
  *
  * Licensed under the Apache License, Version 2.0 (the "License"); you may not
  * use this file except in compliance with the License. You may obtain a copy of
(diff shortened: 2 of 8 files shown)


naixx:
So, what is the status of the PR?

dodgex:
@WonderCsabo was not yet able to review this PR. Yesterday he told me that he will try to get review this and the other PR asap.

AndroidAnnotations/androidannotations-core/androidannotations-api/src/main/java/org/androidannotations/api/view/OnViewChangedListener2.java
```diff
@@ -0,0 +1,20 @@
...
+ *
+ * Unless required by applicable law or agreed To in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+ * License for the specific language governing permissions and limitations under
+ * the License.
+ */
+package org.androidannotations.api.view;
+
+public interface OnViewChangedListener2 {
```
dodgex: How could HasViewsBase look like? It would need to have a `findViewById` with a signature that is compatible to the other two versions. Or do i miss something in your suggestion?

WonderCsabo: You are right, we call `findViewById()` on `HasViews`, for some reason i assumed we just call `this.findViewById()`.

Ah we really should have dropped `@EBean` view support and the whole `HasViews` ceremony. :(

dodgex: Unfortunately

WonderCsabo: Yeah.

WonderCsabo: @dodgex i think we can do another thing. @naixx suggested we could rename the `findViewById()` method, and we could do just that. That way we should not even introduce `HasViews2`. The only change needed is implementing the new method in `Activity` and `ViewGroup`.

dodgex: yeah. we can do that. should i use the new generic variant for the renamed function then?

dodgex: and do you guys have a name suggestion? (I'm pretty bad in naming stuff)

WonderCsabo: If you want to use the generic variant, then we must use type cast in the implementing if `compileSdkVersion` < O. But that would be elegant, we could drop all the individual type casts. So go for it if you think.

Naming is very hard. What about `findViewByIdentifier()` ?

dodgex: for now i used `internalFindViewById` after i asked. if thats ok I'll keep it?

WonderCsabo: I like it, keep it. 👍

naixx: That's most probably won't clash with any user defined `findVuewById`

dodgex:
I just updated the PR with a total rewrite of the fix. should be much better now. :)

dodgex:
I have not yet tested it in an Android O build, but for initial review of the changes it should be okay.

AndroidAnnotations/androidannotations-core/androidannotations/src/main/java/org/androidannotations/internal/helper/ViewNotifierHelper.java
```diff
@@ -49,11 +58,25 @@ public JVar replacePreviousNotifier(JBlock block) {
...
+		JCodeModel codeModel = environment.getCodeModel();
+
+		JDirectClass genericType = codeModel.directClass("T");
+		JMethod findViewById = holder.getGeneratedClass().method(PUBLIC, genericType, "internalFindViewById");
+		findViewById.generify("T", environment.getClasses().VIEW);
+		findViewById.annotate(Override.class);
+
+		JVar idParam = findViewById.param(codeModel.INT, "id");
+		IJExpression findViewByIdExpression = holder.getFindViewByIdExpression(idParam);
+		findViewById.body()._return(JExpr.cast(genericType, findViewByIdExpression));
```
WonderCsabo: Two things:

1. type cast is not needed on Android O .
2. You can remove the cast for other places, since the type will be inferred with this signare

AndroidAnnotations/androidannotations-core/androidannotations/src/main/java/org/androidannotations/holder/EComponentWithViewSupportHolder.java
```diff
@@ -71,10 +71,14 @@
 
 	public EComponentWithViewSupportHolder(AndroidAnnotationsEnvironment environment, TypeElement annotatedElement) throws Exception {
 		super(environment, annotatedElement);
-		viewNotifierHelper = new ViewNotifierHelper(this);
+		viewNotifierHelper = new ViewNotifierHelper(this, environment);
 		keyEventCallbackMethodsDelegate = new KeyEventCallbackMethodsDelegate<>(this);
 	}
 
+	public IJExpression getFindViewByIdExpression(JVar idParam) {
+		return _null();
```
WonderCsabo: Maybe `getFindViewByIdExpression` should be an `abstract` method?

dodgex: I was considering this. Then we'd have the `return _null()` in `EBeanHolder`. I decided to keep it this way, but with no solid reason. If you prefer to have this abstract I can change it.

AndroidAnnotations/androidannotations-core/androidannotations/src/main/java/org/androidannotations/internal/helper/ViewNotifierHelper.java
dodgex: 1. I wanted to not have to check the Android version (or the presence of the generic method) here. Can change that if you want. Not sure if casts have a perf impact?
2. Done.

dodgex: Additional Note: It looks like the AppCompatActivity does not yet have the generic `findViewById` so the cast would be needed on android O + appcompat activity.

i checked the code in [android-o-preview-2](https://android.googlesource.com/platform/frameworks/support/+/android-o-preview-2/v7/appcompat/src/android/support/v7/app/AppCompatActivity.java#189) and [master](https://android.googlesource.com/platform/frameworks/support/+/master/v7/appcompat/src/android/support/v7/app/AppCompatActivity.java#189) branches.

WonderCsabo: Well, it would be nice to gain something from this new signature, if we doing so much work because of it. 😆 Since you already have the check code, i guess you can add it now.

WonderCsabo: Oh, they are overriding `findViewById()`, great... Then leave the cast here for now.

dodgex: i just found that `com.android.support:appcompat-v7:26.0.0-beta2` has the new signature.

WonderCsabo: Cool. Than try to remove the cast with O.

dodgex: The cast is now gone for Android O. But it now requires to be on appcompat 26.0.0-beta2 (maybe beta1 works too; alpha1 does not).

WonderCsabo:
@dodgex it turns out the cast is needed for `Fragments` even on Android O. 😢 

```java
return ((contentView_ == null)?null:contentView_.findViewById(id));
```
Because using ternary operator and not calling the `findViewById()` directly, the compiler cannot deduce type `T`.

dodgex:
@WonderCsabo well while i think it would be nice to not have the cast in the generated code i think the code in AA required to get the cast when we it is needed and not have it when not needed is not worth the benefit. therefore I'd remove the check again and cast always... wdyt?

WonderCsabo:
Okay.

dodgex:
check is gone. we should be good now =)

WonderCsabo:
I ran the test suite with `compileSdkVersion = O`, and everything passed.

WonderCsabo:
Nice work, @dodgex !

cshalysongoncalves:
:+1:

hswope:
I just upgraded to 4.3.1 to fix issues compiling with SDK 26 and now I get a different error: 
error: cannot find symbol method internalFindViewById(int)

PR #2015, opened 2017-06-12
Master

alkaidCUB:
I loved android annotations I will be using it in all my projects in Eclipse IDE. I want to see how it works in Android Studio


dodgex:
Thank you for your feedback. You can check the gradle samples in android studio to see how it works. :)

Issue #1499, opened 2015-07-20
NPE in @EIntentService.onHandleIntent


simonz:
NullPointerException was often occurred in `@EIntentService.onHandleIntent`
This is the generated `@EIntentService` code,

```
    @Override
    public void onHandleIntent(Intent intent) {
        MyIntentService_.super.onHandleIntent(intent);
        String action = intent.getAction();
        if (ACTION_EXECUTE.equals(action)) {
            // ....
        }   
```

sometimes `intent` gets null, so `intent.getAction()` makes NPE.
And I could find the documentation from http://developer.android.com/reference/android/app/IntentService.html#onStartCommand(android.content.Intent,%20int,%20int)

> The Intent supplied to startService(Intent), as given. This may be null if the service is being restarted after its process has gone away, and it had previously returned anything except START_STICKY_COMPATIBILITY.

I think it is safe to check whether `intent` is null.

WonderCsabo:
But does it make sense to start a sticky intent service? Because if I
understand correctly, it cannot actually handle intents after restart
because of the Intent being null.

simonz:
Yes, you're right. I was confused about the flags of `Service`.
I'm using IntentService as NOT_STICKY mode, but I reported NPE from Crashlytics.
the intent cannot be null, according to the document. Anyway I will reopen this issue after I get more information.
Thank you

chrulri:
I don't have more information, except that - according to my Crashlytics monitoring - this is happening from time to time in my app too. I'm using version 4.0.0 of AndroidAnnotations and only start the one IntentService using the builder methods provided by AndroidAnnotations.
Any ideas why the Intent could be occasionally null?

ronpeer7:
Same issue for me.
Any Solutions?

WonderCsabo:
@dodgex according to the JavaDoc, the `intent` parameter can be indeed `null`. So we have to add an additional check here.

dodgex:
PR is up. :)

WonderCsabo:
Finally fixed. Thanks @dodgex.

PR #2016, opened 2017-06-14
Ignore null-Intents in EIntentService classes

dodgex:
This PR avoids potential NullPointerExceptions in IntentServices as `onHandleIntent` might be called with a null intent.

see #1499


AndroidAnnotations/androidannotations-core/androidannotations/src/main/java/org/androidannotations/holder/EIntentServiceHolder.java
```diff
@@ -15,6 +15,7 @@
  */
```
WonderCsabo: Update the licence please.

dodgex: done

Issue #1627, opened 2015-11-16
Android DataBinding support


egslava:
Please, add/document support of Android DataBinding.

There're only two changes:
1. It needs to use `DataBindingUtil.setContentView` instead of `Activity.setContentView`.
2. We need access to result of `setContentView`.

WonderCsabo:
Unfortunately data binding has problems when annotation processors are used (see [last comments](https://bitbucket.org/hvisser/android-apt/issues/38/android-apt-breaks-brand-new-data-binding)). That is why i think it is not worth adding this feature to AA.

Moreover, this fix is not trivial from AA side. We would have to add a new parameter to `@EActivity` and `@EFragment` to instruct AA to use data binding. Also passing the result binding object to so a method for example annotated with `@AfterViews` or such. These changes only remove one line from the user's code... Also data binding is still kind of experimental.

However, please not new annotations now can be easily added by creating a simple AA plugin.

guitcastro:
@WonderCsabo If the problem related with the annotation process is fixed, you will consider add support for data binding, if it comes to be stable ?

I would love to use AA and data binding on my projects

WonderCsabo:
Well, it is not trivial to add it, but of course, it would be nice.

qazt482:
Any update on this? I would really like to use databinding.

guitcastro:
You can use Jack and Jill new annotation processor. It is compatible with Android DataBinding.
I've success used AA with Jack and Jill following this [example](https://github.com/nickwph/annotation-processor-with-java8-jack-and-jill-android/blob/master/app/build.gradle#L56-L73).

SimenB:
Jack doesn't work with data binding. 

https://code.google.com/p/android/issues/detail?id=210615

guitcastro:
Oh, my bad. @SimenB is right.

rkmax:
some update, now that the annotation processor is included on android gradle

brainscript:
**+1** I'd like to see this feature to.

gauravsak:
@WonderCsabo Please change the status of this issue according to the needed efforts. Thanks!

WonderCsabo:
I only experimented in with Android data binding. If you want this feature, please help us by designing the new annotations and parameters, and the generated code. Thanks!

smaugho:
The framework can detect if Data Binding is enabled, if so, I think that it should not only to use `Activity.setContentView.`, but @Bean and ViewById can use this mechanism as well.

[...]

PR #2019, opened 2017-06-26
Add Data Binding support

WonderCsabo:
Implements #1627 .

### DIFF
--- a/AndroidAnnotations/androidannotations-core/androidannotations-api/src/main/java/org/androidannotations/annotations/BindingObject.java
+++ b/AndroidAnnotations/androidannotations-core/androidannotations-api/src/main/java/org/androidannotations/annotations/BindingObject.java
@@ -1,0 +1,61 @@
+/**
+ * Copyright (C) 2016-2017 the AndroidAnnotations project
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may not
+ * use this file except in compliance with the License. You may obtain a copy of
+ * the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed To in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+ * License for the specific language governing permissions and limitations under
+ * the License.
+ */
+package org.androidannotations.annotations;
+
+import java.lang.annotation.ElementType;
+import java.lang.annotation.Retention;
+import java.lang.annotation.RetentionPolicy;
+import java.lang.annotation.Target;
+
+/**
+ * <p>
+ * Can be used on fields, methods and method parameters in enhanced Activities,
+ * Fragments and ViewGroups to access the binding object used in Data Binding.
+ * </p>
+ * <p>
+ * The declaring class must be annotated with {@link DataBound}. The type must
+ * extends <code>ViewDataBinding</code>. The following injections are allowed:
+ * </p>
+ * <blockquote>
+ *
+ * <pre>
+ * &#064;DataBound
+ * &#064;EActivity(R.layout.my_activity)
+ * public class MyActivity extends Activity {
+ *
+ * 	&#064;BindingObject
+ * 	MyActivityBinding binding;
+ *
+ * 	&#064;BindingObject
+ * 	void methodInjection(MyActivityBinding binding) {
+ * 		// use binding
+ * 	}
+ *
+ * 	void paramInjection(&#064;BindingObject MyActivityBinding binding) {
+ * 		// use binding
+ * 	}
+ *
+ * }
+ * </pre>
+ *
+ * </blockquote>
+ *
+ * @see DataBound
+ */
+@Retention(RetentionPolicy.CLASS)
+@Target({ ElementType.FIELD, ElementType.METHOD, ElementType.PARAMETER })
+public @interface BindingObject {
+}
--- a/AndroidAnnotations/androidannotations-core/androidannotations-api/src/main/java/org/androidannotations/annotations/DataBound.java
+++ b/AndroidAnnotations/androidannotations-core/androidannotations-api/src/main/java/org/androidannotations/annotations/DataBound.java
@@ -1,0 +1,47 @@
+/**
+ * Copyright (C) 2016-2017 the AndroidAnnotations project
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may not
+ * use this file except in compliance with the License. You may obtain a copy of
+ * the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed To in writing, software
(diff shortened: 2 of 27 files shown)


WonderCsabo:
@dodgex @smaugho 

I wonder maybe it is cleaner and better to access the binding object this way:
```java
@DataBound
@EActivity(R.layout.databound_activity)
public class DataBoundActivity extends Activity {

    @DataBinding
    DataBoundActivityBinding binding;

    @AfterViews
    void afterViews() {
      binding.setName("AA");
    }
}
```

smaugho:
Right now in Android annotations, for most of the injected data (@Bean, @FragmentArg, @Extra, ...) it is supported both, the normal injection in the field, and the method based injection, so I guess the best to keep the same logic will be to implement both methods (probably not parameter support, like for multiple parameters, since it is not the case here).. not sure if the InjectHelper would help for this right now, but probably it does, maybe some adjustment to avoid checking the multiple parameters injection.

WonderCsabo:
Yep, method injection would work.

WonderCsabo:
@smaugho @dodgex i changed the implementation to support both field and method injection.

AndroidAnnotations/androidannotations-core/androidannotations-api/src/main/java/org/androidannotations/annotations/BindingObject.java
```diff
@@ -0,0 +1,61 @@
...
+ * </p>
+ * <p>
+ * The declaring class must be annotated with {@link DataBound}. The type must
+ * extends <code>ViewDataBinding</code>. The following injections are allowed:
+ * </p>
+ * <blockquote>
+ *
+ * <pre>
+ * &#064;DataBound
+ * &#064;EActivity(R.layout.my_activity
```
dodgex: missing closing parenthesis.

AndroidAnnotations/androidannotations-core/androidannotations-api/src/main/java/org/androidannotations/annotations/DataBound.java
```diff
@@ -0,0 +1,47 @@
...
+ * Can be used on enhanced Activities, Fragments and ViewGroups to mark
+ * them to use the Data Binding library to inflate the layout.
+ * </p>
+ * <blockquote>
+ *
+ * Example :
+ *
+ * <pre>
+ * &#064;DataBound
+ * &#064;EActivity(R.layout.my_activity
```
dodgex: missing closing parenthesis

AndroidAnnotations/androidannotations-core/androidannotations/src/main/java/org/androidannotations/holder/DataBindingDelegate.java
```diff
@@ -0,0 +1,51 @@
...
+	JFieldVar getDataBindingField() {
+		if (dataBindingField == null) {
+			setDataBindingField();
+		}
+
+		return dataBindingField;
+	}
+
+	private void setDataBindingField() {
+		dataBindingField = holder.generatedClass.field(JMod.PRIVATE, getClasses().VIEW_DATA_BINDING, "viewDataBinding" + ModelConstants.generationSuffix());
```
dodgex: not sure but iirc we use static import for `generationSuffix` in other classes

AndroidAnnotations/androidannotations-core/androidannotations/src/main/java/org/androidannotations/holder/EComponentWithViewSupportHolder.java
```diff
@@ -44,6 +44,7 @@
 import com.helger.jcodemodel.JDefinedClass;
 import com.helger.jcodemodel.JExpr;
 import com.helger.jcodemodel.JFieldRef;
+import com.helger.jcodemodel.JFieldVar;
```
dodgex: looks like an unused import added?

WonderCsabo: `getDataBindingField()` uses this class as return type.

WonderCsabo:
I also noticed the missing header in R.java, but the maven license plugin did not add it so i did not bother.

AndroidAnnotations/androidannotations-core/androidannotations/src/main/java/org/androidannotations/holder/EComponentWithViewSupportHolder.java
dodgex: Damn i checked like 10 Times and yet missed it. :D

dodgex:
Okay, then no license header in `R.java`. :)

dodgex:
Thank you very much!

luongvo:
This issue should be released soon ASAP, I met this issue and finding another workaround way to pass this while waiting for a newer version :(

WonderCsabo:
@luongvo  You can use the snapshot version in your project.

AndroidAnnotations/androidannotations-core/androidannotations/src/test/java/org/androidannotations/databinding/R.java
```diff
@@ -0,0 +1,9 @@
+package org.androidannotations.databinding;
+
+public class R {
+
+	public static class layout {
+
+		public static final int activity_main = 0x7f06000a;
+	}
+}
```
luongvo: Why we need to push this file onto git? @WonderCsabo

WonderCsabo: Because this is not an Android project, and the R files are not generated, those are mock files.

luongvo:
@WonderCsabo ohhh, just got a newer snapshot from here https://oss.sonatype.org/content/repositories/snapshots/org/androidannotations/androidannotations/ Thanks 🚀

WonderCsabo:
@dodgex i pushed the documentation to my fork. https://github.com/WonderCsabo/androidannotations/wiki

dodgex:
Instead of `[...]` I'd prefer just a closing `}` as in our other samples. Also you might want to mention that method injection is supported too?

WonderCsabo:
@dodgex i updated the doc.

dodgex:
if you add the missing space in `void injectBinding2(@BindingObjectMainActivityBinding binding) {` you can merge this.

WonderCsabo:
Thanks, i merged it.

dodgex:
:D The space is at the wrong position now `void injectBinding2(@Binding ObjectMainActivityBinding binding) {`

WonderCsabo:
🤦‍♂️ Fixed it.

adroitandroid:
Hi @WonderCsabo thanks for this! Don't see 4.4 release tag, how can I use this?

WonderCsabo:
4.4 is not released, yet, but you can use the [snapshot](https://github.com/androidannotations/androidannotations/wiki/Building-Project-Gradle#snapshots). For `@EViewGroup`, you can use a normal root view in the layout, and it will work, however this will an unnecessary layer. BTW, databinding supports `<merge>`, despite the editor says it does not.

PR #2022, opened 2017-07-06
Support R2 class for resource values

WonderCsabo:
This PR adds support to ButterKnife R2 generator plugin. Please note only the SNAPSHOT version of that plugin generates all types of resources. For rationale, see the commit message.

### DIFF
--- a/AndroidAnnotations/androidannotations-core/androidannotations/src/main/java/org/androidannotations/internal/Options.java
+++ b/AndroidAnnotations/androidannotations-core/androidannotations/src/main/java/org/androidannotations/internal/Options.java
@@ -41,6 +41,7 @@ public class Options {
 		addSupportedOption(AndroidManifestFinder.OPTION_MANIFEST);
 		addSupportedOption(AndroidManifestFinder.OPTION_LIBRARY);
 		addSupportedOption(ProjectRClassFinder.OPTION_RESOURCE_PACKAGE_NAME);
+		addSupportedOption(ProjectRClassFinder.OPTION_USE_R2);
 		addSupportedOption(ModelConstants.OPTION_CLASS_SUFFIX);
 		addSupportedOption(FileAppender.OPTION_LOG_FILE);
 		addSupportedOption(LoggerContext.OPTION_LOG_LEVEL);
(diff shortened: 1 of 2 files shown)


dodgex:
Hey Csaba, thank you for the PR. I got 2 questions.

1) What is the purpose of the R2 class? what is the advantage over R?
2) (maybe already part of the answer to 1 but:) do we always want to use R2 if it is present?

WonderCsabo:
Hey, 

I tried to explain it in the commit message, but maybe it was not clear, sorry. 

1. The problem is with R class is that in case of library projects, the R.*.* fields are not final, so we cannot use them in annotation parameters. For the same reason, we also cannot read them for validation. To work around this, we added the `resName` parameter everywhere, which accepts a String. But that is inconvenient to use, you can have mistakes, there is no code completion etc. The R2 class has final fields, so it can be used in annotation parameters. But to actually make it work, we also have to read the values for validation, and we can only do that from the final R2 fields. This is what this PR does. 

2.Yes, maybe we should only the this if enabled by some option. I can put this behind an option flag.

dodgex:
Thank you for explaining it again. I did not look at the commit message, so i missed the explaination there.

~~But I'm a bit confused... We already have `@ViewById(R.id.myTextView)` so I'm not sure what the problem is with R and annotation parameters...~~ 

Okay, ignore that... reading your explaination a second time revels `in case of library projects`.

Maybe it is enough to put this behind the [library](https://github.com/androidannotations/androidannotations/wiki/CustomizeAnnotationProcessing#library) option?

WonderCsabo:
@dodgex i think we should put this behind a new flag.

WonderCsabo:
@dodgex updated.

dodgex:
looks fine.

WonderCsabo:
@dodgex documentation in my fork:
https://github.com/WonderCsabo/androidannotations/wiki/Library-projects#using-r2-class
https://github.com/WonderCsabo/androidannotations/wiki/CustomizeAnnotationProcessing#user2

dodgex:
looks good

WonderCsabo:
I merged the doc.

LuigiPapino:
This just saved me a lot of time. And will be very useful for instant apps.
It's working fine for me on a quiet complex project, with three layers (installedapp-> base -> sharedui).

PR #2023, opened 2017-07-07
Use normal R class even when validating with R2

WonderCsabo:
The final resource values will not be the same as the compile-time values,
so we have to use the R class in the generated code, even if the user set
to use the R2 class in annotation parameters.

Follow-up fix for:
https://github.com/androidannotations/androidannotations/pull/2022


dodgex:
Thank you for both PR :)

PR #2029, opened 2017-07-26
Update Kotlin example

WonderCsabo:
The manifest file is now found automatically, configuring it by hand
may cause issues.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -19,7 +19,7 @@ android:
     - tools # https://github.com/travis-ci/travis-ci/issues/6059
     - platform-tools
     - android-25
-    - build-tools-25.0.2
+    - build-tools-26.0.0
 
 before_script:
   - export TERM=dumb
--- a/examples/gradle/build.gradle
+++ b/examples/gradle/build.gradle
@@ -1,16 +1,16 @@
 buildscript {
     repositories {
-        mavenCentral()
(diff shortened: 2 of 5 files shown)


dodgex:
Thank you, could you also update the versions in the other samples? To have them in sync.

WonderCsabo:
https://github.com/androidannotations/androidannotations/issues/2028#issuecomment-318151146

Maybe because of this, we should not remove the kapt block from the example, because it fixes flavors with the stable Android Gradle plugin at least.

dodgex:
Maybe we should try to enhance the manifest finder to support the new(?) behaviour of `kapt` if the do not "fix" the [issue](https://youtrack.jetbrains.com/issue/KT-19245) you opened at kotlin.

I have an idea that could work, if i correctly understand the problem with the directories. I'll see if i get a hand on it this evening or tomorrow.

WonderCsabo:
@dodgex thanks!

dodgex:
[PR is up](https://github.com/androidannotations/androidannotations/pull/2030)

WonderCsabo:
@dodgex updated.

dodgex:
There is a travis issue due to the upgrade to 26 build tools. :/

WonderCsabo:
@dodgex build fixed.

Issue #2056, opened 2017-10-24
EView's onDetachedFromWindow call super.onAttachedToWindow not super.onDetachedFrom Window


RobinLJB:
I found a bug that when I create an EView and I overwrite onDetachedFromWindow method.

Then, the generated Class's `onDetachedFromWindow` method  will  call ` super.onAttachedToWindow`
instead of `super.onDetachedFromWindow` .

so, I lookup the package "androidannotations-4.3.0.jar"  and found that ,
in this class `org.androidannotations.holder.EViewHolder`
![source code](https://user-images.githubusercontent.com/23262690/31939373-b658f99a-b8ec-11e7-86cf-35ef7f888937.png)

then I fixed this method and change  `onAttachedToWindowMethod`  to `onDetachedFromWindowMethod`  locally  and it worked

**AndroidAnnotations version**:  4.3

**Android compile SDK version**:  26.0.1

your code
   @Override
    protected void onDetachedFromWindow() {
        bus.unregister(this);
        super.onDetachedFromWindow();
    }

**Expected generated code**:
 @Override
    public void onDetachedFromWindow() {
        getContext().unregisterReceiver(onActionDiscoveryStartedReceiver_);
        getContext().unregisterReceiver(onActionDiscoveryFinishedReceiver_);
        getContext().unregisterReceiver(onActionStateChangedReceiver_);
        super.onDetachedFromWindow();
    }

**Actual generated code**:
@Override
    public void onDetachedFromWindow() {
        getContext().unregisterReceiver(onActionDiscoveryStartedReceiver_);
        getContext().unregisterReceiver(onActionDiscoveryFinishedReceiver_);
        getContext().unregisterReceiver(onActionStateChangedReceiver_);
        super.onAttachedToWindow();
    }

WonderCsabo:
Nice catch, thanks!

PR #2062, opened 2017-10-30
Fix invoking wrong super method from onDetachedFromWindow

WonderCsabo:
#2056


dodgex:
You can merge this once you rebased - what should fix the build i think. :)

PR #2068, opened 2017-10-31
Add support for kotlin-allopen plugin

dodgex:
I should have applied the `@KotlinOpen` annotation to all annotations that override a class or method.

### DIFF
--- a/AndroidAnnotations/androidannotations-core/androidannotations-api/src/main/java/org/androidannotations/api/KotlinOpen.java
+++ b/AndroidAnnotations/androidannotations-core/androidannotations-api/src/main/java/org/androidannotations/api/KotlinOpen.java
@@ -1,0 +1,31 @@
+/**
+ * Copyright (C) 2016-2017 the AndroidAnnotations project
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may not
+ * use this file except in compliance with the License. You may obtain a copy of
+ * the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed To in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+ * License for the specific language governing permissions and limitations under
+ * the License.
+ */
+package org.androidannotations.api;
+
+import java.lang.annotation.ElementType;
+import java.lang.annotation.Retention;
+import java.lang.annotation.RetentionPolicy;
+import java.lang.annotation.Target;
+
+/**
+ * Utility annotation to easily support kotlin-allopen plugin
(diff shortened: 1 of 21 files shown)


dodgex:
I also updated the kotlin sample project to use the latest kotlin version and the allopen plugin. :)

dodgex:
Aaaaaaaaand... I just realized I missed the Plugin annotations once again... :) I have to leave now, but will add them as soon as I'm back home. :D

WonderCsabo:
Hmm, maybe it is not a good idea, because users not having this plugin will face problems. 

Lets create a method, then, which is not `open`, and uses this mechanism, and add KotlinDoc to it explaining this.

examples/kotlin/src/main/kotlin/org/androidannotations/gradle/activity/HelloAndroidActivity.kt
```diff
@@ -43,3 +44,19 @@ open class HelloAndroidActivity : Activity() {
         helloTextView.text = helloMessage
     }
 }
+
+/**
+ * As of Kotlin 1.0.6+ and AndroidAnnotations 4.4.0, you can use the kotlin-allopen plugin
+ * to remove the need to explicitly declare enhanced classes or methods as <code>open</code>.
+ *
+ * See the build.gradle file or [the Kotlin Documentation](https://kotlinlang.org/docs/reference/compiler-plugins.html#all-open-compiler-plugin)
```
WonderCsabo: I would rather link to our kotlin suppor twiki page. We are not having one, but we should open it. :)
(We should list the `KotlinOpen` annotation there for example.)

Also we could add some gotchas for Kotlin, like `lateinit` and `@JvmField`.

dodgex:
I added a link to https://github.com/androidannotations/androidannotations/wiki/Kotlin 😉 instead of the kotlin docs

AndroidAnnotations/androidannotations-core/androidannotations-api/src/main/java/org/androidannotations/api/KotlinOpen.java
```diff
@@ -0,0 +1,31 @@
...
+
+import java.lang.annotation.ElementType;
+import java.lang.annotation.Retention;
+import java.lang.annotation.RetentionPolicy;
+import java.lang.annotation.Target;
+
+/**
+ * Utility annotation to easily support kotlin-allopen plugin
+ *
+ * @see "https://kotlinlang.org/docs/reference/compiler-plugins.html#all-open-compiler-plugin"
```
WonderCsabo: Lets add the AA wiki here as well.

dodgex: Done. I replaced the link and added it to the wiki page

WonderCsabo:
Thanks!

Issue #2048, opened 2017-09-22
IllegalStateException with AndroidAnnotations 4.1 and @EViewGroup / @OrmLiteDao


ytakahashi1981:
I have the same issue as #1887.
I updated #1887, but there is no response maybe due to issue status(closed).

I cannot upgrade AA to 4.x because of upgration cost with this issue...
(many many views are using OrmLiteDao annotations in my product...)

Maybe you can reproduce with these simple sources.
(My AA version is 4.3.1 but reproduced with 4.1)

TestDao.java
```TestDao.java
public class TestDao extends BaseDaoImpl<TestEntity, Integer> {
	protected TestDao(ConnectionSource connectionSource, Class<TestEntity> dataClass) throws SQLException {
		super(connectionSource, dataClass);
	}
}
```

TestEntity.java
```TestEntity.java
@DatabaseTable(tableName = "Test", daoClass = TestDao.class)
public class TestEntity {
}
```

TestView.java
```TestView.java

@EViewGroup
public class TestView extends FrameLayout {

	@OrmLiteDao(helper = DBHelper.class)
	TestDao testDao;

	public TestView(@NonNull Context context) {
		super(context);
	}

}
```

DBHelper.java
```DBHelper.java
public class DBHelper extends OrmLiteSqliteOpenHelper {
	public DBHelper(Context context) {
		super(context, "test.db", null, 1);
	}

	@Override
	public void onCreate(SQLiteDatabase database, ConnectionSource connectionSource) {}

	@Override
	public void onUpgrade(SQLiteDatabase database, ConnectionSource connectionSource, int oldVersion, int newVersion) {}
}
```

TomWangTW:
Hi,
Have you try this?
For using OrmLiteDao, addd these new dependencies in your build.gradle:

compile "org.androidannotations:ormlite-api:4.0.0"
apt "org.androidannotations:ormlite:4.0.0"

Change your imports to
org.androidannotations.ormlite.annotations.OrmLiteDao

https://github.com/androidannotations/androidannotations/wiki/4.0.0-migration-guide

ytakahashi1981:
@TomWangTW 
Hi. Thanks for reply.
These dependencies are already added.

I can use @OrmLiteDao with @EActivity, @EFragment, @EBean classes, but, can't use only with @EViewGroup due to IllegalStateException(like #1887, the error appears in compile time).

ytakahashi1981:
I found what is wrong (maybe).

[...]

PR #2069, opened 2017-10-31
Fix IllegalStateException for @OrmLiteDao in view

WonderCsabo:
#2048

This commit contains a bigger refactor as well: the previous
implementation contained HasLifecycleMethods interface, which was
implemented by most all the components, even if they do not have
the specific methods. This is now changed.

### DIFF
--- a/AndroidAnnotations/androidannotations-core/androidannotations/src/main/java/org/androidannotations/holder/EServiceHolder.java
+++ b/AndroidAnnotations/androidannotations-core/androidannotations/src/main/java/org/androidannotations/holder/EServiceHolder.java
@@ -1,5 +1,6 @@
 /**
  * Copyright (C) 2010-2016 eBusiness Information, Excilys Group
+ * Copyright (C) 2016-2017 the AndroidAnnotations project
  *
  * Licensed under the Apache License, Version 2.0 (the "License"); you may not
  * use this file except in compliance with the License. You may obtain a copy of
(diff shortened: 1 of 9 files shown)


AndroidAnnotations/androidannotations-core/androidannotations/src/main/java/org/androidannotations/internal/core/handler/ReceiverHandler.java
```diff
@@ -138,22 +140,35 @@ private JFieldVar createReceiverField(HasReceiverRegistration holder, String rec
...
-			unregisterBlock = holder.getOnStopBeforeSuperBlock();
-			break;
-		case OnResumeOnPause:
-			registerBlock = holder.getOnResumeAfterSuperBlock();
-			unregisterBlock = holder.getOnPauseBeforeSuperBlock();
-			break;
-		case OnAttachOnDetach:
-			registerBlock = holder.getOnAttachAfterSuperBlock();
-			unregisterBlock = holder.getOnDetachBeforeSuperBlock();
+
```
WonderCsabo: This code is now rather ugly, but this is the price for removing the invalid abstraction `HasLifecycleMethods`. Do you have a better idea? @dodgex

AndroidAnnotations/androidannotations-core/androidannotations/src/main/java/org/androidannotations/holder/EServiceHolder.java
```diff
@@ -100,46 +103,18 @@ public JBlock getIntentFilterInitializationBlock(IntentFilterData intentFilterDa
 		return getInitBodyInjectionBlock();
 	}
 
+
```
dodgex: unnecessary newline? :)

AndroidAnnotations/androidannotations-core/androidannotations/src/main/java/org/androidannotations/internal/core/handler/ReceiverHandler.java
dodgex: Currently i cant think of a better way to solve this. :/

AndroidAnnotations/androidannotations-core/androidannotations/src/main/java/org/androidannotations/holder/EServiceHolder.java
WonderCsabo: Fixed.

dodgex:
Thanks

Issue #2143, opened 2018-06-09
Infinite loop when a nested element is valid but the enclosing one not


dodgex:
**AndroidAnnotations version**: 4.4.0

Example: Having a validated `@EBean` in an invalidated `@EFragment` causes the generation to loop infinite. This happens because the bean requires the fragment class to be generated, but as the fragment never gets generated the bean waits forever.

Note: Already working on a fix.

PR #2146, opened 2018-06-09
Fix infinite loop when enclosing class is invalid

dodgex:
This fixes the infinite loop mentioned in #2143


dodgex:
The build fails due to invalid license header as i already used 2018 but the plugin checks for 2017 until https://github.com/androidannotations/androidannotations/pull/2142 is merged.

dodgex:
This should be merged after #2148 and the log call in `ModelProcessor` needs to be updated.

Issue #2145, opened 2018-06-09
Logging broken when first argument is an element


dodgex:
While working on https://github.com/androidannotations/androidannotations/issues/2143 I found that for some places the logging might be broken.

e.g.
https://github.com/androidannotations/androidannotations/blob/074df636fbd482cea3f797eece296524a877fd67/AndroidAnnotations/androidannotations-core/androidannotations/src/main/java/org/androidannotations/internal/process/ModelProcessor.java#L203

results in the message:
> 17:54:26.825 [Task worker for ':' Thread 2] ERROR o.a.i.p.ModelProcessor:203 - Skip element  because it's abstract

This happens as the signatures of the trace methods are
```
void trace(String message, Object... args);
void trace(String message, Element element, Object... args);
```
Instead of using the first (and in this case desired version) the compiler uses the second version of `trace` as the first parameter after the message is an `Element` adn so it is not usedas replacement for the `{}` the log message but the message (without a valuefor `{}`) is attached to the element.

I think we should review all places where we use these logs and possibly even refactor the logger to avoid similar issues. maybe move the `element` in front of the string parameter?

@WonderCsabo wdyt?

dodgex:
`trace` is only called twice. both times it is not intended to attach the message to the element. I'll remove that 3-arg version of trace at it imho makes no sense to have it (can we actually have trace messages on elements?!)

dodgex:
same for `debug`

dodgex:
aaaaaaaand `info`

PR #2148, opened 2018-06-10
Fix logging and prevent similar issues

dodgex:
This PR is intended to fix the logging issue mentioned in #2145 and prevent similar misuse in the future.

While the only place where this issue actually hapened where two `trace()` statements in `ModelProcessor` I refactored the log methods to have the `message` argument last or second to last if there are `args` allowed.


dodgex:
The build fails due to invalid license header as i already used 2018 but the plugin checks for 2017 until #2142 is merged.

dodgex:
This needs to be merged BEFORE #2146 as there is a new log statement that needs to be updated to the changes of this PR

PR #2156, opened 2018-07-20
comment unused variable

khalil2535:


khalil2535:
hmm little strange , sorry I'm still leaning …

PR #2167, opened 2018-10-13
Check a full name instead of just the classSuffix

dodgex:
... as a single '_' is no longer valid in JDK 11 (and mabye 10?)


AndroidAnnotations/androidannotations-core/androidannotations/src/main/java/org/androidannotations/helper/ModelConstants.java
```diff
@@ -57,7 +57,7 @@ public static void init(AndroidAnnotationsEnvironment environment) {
 			throw new IllegalArgumentException("'" + classSuffix + "' may not be an empty string.");
 		}
 
-		if (!SourceVersion.isName(classSuffix) || classSuffix.contains(".")) {
+		if (!SourceVersion.isName("validName" + classSuffix) || classSuffix.contains(".")) {
```
WonderCsabo: I would capitalize ValidName.

dodgex:
Updated :)

Issue #2168, opened 2018-10-13
Crash during make project: IllegalArgumentException: The expression must be false but it is not: trying to create the same field twice: fragmentSupport_


Namnodorel:
**AndroidAnnotations version**: 4.5.0

**Android compile SDK version**: 28

**Annotated code**:
```java
@EBean
public class MyClass{
    //Doesn't use any AA features, it's just annotated so it can be injected somewhere else as an attribute
    public MyClass(){
        //...
    }
    //...
}
```

**Stacktrace in case of AA crash**:

```
error: Something went wrong: Unexpected error in AndroidAnnotations 4.5.0!
You should check if there is already an issue about it on https://github.com/androidannotations/androidannotations/search?q=java.lang.IllegalArgumentException&type=Issues
If none exists, please open a new one with the following content and tell us if you can reproduce it or not. Don't forget to give us as much information as you can (like parts of your code in failure).
Java version: javac 1.8.0_92
Javac processors options: androidManifestFile=app\build\intermediates\merged_manifests\debug\AndroidManifest.xml
Stacktrace: java.lang.IllegalArgumentException: The expression must be false but it is not: trying to create the same field twice: fragmentSupport_
at com.helger.jcodemodel.util.JCValueEnforcer.isFalse(JCValueEnforcer.java:144)
at com.helger.jcodemodel.JDefinedClass.field(JDefinedClass.java:402)
at com.helger.jcodemodel.JDefinedClass.field(JDefinedClass.java:375)
at org.androidannotations.internal.core.helper.ActivityIntentBuilder.addFragmentConstructor(ActivityIntentBuilder.java:116)
at org.androidannotations.internal.core.helper.ActivityIntentBuilder.createAdditionalConstructor(ActivityIntentBuilder.java:111)
at org.androidannotations.internal.core.helper.ActivityIntentBuilder.build(ActivityIntentBuilder.java:71)
at org.androidannotations.holder.EActivityHolder.<init>(EActivityHolder.java:115)
at org.androidannotations.internal.core.handler.EActivityHandler.createGeneratedClassHolder(EActivityHandler.java:47)
at org.androidannotations.internal.core.handler.EActivityHandler.createGeneratedClassHolder(EActivityHandler.java:39)
at org.androidannotations.internal.process.ModelProcessor.generateElements(ModelProcessor.java:203)
at org.androidannotations.internal.process.ModelProcessor.process(ModelProcessor.java:74)
[...]

PR #2149, opened 2018-06-10
Add support for AndroidX classes

dodgex:
Do NOT merge yet. Needs to be rebased after my other recent PRs have been merged as this contains thíer changes too.

The intention is to add support for new AndroidX classes in addition to support library. 

Checklist of classes I consider done.
- [x] ANDROIDX_ACTIVITY_COMPAT
- [x] ANDROIDX_FRAGMENT
- [x] ANDROIDX_FRAGMENT_ACTIVITY
- [x] ANDROIDX_LOCAL_BROADCAST_MANAGER
- [x] ANDROIDX_CONTEXT_COMPAT
- [x] ANDROIDX_PREFERENCE
- [x] ANDROIDX_PREFERENCE_FRAGMENT
- [x] ANDROIDX_PREFERENCE_FRAGMENTCOMPAT
- [x] ANDROIDX_PREFERENCE_CLICK_LISTENER
- [x] ANDROIDX_PREFERENCE_CHANGE_LISTENER
- [x] ANDROIDX_APPCOMPAT_ACTIVITY
- [x] ANDROIDX_VIEW_PAGER
- [x] ANDROIDX_PAGE_CHANGE_LISTENER

@WonderCsabo If you want to support me, feel free to create tests for classes I have checked. ;)

PR #2169, opened 2018-10-13
Fix the order of some Android X checks to avoid conflicts with Support v4

dodgex:
See #2168 

A project that is migrated to Android X no longer has support v4 classes on its classpath... at least it seems so.

One can have a project migrated to Android X and thus is no longer able to import e.g. `android.support.v4.app.Fragment` in own code. If no 3rd party library is using the support v4 fragment compiling with AA 4.5.0 is no problem. Having a 3rd party dependency that uses support v4 does not mean you can use the classes in your code as there is a tool called `jetifier` that migrates the 3rd party code on the fly to use Android X. but during the actual build process the support v4 classes are still on classpath and so available to AndroidAnnotations.

In the Android X support PR #2149 this was not known and thus we have some cases where the build fails with android x + support v4 libs on classpath.

This PR aims to fix the issue by changing order of checks so that Android X dominates support v4.


WonderCsabo:
Thanks for the quick fix and detailed description!

PR #2175, opened 2018-10-26
Support InstantApps, including for 3.3-beta01

dodgex:
Add support for Instant Apps (https://github.com/androidannotations/androidannotations/issues/2017)

While it seems that Instant Apps should already be supported for version 3.1.x of the Android Gradle Plugin. But with 3.2.x and 3.3.x changes have been made and this PR aims to ensure Instant App support for those plugins too.

For instant apps it is now required to set `"instantAppFeature": "true"` in `annotationProcessorOptions`, similar to library projects.


dodgex:
Hey @WonderCsabo, would be awesome if you could do an initial review and tell me if you are okay with the way i solved the issue. :)

WonderCsabo:
Looks good, thanks!

WonderCsabo:
Please add this option to the wiki.

dodgex:
https://github.com/androidannotations/androidannotations/wiki/CustomizeAnnotationProcessing#instantappfeature

WonderCsabo:
Great, thanks.

PR #2185, opened 2018-12-06
Add @KotlinOpen to class annotations only

WonderCsabo:
The kotlin-open is making the class and the members as well public,
so adding this annotation to member AA annotation is not needed has
no effect.


dodgex:
thanks :)

PR #2196, opened 2019-01-29
Fix failing Robolectric tests

WonderCsabo:


dodgex:
Is it correct that the PR only changes the version?

WonderCsabo:
Yes.

PR #2199, opened 2019-01-30
Update license header for 2019

WonderCsabo:


AndroidAnnotations/androidannotations-core/androidannotations-api/pom.xml
```diff
@@ -1,8 +1,7 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <!--
 
-    Copyright (C) 2010-2016 eBusiness Information, Excilys Group
```
dodgex: I think we should not remove the old excilys copyright.

AndroidAnnotations/androidannotations-core/androidannotations-api/src/main/java/org/androidannotations/annotations/Background.java
```diff
@@ -1,6 +1,5 @@
 /**
- * Copyright (C) 2010-2016 eBusiness Information, Excilys Group
```
dodgex: I think we should not remove the old excilys copyright.

dodgex:
i added two comments, but it seems you have removed all of the excilys copyright notes. unfortunately we are not allowed to do this. :(

WonderCsabo:
I did not mean to do it. I just changed header_both.txt and header_new.txt with 2019, and hit `mvn license:format`. What is the correct way?

dodgex:
I always used grep(win) to replace the old with New Version string

WonderCsabo:
@dodgex hopefully it is correct now, please check it again!

dodgex:
Looks good, maybe run the `check_missing_copyright.sh` once to ensure there are no files changed since we took the project that miss our copyright. if there are none feel free to merge.

WonderCsabo:
@dodgex there are loads of files with missing copyright. :( How do you add them? BTW, i forgot about this script, it is great!

dodgex:
@WonderCsabo honestly... i always did it by hand.

WonderCsabo:
@dodgex i added the new copyright file for all files reported by `check_missing_copyright.sh`. I think we should incorporate that into our CI build, to make sure those are added immediately in PRs.

dodgex:
@WonderCsabo something is not right i think. Your second commit adds 239 lines and I'm pretty sure that most of those files where not changed in the last 2 years. I'd say that at most 10 files should be missing our copyright header.

If i get time this evening I'll check what the script output is for me at home.

WonderCsabo:
@dodgex these files were emmitted by the script. I think they are indeed changed, because we reformatted the whole project.

dodgex:
@WonderCsabo you are right, when i checked i was on my forks develop branch that did not have the latest changes from the main repo.

Issue #1636, opened 2015-11-23
Add @RootFragment annotation for @EBean


LouisCAD:
Hi!

Currently, the only way to access the parent `Fragment` in an `@EBean` is by writing a setter in the bean manually, or by making a `@FragmentById`/`@FragmentByTag` public in the `@RootContext`annotated `Activity`. The code is not that beautiful, so why not create an `@RootFragment` annotation?

It would work the same way as `@RootContext`, excepted that it would not need to be re-injected if the Fragment instance is retained (see `setRetainInstance()` & `getRetainInstance()`) and the field is annotated with `NonConfigurationInstance` when config changes occurs (screen rotation, locale change, etc) as the Fragment itself (only it's View hierarchy) is not destroyed if retained.

Examples speak better than words, so here's how it would look like:

``` java
@EBean
public class MyRecyclerView extends RecyclerView.Adapter {

    @RootFragment
    MyEFragment mFragment;

    // I can now access the fragment, it's @InstanceState members,
    // and keep easily my adapter state, such as selected items
}
```

Below are the quotes to "help" implementation, from Gitter:

> - @WonderCsabo: the problem is that it is not trivial to implement
> - @LouisCAD: What would be the hard part(s)? Wouldn't it work by adding a setFragment_(MyEFragment fragment) call in onCreate() for all @EBeans in MyEFragment containing an @RootFragment field?
> - @WonderCsabo: Yeah, it is not too hard, that was an overstatement.

I'm on the way to learn how Annotation Processors work under the hood, so I hope I'll be able to add this myself (If I begin working on this, I'll post updates here), but if in the meantime, someone with the skills wants to add this new feature, I'd be very grateful (and I hope I would not be the only one).

gauravsak:
@WonderCsabo I think this should be marked as an enhancement. Is that right?

smaugho:
About this enhancement, we know that `@EBean` can have a the Singleton scope, in which case, I think it doesn't make sense that it'll be able to Inject a Fragment. 

So @WonderCsabo, @dodgex confirm me please if an implementation like this would work:

1. `@RootFragment` can only be injected in `@EBean` which aren't Singleton. 

I was thinking in using "setFragment_" on the Bean as proposed above, but that has a problem, when the "AfterInject" is called, the fragment will be no assigned yet (since it is called when the Bean `_getInstance(Context)` is called. A better solution would be then:

[...]

PR #2202, opened 2019-02-14
Implements @RootFragment annotation support.

smaugho:
I've implemented the support for a `@RootFragment` which solves the issue #1636 . Please let me know about any question/suggestion regarding the solution. 

Thanks.

### DIFF
--- a/AndroidAnnotations/androidannotations-core/androidannotations-api/src/main/java/org/androidannotations/annotations/RootFragment.java
+++ b/AndroidAnnotations/androidannotations-core/androidannotations-api/src/main/java/org/androidannotations/annotations/RootFragment.java
@@ -1,0 +1,55 @@
+/**
+ * Copyright (C) 2016-2019 the AndroidAnnotations project
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may not
+ * use this file except in compliance with the License. You may obtain a copy of
+ * the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed To in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+ * License for the specific language governing permissions and limitations under
+ * the License.
+ */
+package org.androidannotations.annotations;
+
+import java.lang.annotation.ElementType;
+import java.lang.annotation.Retention;
+import java.lang.annotation.RetentionPolicy;
+import java.lang.annotation.Target;
+
+/**
+ * <p>
+ * Use it on Fragment fields in an {@link EBean} annotated classes to inject the
+ * Fragment which originally injected the bean instance.
+ * </p>
+ * <p>
+ * This field may not be injected at runtime if the fragment used to create the
+ * bean is not of the appropriate type, or if the bean is injected in an
+ * activity.
+ * </p>
+ * <blockquote>
+ *
+ * Example :
+ *
+ * <pre>
+ * &#064;EBean
+ * public class MyClass {
+ * 
+ * 	&#064;RootFragment
+ * 	Fragment fragment;
+ * 
+ * 	// Only injected if the root fragment implements a given interface
+ * 	&#064;RootFragment
+ * 	MyInterface fragment;
+ * }
+ * </pre>
+ *
+ * </blockquote>
+ */
+@Retention(RetentionPolicy.CLASS)
+@Target({ ElementType.FIELD, ElementType.METHOD, ElementType.PARAMETER })
+public @interface RootFragment {
+}
--- a/AndroidAnnotations/androidannotations-core/androidannotations-testutils/src/main/java/org/androidannotations/testutils/ProcessorTestHelper.java
+++ b/AndroidAnnotations/androidannotations-core/androidannotations-testutils/src/main/java/org/androidannotations/testutils/ProcessorTestHelper.java
@@ -293,6 +293,10 @@ public class ProcessorTestHelper {
 				addCollection(files, (Type) element);
 			} else if (element instanceof String) {
 				files.add(new File((String) element));
+			} else if (element instanceof String[]) {
+				for (String subElement : (String[]) element) {
+					files.add(new File(subElement));
+				}
 			}
 		}
 		return compileFiles(files);
--- a/AndroidAnnotations/androidannotations-core/androidannotations/src/main/java/org/androidannotations/holder/EBeanHolder.java
+++ b/AndroidAnnotations/androidannotations-core/androidannotations/src/main/java/org/androidannotations/holder/EBeanHolder.java
@@ -42,15 +42,18 @@ public class EBeanHolder extends EComponentWithViewSupportHolder {
 
 	public static final String GET_INSTANCE_METHOD_NAME = "getInstance" + generationSuffix();
 
+	private JFieldVar rootFragmentField;
 	private JFieldVar contextField;
+
 	private JMethod constructor;
+	private JMethod overloadedConstructor;
 
 	public EBeanHolder(AndroidAnnotationsEnvironment environment, TypeElement annotatedElement) throws Exception {
 		super(environment, annotatedElement);
-		setConstructor();
+		setConstructors();
 	}
 
-	private void setConstructor() {
+	private void setConstructors() {
 		constructor = generatedClass.constructor(PRIVATE);
 		JVar constructorContextParam = constructor.param(getClasses().CONTEXT, "context");
 		JBlock constructorBody = constructor.body();
@@ -60,6 +63,16 @@ public class EBeanHolder extends EComponentWithViewSupportHolder {
 			constructorBody.add(JExpr.invokeSuper().arg(constructorContextParam));
 		}
 		constructorBody.assign(getContextField(), constructorContextParam);
+
+		overloadedConstructor = generatedClass.constructor(PRIVATE);
+		JVar overloadedConstructorContextParam = overloadedConstructor.param(getClasses().CONTEXT, "context");
+		JVar overloadedConstructorRootFragmentParam = overloadedConstructor.param(getClasses().OBJECT, "rootFragment");
+		JBlock overloadedConstructorBody = overloadedConstructor.body();
+		if (superConstructor.getParameters().size() == 1) {
+			overloadedConstructorBody.add(JExpr.invokeSuper().arg(constructorContextParam));
+		}
+		overloadedConstructorBody.assign(getContextField(), overloadedConstructorContextParam);
+		overloadedConstructorBody.assign(getRootFragmentField(), overloadedConstructorRootFragmentParam);
 	}
 
 	public JFieldVar getContextField() {
@@ -69,9 +82,21 @@ public class EBeanHolder extends EComponentWithViewSupportHolder {
 		return contextField;
 	}
 
+	public JFieldVar getRootFragmentField() {
+		if (rootFragmentField == null) {
+			rootFragmentField = generatedClass.field(PRIVATE, getClasses().OBJECT, "rootFragment" + generationSuffix());
+		}
+		return rootFragmentField;
+	}
+
 	@Override
 	protected void setContextRef() {
 		contextRef = getContextField();
+	}
+
+	@Override
+	protected void setRootFragmentRef() {
+		rootFragmentRef = getRootFragmentField();
 	}
 
 	@Override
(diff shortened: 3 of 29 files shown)


AndroidAnnotations/androidannotations-core/androidannotations/src/main/java/org/androidannotations/holder/EBeanHolder.java
```diff
@@ -40,15 +40,18 @@
 
 	public static final String GET_INSTANCE_METHOD_NAME = "getInstance" + generationSuffix();
 
+	private JFieldVar rootFragmentField;
```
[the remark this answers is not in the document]
smaugho: all copyrights solved on this PR as well

AndroidAnnotations/androidannotations-core/androidannotations-api/src/main/java/org/androidannotations/annotations/RootFragment.java
```diff
@@ -0,0 +1,54 @@
+/**
+ * Copyright (C) 2010-2016 eBusiness Information, Excilys Group
```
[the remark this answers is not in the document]
smaugho: all copyrights solved on this PR as well

AndroidAnnotations/androidannotations-core/androidannotations/src/main/java/org/androidannotations/internal/core/handler/RootFragmentHandler.java
```diff
@@ -0,0 +1,92 @@
...
+
+	private final InjectHelper<EBeanHolder> injectHelper;
+
+	public RootFragmentHandler(AndroidAnnotationsEnvironment environment) {
+		super(RootFragment.class, environment);
+		injectHelper = new InjectHelper<>(validatorHelper, this);
+	}
+
+	@Override
+	public void validate(Element element, ElementValidation validation) {
```
WonderCsabo: I was wondering we should allow only Fragment extending types, however i see that you deliberately want to inject custom interfaces as well. `@RootContext` only allows types extending `Context` to be injected. I think we should do similar here.

WonderCsabo:
@smaugho please add compile time tests to verify validation and that correct code compiles, and runtime tests as well.

AndroidAnnotations/androidannotations-core/androidannotations/src/main/java/org/androidannotations/internal/core/handler/RootFragmentHandler.java
smaugho: Well, we could definitely check if it extends Fragment (support one, or system one)... I guess the annotation `@RootFragment` would be more clear then... I definitely would like to inject interfaces, though I could live with not being able to do it on this annotation ;)

In fact, I added an annotation `@RootView`  to the Plugin I'm building, which basically will use the "fragment or context activity" past to the EBean, and that one will support using Interfaces then. 

From an architecture point of view, it makes sense then, it would be quite usefull for implementing MVP and MVVM (though with some improvements to avoid holding a real instance of the activity/fragment on the beans). 

Well, this `@RootView` could be implemented directly on AA, but I think it will make more sense on this Plugin handling architectural patterns, but let me know if you think is useful on AA itself.

WonderCsabo: Let's require `Fragment` here then.

AndroidAnnotations/androidannotations-core/androidannotations/src/test/java/org/androidannotations/rootfragment/RootFragmentTest.java
```diff
@@ -0,0 +1,66 @@
...
+	}
+
+	@Test
+	public void eBeanOnCustomFragmentCompiles() {
+		assertCompilationSuccessful(compileFiles(SomeBeanWithRootFragmentWithCustomFragment.class));
+	}
+
+	@Test
+	public void eBeanOnCustomSupportFragmentsCompiles() {
+		assertCompilationSuccessful(compileFiles(defPath("support/Fragment.java", "CustomSupportFragment.java", "SomeBeanWithRootFragmentWithCustomSupportFragment.java")));
```
WonderCsabo: Why is this needed? I see similar tests, for example `ActivityIntentFragmentTest`, which is using `toPath()`.

AndroidAnnotations/androidannotations-core/androidannotations-testutils/src/main/java/org/androidannotations/testutils/ProcessorTestHelper.java
```diff
@@ -293,6 +293,10 @@ public CompileResult compileFiles(Object... elements) {
 				addCollection(files, (Type) element);
 			} else if (element instanceof String) {
 				files.add(new File((String) element));
+			} else if (element instanceof String[]) {
```
WonderCsabo: Why is this needed?

AndroidAnnotations/androidannotations-core/androidannotations/src/test/java/org/androidannotations/rootfragment/RootFragmentTest.java
WonderCsabo:
```java
CompileResult result = compileFiles(toPath(ActivityIntentFragmentTest.class, "support/Fragment.java"), //
				toPath(ActivityIntentFragmentTest.class, "support/ActivityCompat.java"), ActivityInManifest.class);
```

Why did you add the `defPath()` method?

AndroidAnnotations/androidannotations-core/androidannotations-test/src/test/java/org/androidannotations/test/rootfragment/RootFragmentTest.java
```diff
@@ -0,0 +1,74 @@
...
+import org.junit.runner.RunWith;
+import org.robolectric.Robolectric;
+import org.robolectric.RobolectricTestRunner;
+
+@RunWith(RobolectricTestRunner.class)
+public class RootFragmentTest {
+
+	@Test
+	public void injectFragmentWhichInjectedBean() {
+		CustomFragment fragment = new CustomFragment_();
```
WonderCsabo: I would create the `Fragment` properly using `Robolectric`:
```java
CustomFragment fragment = Robolectric.buildFragment(CustomFragment_.class).create().get();
```

This applies to other test methods as well.

AndroidAnnotations/androidannotations-core/androidannotations-test/src/main/java/org/androidannotations/test/rootfragment/ActivityWithBeansWithRootFragment.java
```diff
@@ -0,0 +1,29 @@
...
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed To in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+ * License for the specific language governing permissions and limitations under
+ * the License.
+ */
+package org.androidannotations.test.rootfragment;
```
WonderCsabo: Please move all this tests under `ebean` package. (`@RootContext` is also tested under that).

AndroidAnnotations/androidannotations-core/androidannotations/src/test/java/org/androidannotations/rootfragment/CustomFragment.java
```diff
@@ -0,0 +1,22 @@
...
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed To in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+ * License for the specific language governing permissions and limitations under
+ * the License.
+ */
+package org.androidannotations.rootfragment;
```
WonderCsabo: Please move all the compile-time tests as well under `ebean` package.

AndroidAnnotations/androidannotations-core/androidannotations/src/main/java/org/androidannotations/holder/EBeanHolder.java
```diff
@@ -58,6 +62,16 @@ private void setConstructor() {
...
 		}
 		constructorBody.assign(getContextField(), constructorContextParam);
+
+		overloadedConstructor = generatedClass.constructor(PRIVATE);
+		JVar overloadedConstructorContextParam = overloadedConstructor.param(getClasses().CONTEXT, "context");
+		JVar overloadedConstructorRootFragmentParam = overloadedConstructor.param(getClasses().OBJECT, "rootFragment");
+		JBlock overloadedConstructorBody = overloadedConstructor.body();
+		if (superConstructor.getParameters().size() == 1) {
+			overloadedConstructorBody.invoke("super").arg(constructorContextParam);
+		}
```
WonderCsabo: Travis fails because of this i think. You have to write:
```java
overloadedConstructorBody.add(JExpr.invokeSuper().arg(constructorContextParam));
```

AndroidAnnotations/androidannotations-core/androidannotations/src/test/java/org/androidannotations/rootfragment/RootFragmentTest.java
smaugho: @WonderCsabo , "defPath" accepts multiple paths (which is needed when having several "support-dependent files".. Also, it simplifies having to pass "the current class" per each call to "toPath", it is a little improvement, without it, when you have something like three files (like in that example above), basically you have to make a huge line with "toPath" and passing the test class to that method  (so 3 times). That doesn't seem that clean. 

"def" just assumes that the "package-holder class" is the same than the one of the test

AndroidAnnotations/androidannotations-core/androidannotations-testutils/src/main/java/org/androidannotations/testutils/ProcessorTestHelper.java
smaugho: @WonderCsabo , this comes with the function I explained above, "defPath", which basically returns an array of strings (multiple paths), so on this one, that is simply parsed to get each one of the paths and add them to the files.

AndroidAnnotations/androidannotations-core/androidannotations-test/src/test/java/org/androidannotations/test/rootfragment/RootFragmentTest.java
smaugho: Yeap, I tried to follow how I saw it in others.. mm.. but if the Fragment is not "added" in any FragmentManager, I don't think it would make that difference using Robolectric for it... or is there any?

AndroidAnnotations/androidannotations-core/androidannotations-test/src/main/java/org/androidannotations/test/rootfragment/ActivityWithBeansWithRootFragment.java
smaugho: That explains me why I didn't see any for @RootContext, moving them will required a little refactorization 😅 .

AndroidAnnotations/androidannotations-core/androidannotations-test/src/test/java/org/androidannotations/test/rootfragment/RootFragmentTest.java
WonderCsabo: Robolectric is adding it to a `FragmentManager`, using the method i described above.

AndroidAnnotations/androidannotations-core/androidannotations/src/test/java/org/androidannotations/rootfragment/RootFragmentTest.java
WonderCsabo: I see it now, thanks for improving it!

AndroidAnnotations/androidannotations-core/androidannotations-testutils/src/main/java/org/androidannotations/testutils/ProcessorTestHelper.java
WonderCsabo: I understand, thanks.

AndroidAnnotations/androidannotations-core/androidannotations-test/src/test/java/org/androidannotations/test/rootfragment/RootFragmentTest.java
smaugho: Ah got it, if so then let's do it like that, since that will bring it closer to the real runtime scenario.

smaugho:
@WonderCsabo Travis failed this time but seems to be due to some missing copyrights in some other files

smaugho:
I will add them..

WonderCsabo:
Thanks, that was my fault. I forgot to rebase the branch to include the license check...

WonderCsabo:
BTW, next time please create separate commits for unrelated changes such as this.

WonderCsabo:
This depends on https://github.com/androidannotations/androidannotations/pull/2211 , right?

smaugho:
Right now yes 😅

dodgex:
https://github.com/androidannotations/androidannotations/pull/2211 has been merged.

smaugho:
I will rebase this branch, to simplify its revision.

WonderCsabo:
@smaugho please rebase this.

smaugho:
Done

dodgex:
Thanks! :)

Issue #2203, opened 2019-02-14
Bug in generated code with @RootContext


smaugho:
Hi, when using a method to inject the Context, and a class which is different than `Context` is used, the generated code is not correct. 

I saw this issue while testing #2202 , both annotations have the same bug then. (Note I tested with the normal "develop" branch, to be sure this was not related to the changes on #2202). 

**Annotated code**:
```java
@EBean
public class MainViewModel {

    @RootContext
    void someMethod(Activity anything) {
        //Some code
    }

}
```

**Expected generated code**:
```java
    private void init_() {
        if (context_ instanceof Activity) {
            Activity anything = ((Activity) context_);
            someMethod2(anything);
        } else {
            Log.w("MainViewModel_", (("Due to Context class "+ context_.getClass().getSimpleName())+", the @RootContext Activity won't be populated"));
        }
    }
```

**Actual generated code**:
```java
    private void init_() {
        {
            Activity anything = null;
            someMethod2(anything);
        }
        if (context_ instanceof Activity) {
            anything = ((Activity) context_);
        } else {
            Log.w("MainViewModel_", (("Due to Context class "+ context_.getClass().getSimpleName())+", the @RootContext Activity won't be populated"));
        }
    }
```

The code doesn't compile in general since the field in the "if (context ..." block `anything` doens't exist

dodgex:
Thank you! :)

PR #2204, opened 2019-02-15
Solves bug on generated code by @RootContext annotation when used on …

smaugho:
This PR solves a bug found and explained on #2203


AndroidAnnotations/androidannotations-core/androidannotations/src/main/java/org/androidannotations/internal/core/handler/RootContextHandler.java
```diff
@@ -82,13 +82,13 @@ public void assignValue(JBlock targetBlock, IJAssignmentTarget fieldRef, EBeanHo
 		} else {
 			AbstractJClass extendingContextClass = getEnvironment().getJClass(typeQualifiedName);
 
-			JConditional cond = getInvocationBlock(holder)._if(holder.getContextRef()._instanceof(extendingContextClass));
```
dodgex: could you please rebase on develop and check if the header `Copyright (C) 2016-2019 the AndroidAnnotations project` has been added to the file? if not please add it.

smaugho: @dodgex done for this PR

PR #2205, opened 2019-02-15
 Check for missing copyright line in CI

WonderCsabo:


WonderCsabo:
@dodgex i verified the [build fails](https://travis-ci.org/androidannotations/androidannotations/builds/493960984) with appropriate message with missing copryight.

dodgex:
nice one! :)

I think new files without copyright get cought by the maven header plugin right? does it mention the correct header (the one without excilys copyright)?

WonderCsabo:
Yes, as the [main header](https://github.com/androidannotations/androidannotations/blob/develop/AndroidAnnotations/pom.xml#L317) is the one without excilys.

Issue #1488, opened 2015-07-07
Provide an Activity scope for @EBean


himamis:
Description: https://github.com/excilys/androidannotations/issues/205#issue-4756568.
As of now, beans needed in the Activity scope are created in the activity, and the passed down with constructors and setters to other beans and fragments.

WonderCsabo:
Well, this was [rejected](https://github.com/excilys/androidannotations/issues/205#issuecomment-14354316) more than two years ago. Please note AA dependency injection is really limited. I am not sure we should try to improve it. Maybe we can try to provide better integration with Dagger or sg.

himamis:
RoboGuice has ContextSingleton and FragmentSingleton annotations working similar as described above. I will do a test if they work with AA.

WonderCsabo:
BTW, i am okay to accept contributions to implement this, for example this one is not that hard: https://github.com/excilys/androidannotations/issues/205#issue-4756568. But i do not have time to work on this solely myself.

kenvanhoeylandt:
I think it would be better to have a more generic approach through a "Context" scope:

A bean with the scope`Scope.Context` is a bean that is bound to 1 specific context, like an `Activity`, an `Application` or a `Service`. When this bean is linked within other beans/fragments of this context, it will re-use the original bean of that context rather than instantiate a new bean.

I wrote some sample code at https://github.com/excilys/androidannotations/issues/1663

gauravsak:
I think this issue is resolved as you can use @RootContext to access the activity in the bean. @WonderCsabo Please mark as closed. Thanks!

WonderCsabo:
@gauravsak no, it was not resolved. This issue is not about accessing the root Context from the bean, it is about how the bean instances are created. Currently we always create a new instance for non-singleton beans, but this issue wants only one instance per Activity.

smaugho:
Hi, I would like to implement this one, using the proposal on #2198 . Please @WonderCsabo , @dodgex  let me know if you agree with that solution.

smaugho:
@WonderCsabo, I'm implementing this feature. I've added the BeanHolder interface to support the Activity scope. But doing it I realized that this can be right now extended to the Fragments, so we could have also the `@EBean` scope `Fragment`, specially since now we've added the support for `@RootFragment` on #2202 . 

Seems less useful, but seems trivial to add. A use case would be if you have several beans in a same fragment, and you want from several of them inject another which is the same instance for all of them. 

[...]

PR #2206, opened 2019-02-15
Provide Activity scope for @EBean

smaugho:
This PR adds the scope Activity and scope Fragment to the `@EBean` annotation, solving #1488  

Please note, that this PR is for a revision, cause'  it depends on #2202 . Once I get #2202, I should make a rebase, and then this one would be ready to merge. 

But please, have a look at it, and provide feedback if possible. 

Thanks!


AndroidAnnotations/androidannotations-core/androidannotations-api/src/main/java/org/androidannotations/annotations/EBean.java
```diff
@@ -110,6 +110,22 @@
 		 */
 		Default, //
 
+		/**
+		 * A new instance of the bean is created the first time it is needed in an
```
WonderCsabo: A new instance of the bean is created the first time it is needed in an
Activity scope, it is then retained and the same instance is always returned
from within the same Activity. Different Activities hold different copies of
the bean.

AndroidAnnotations/androidannotations-core/androidannotations-api/src/main/java/org/androidannotations/annotations/EBean.java
```diff
@@ -110,6 +110,22 @@
...
+		 * activity scope, it is then retained and the same instance is always returned
+		 * from within the same activity. Different activities, hold different copies of
+		 * the bean
+		 */
+		Activity,
+
+		/**
+		 * A new instance of the bean is created the first time it is needed in aa
```
WonderCsabo: A new instance of the bean is created the first time it is needed in a
Fragment scope, it is then retained and the same instance is always returned
from within the same Fragment. Different Fragments hold different copies of
the bean.

AndroidAnnotations/androidannotations-core/androidannotations/src/main/java/org/androidannotations/holder/EBeanHolder.java
```diff
@@ -111,23 +116,30 @@ public void invokeInitInConstructors() {
...
 
 		JBlock factoryMethodBody = factoryMethod.body();
 
-		/*
-		 * Singletons are bound to the application context
-		 */
-		if (hasSingletonScope) {
+		switch (scope) {
+		case Default:
+		case Fragment:
```
WonderCsabo: How come `Default` and `Fragment` is in the same branch? Maybe i do not understand something?

smaugho: Because they share common code, but note that below it is called `createOverloadedFactoryMethod`, and it is there where a difference between both scopes is set. The fragment scope code is:

```java
    public static TestModel_ getInstance_(Context context, Object rootFragment) {
        if (rootFragment instanceof BeanHolder) {
            BeanHolder beanHolder = ((BeanHolder) rootFragment);
            TestModel_ bean = beanHolder.getBean(TestModel_.class);
            if (bean == null) {
                bean = new TestModel_(context, rootFragment);
                beanHolder.putBean(TestModel_.class, bean);
            }
            return bean;
        }
        return new TestModel_(context, rootFragment);
    }
```

The Default scope only has:

```java
    public static TestModel_ getInstance_(Context context, Object rootFragment) {
        return new TestModel_(context, rootFragment);
    }
```

smaugho:
I added all the tests, did also some organization in existing tests, so please have a look to it. Note that the "Cyclic Tests" are using a Fragment now done only for that end. Also added "Scope Tests" and moved some existing tests related to the "Singleton" scope under that one.

smaugho:
Ah Important, this PR depends on #2202, it is basically built over that one (so if you want to see the changes introduced only by this PR, compare it not with "develop", but with #2202 branch. I guess that for merging it, we should firstly do #2202, and afterward this one.

WonderCsabo:
@smaugho this should be rebased, right?

smaugho:
It should be rebased. But I would wait to merging #2202 , since it depends on it as well, so that I have to rebase only once.

dodgex:
#2202 has been merged. could you please rebase and ensure the conflicts are gone? thanks :)

dodgex:
@WonderCsabo you still have requested changes here. are they resolved but you missed to update your review or are the changes still pending?

WonderCsabo:
@dodgex i think they were resolved, waiting for rebase now. :)

dodgex:
@WonderCsabo okay, after rebase approve the PR and then I'll review.

smaugho:
Gonna rebase now 👍

PR #2207, opened 2019-02-16
Update dependencies

WonderCsabo:


WonderCsabo:
Note: Robolectric update is not feasible currently, as Robolectric 4.2 requires AndroidX testing. We can upgrade to that if we ever manage to migrate to Gradle (which could also solve other limitations).

dodgex:
Thanks

PR #2215, opened 2019-02-18
Add useful methods to PluginClassHolder

smaugho:
Add methods to PluginClassHolder to facilitate the work with Plugins. Also included a little check in EServiceHoldr to ensure `getStartLifecycleAfterSuperBlock doesn't return NULL (similar to how it works in other places).

### DIFF
--- a/AndroidAnnotations/androidannotations-core/androidannotations/src/main/java/org/androidannotations/AndroidAnnotationsEnvironment.java
+++ b/AndroidAnnotations/androidannotations-core/androidannotations/src/main/java/org/androidannotations/AndroidAnnotationsEnvironment.java
@@ -26,9 +26,9 @@ import javax.lang.model.element.Element;
 import org.androidannotations.handler.AnnotationHandler;
 import org.androidannotations.handler.GeneratingAnnotationHandler;
 import org.androidannotations.helper.AndroidManifest;
+import org.androidannotations.helper.ClassesHolder;
 import org.androidannotations.holder.GeneratedClassHolder;
 import org.androidannotations.internal.model.AnnotationElements;
-import org.androidannotations.internal.process.ProcessHolder;
 import org.androidannotations.plugin.AndroidAnnotationsPlugin;
 import org.androidannotations.rclass.IRClass;
 
@@ -76,7 +76,7 @@ public interface AndroidAnnotationsEnvironment {
 
 	GeneratedClassHolder getGeneratedClassHolder(Element element);
 
-	ProcessHolder.Classes getClasses();
+	ClassesHolder.Classes getClasses();
 
 	List<Class<? extends Annotation>> getGeneratingAnnotations();
 
--- a/AndroidAnnotations/androidannotations-core/androidannotations/src/main/java/org/androidannotations/helper/ClassesHolder.java
+++ b/AndroidAnnotations/androidannotations-core/androidannotations/src/main/java/org/androidannotations/helper/ClassesHolder.java
@@ -1,0 +1,229 @@
+/**
+ * Copyright (C) 2016-2019 the AndroidAnnotations project
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may not
+ * use this file except in compliance with the License. You may obtain a copy of
+ * the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed To in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+ * License for the specific language governing permissions and limitations under
+ * the License.
+ */
+package org.androidannotations.helper;
+
+import java.io.FileInputStream;
+import java.io.InputStream;
+import java.io.Serializable;
+import java.sql.SQLException;
+import java.util.Arrays;
+import java.util.Collections;
+import java.util.HashMap;
+import java.util.HashSet;
+import java.util.List;
+import java.util.Map;
+
+import javax.annotation.processing.ProcessingEnvironment;
+
+import com.helger.jcodemodel.AbstractJClass;
+import com.helger.jcodemodel.JClassAlreadyExistsException;
+import com.helger.jcodemodel.JCodeModel;
+import com.helger.jcodemodel.JDefinedClass;
+
+public class ClassesHolder {
+
+	// CHECKSTYLE:OFF
+	public class Classes {
+
+		/*
+		 * Java
+		 */
+		public final AbstractJClass RUNTIME_EXCEPTION = refClass(RuntimeException.class);
+		public final AbstractJClass EXCEPTION = refClass(Exception.class);
+		public final AbstractJClass THROWABLE = refClass(Throwable.class);
+		public final AbstractJClass CHAR_SEQUENCE = refClass(CharSequence.class);
+		public final AbstractJClass CLASS_CAST_EXCEPTION = refClass(ClassCastException.class);
+		public final AbstractJClass SERIALIZABLE = refClass(Serializable.class);
+		public final AbstractJClass STRING = refClass(String.class);
+		public final AbstractJClass STRING_BUILDER = refClass(StringBuilder.class);
+		public final AbstractJClass SYSTEM = refClass(System.class);
+		public final AbstractJClass INPUT_STREAM = refClass(InputStream.class);
+		public final AbstractJClass FILE_INPUT_STREAM = refClass(FileInputStream.class);
+		public final AbstractJClass SQL_EXCEPTION = refClass(SQLException.class);
+		public final AbstractJClass COLLECTIONS = refClass(Collections.class);
+		public final AbstractJClass THREAD = refClass(Thread.class);
+		public final AbstractJClass HASH_MAP = refClass(HashMap.class);
+		public final AbstractJClass LIST = refClass(List.class);
+		public final AbstractJClass OBJECT = refClass(Object.class);
+		public final AbstractJClass ARRAYS = refClass(Arrays.class);
+		public final AbstractJClass HASH_SET = refClass(HashSet.class);
+		public final AbstractJClass CLASS = refClass(Class.class);
+		public final AbstractJClass MAP = refClass(Map.class);
+
+		/*
+		 * Android
+		 */
+		public final AbstractJClass LOG = refClass(CanonicalNameConstants.LOG);
+		public final AbstractJClass BUNDLE = refClass(CanonicalNameConstants.BUNDLE);
+		public final AbstractJClass ACTIVITY = refClass(CanonicalNameConstants.ACTIVITY);
+		public final AbstractJClass EDITABLE = refClass(CanonicalNameConstants.EDITABLE);
+		public final AbstractJClass TEXT_WATCHER = refClass(CanonicalNameConstants.TEXT_WATCHER);
+		public final AbstractJClass SEEKBAR = refClass(CanonicalNameConstants.SEEKBAR);
+		public final AbstractJClass ON_SEEKBAR_CHANGE_LISTENER = refClass(CanonicalNameConstants.ON_SEEKBAR_CHANGE_LISTENER);
+		public final AbstractJClass TEXT_VIEW = refClass(CanonicalNameConstants.TEXT_VIEW);
+		public final AbstractJClass TEXT_VIEW_ON_EDITOR_ACTION_LISTENER = refClass(CanonicalNameConstants.TEXT_VIEW_ON_EDITOR_ACTION_LISTENER);
+		public final AbstractJClass COMPOUND_BUTTON = refClass(CanonicalNameConstants.COMPOUND_BUTTON);
+		public final AbstractJClass COMPOUND_BUTTON_ON_CHECKED_CHANGE_LISTENER = refClass(CanonicalNameConstants.COMPOUND_BUTTON_ON_CHECKED_CHANGE_LISTENER);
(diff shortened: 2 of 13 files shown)


AndroidAnnotations/androidannotations-core/androidannotations/src/main/java/org/androidannotations/plugin/PluginClassHolder.java
```diff
@@ -59,4 +62,13 @@ protected AbstractJClass getJClass(Class<?> clazz) {
 	protected JCodeModel getCodeModel() {
 		return environment().getCodeModel();
 	}
+
+	protected ProcessingEnvironment getProcessingEnvironment() {
+		return environment().getProcessingEnvironment();
+	}
+
+	protected ProcessHolder.Classes getClasses() {
```
WonderCsabo: `ProcessHolder` is an internal class, which is not meant to be used by plugins...

AndroidAnnotations/androidannotations-core/androidannotations/src/main/java/org/androidannotations/plugin/PluginClassHolder.java
```diff
@@ -59,4 +62,13 @@ protected AbstractJClass getJClass(Class<?> clazz) {
 	protected JCodeModel getCodeModel() {
 		return environment().getCodeModel();
 	}
+
+	protected ProcessingEnvironment getProcessingEnvironment() {
```
WonderCsabo: Why would plugins need this? The plugin system is created in a way that plugins should only be called by handlers.

smaugho: @WonderCsabo , any plugin with "a little of complexity" could need to do stuff with the ProcessingEnvironment. Like getting the ElementUtils or TypeUtils to make any enhanced processing over elements. This is just a convenient method, the plugin always can call environment().getProcessingEvironment().. but I don't see the point in not facilitating this (in the same way that it was done with the getCodeModel()..

AndroidAnnotations/androidannotations-core/androidannotations/src/main/java/org/androidannotations/plugin/PluginClassHolder.java
smaugho: @WonderCsabo , though this list of classes is really a convenient one. Plugins need to create AbstractJClass for things like the Context, Object, or several one which are already done on AA, stored in the Classes lists. 

The plugin devs can of course call `getJClass()` and pass the class name.. just thought that if this is in fact available in AA, it was good idea to make it easier to use for plugin devs.. But I do understand about the private status of the list. 

Even though, plugin devs can always access to that list through the environment, so there is not really a "big point"  in hiding it, it is available anyway.

WonderCsabo: But in that case we should move it out from the `internal` package.

AndroidAnnotations/androidannotations-core/androidannotations/src/main/java/org/androidannotations/plugin/PluginClassHolder.java
WonderCsabo: You are right, sorry

AndroidAnnotations/androidannotations-core/androidannotations/src/main/java/org/androidannotations/plugin/PluginClassHolder.java
smaugho: Let me know where would be good to place it. Also if you agree to "facilitate it".. I understand that AA doesn't need to have a compromise to facilitate a list of `AbstractJClass` objects for the plugins.. so I do agree that this is not needed. It simply will be an extra resource for the Plugins Devs, so that they don't have to create their own "CLASSES".. anyway, sometimes is required to do it, even if we expose those Classes, I have a lot of them already defined for the Architecture Components:

```java
public class ArchiCanonicalNameConstants {

    public static final String VIEW_MODEL_PROVIDERS = "android.arch.lifecycle.ViewModelProviders";

    public static final String VIEW_MODEL = "android.arch.lifecycle.ViewModel";

    public static final String ANDROID_VIEW_MODEL = "android.arch.lifecycle.AndroidViewModel";

    public static final String OBSERVER = "android.arch.lifecycle.Observer";

    public static final String LIFECYCLE_OWNER = "android.arch.lifecycle.LifecycleOwner";

    public static final String LIFECYCLE_OBSERVER = "android.arch.lifecycle.LifecycleObserver";

    public static final String LIVE_DATA = "android.arch.lifecycle.LiveData";

    public static final String MUTABLE_LIVE_DATA = "android.arch.lifecycle.MutableLiveData";

    public static final String SINGLE_LIVE_EVENT = "pl.com.dspot.archiannotations.api.SingleLiveEvent";

}
```

WonderCsabo: I am AFK, feel free to suggest location, i can check it later.

smaugho: BTW, if this is not "exposed", the correct way to do it is not to facilitate it in the `AndroidAnnotationsEnvironment` then right?... cause anyway, it is exposed there.. as long as it is there, devs will be using them (being internal or not).

smaugho: > 
> 
> I am AFK, feel free to suggest location, i can check it later.

ok ok

smaugho: @WonderCsabo , lamentably moving this class out of the ProcessHolder is not that simple. Basically, the class is not static, it uses the method "refClass" from the ProcessHolder which is basically the base method for the `getJClass()` used through whole the system. 

So basically we'll have to move all the ProcessHolder out (and maybe this was your idea, though I thought it was only to move Classes). If you are ok with moving all the ProcessHolder, a good place for it would be to the pacakge `org.androidannotations.process`

Moving only "Classes" implies to make some sort of "ClassesHolder". The idea is not that bad, since the ProcessHolder stores as well every GeneratedClassHolder in the system, which has nothing to do with the `Classes`, `getJClass` or similar methods. If you prefer adding this Class, which should be present in the AndroidAnnotationEnvironment then, let me know.. the proposed package for it would be better something like `org.androidannotations.classes`.

WonderCsabo: Do plugins need `ProcessHolder`, or only `Classes`?

smaugho: Only classes.

smaugho: In fact the other methods in the ProcessHolder, which are:

```java
public void put(Element element, GeneratedClassHolder generatedClassHolder);

public GeneratedClassHolder getGeneratedClassHolder(Element element);
```

These are used only mainly internally (`ModelProcessor`, `InternalAnnotationProcessor). 

Though, the PrefHandler uses it as well, which I think that will break in fact the Incremental compilation if used...

WonderCsabo: We should only move out Classes from the `internal` package, then.

smaugho: I've made another commit on this PR. I moved the Classes toward a ClassesHolder which is on the "helper" package.. I'm only not sure about the package, but this "Holder" has nothing to do with the Holders which are in the "holder" package..

AndroidAnnotations/androidannotations-core/androidannotations/src/main/java/org/androidannotations/helper/ClassesHolder.java
```diff
@@ -0,0 +1,229 @@
...
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed To in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+ * License for the specific language governing permissions and limitations under
+ * the License.
+ */
+package org.androidannotations.helper;
```
WonderCsabo: Nice, i also wanted to suggest this package!

dodgex:
thanks

AndroidAnnotations/androidannotations-core/androidannotations/src/main/java/org/androidannotations/AndroidAnnotationsEnvironment.java
```diff
@@ -76,7 +76,7 @@
 
 	GeneratedClassHolder getGeneratedClassHolder(Element element);
 
-	ProcessHolder.Classes getClasses();
+	ClassesHolder.Classes getClasses();
```
dodgex: Damn... right after mergin i came up with a question.... don't the changes to `AndroidAnnotationsEnvironment` break existing plugins if the use this method?!

dodgex: @WonderCsabo @smaugho I hope I'm wrong with this thougt. :/

WonderCsabo: Yes, it will. However i think we can allow breaking changes for plugins, but not allow breaking changes for users. WDYT?

WonderCsabo: But we should raise the minor version, and document the changes.

dodgex: @WonderCsabo it depends on how active the plugin devs are and if they update the plugins to be compatible once again. out-of-date updates mean breaking change for the users of those pluigns.

WonderCsabo: Hmm, you are correct. I am afraid we cannot have this change without moving to 5.0. So we have to revert moving the class.

WonderCsabo: @smaugho can you do that?

dodgex: we could revert the PR merge and @smaugho can open a new PR with the other changes.

smaugho: @WonderCsabo, @dodgex I guess it should be done 😓 ... Could we open a new branch for version 5 realease and keep these changes there?

WonderCsabo: Yes, please!

WonderCsabo: And can you please open another PR which reverts moving `Classes`?

smaugho: Yes @WonderCsabo , will try to do it today or tomorrow. Please if possible create this branch for version 5, in order to submit this one there (so I guess, I have to make a PR to develop which reverts the changes, and another one to the branch for version 5, which remakes it.

WonderCsabo: For now just move back Classes. For version 5.0, we will just revert that PR.

smaugho: Hi @WonderCsabo , @dodgex .. sorry that still didn't make the change... had quite a hectic week. I see you're considering let a breaking change on the release, I'll wait for your decision on that. I don't think that is something uncommon, I'm used to see that in many libs around, where a specific version works with some specific version of other lib (or with a version range of the other lib). 

Also in more codes, the "Classes" are used like: `getClasses().SOME_OBJECT`, and that will not be broken, how this will break it is if the programmer assigns the classes object, like `Classes classes = pluginHolder.getClassess()` ...

WonderCsabo: it will break if the programmer imported the `Classes` class. Please revert that move.

dodgex:
Unfortunately we have to revert this PR due to breaking changes: Closing PR due to breaking changes. See https://github.com/androidannotations/androidannotations/pull/2230

PR #2217, opened 2019-02-20
Add methods to register Enhanced Components Annotation for Generating…

smaugho:
This adds the method for the plugins to be able to add elements to the Annotations lists.

### DIFF
--- a/AndroidAnnotations/androidannotations-core/androidannotations/src/main/java/org/androidannotations/helper/ModelConstants.java
+++ b/AndroidAnnotations/androidannotations-core/androidannotations/src/main/java/org/androidannotations/helper/ModelConstants.java
@@ -19,6 +19,8 @@ package org.androidannotations.helper;
 import static java.util.Arrays.asList;
 
 import java.lang.annotation.Annotation;
+import java.util.ArrayList;
+import java.util.Collections;
 import java.util.List;
 
 import javax.lang.model.SourceVersion;
@@ -43,10 +45,10 @@ public abstract class ModelConstants {
 	private static String generationSuffix = "_";
 	private static String classSuffix;
 
-	public static final List<Class<? extends Annotation>> VALID_ENHANCED_VIEW_SUPPORT_ANNOTATIONS = asList(EActivity.class, EViewGroup.class, EView.class, EBean.class, EFragment.class);
+	private static List<Class<? extends Annotation>> validEnhancedViewSupportAnnotations = new ArrayList<>(asList(EActivity.class, EViewGroup.class, EView.class, EBean.class, EFragment.class));
 
-	public static final List<Class<? extends Annotation>> VALID_ENHANCED_COMPONENT_ANNOTATIONS = asList(EApplication.class, EActivity.class, EViewGroup.class, EView.class, EBean.class, EService.class,
-			EIntentService.class, EReceiver.class, EProvider.class, EFragment.class);
+	private static List<Class<? extends Annotation>> validEnhancedComponentAnnotations = new ArrayList<>(
+			asList(EApplication.class, EActivity.class, EViewGroup.class, EView.class, EBean.class, EService.class, EIntentService.class, EReceiver.class, EProvider.class, EFragment.class));
 
 	private ModelConstants() {
 	}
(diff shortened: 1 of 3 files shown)


AndroidAnnotations/androidannotations-core/androidannotations/src/main/java/org/androidannotations/helper/ModelConstants.java
```diff
@@ -70,4 +71,21 @@ public static String classSuffix() {
...
+
+	public static List<Class<? extends Annotation>> validEnhancedViewSupportAnnotations() {
+		return validEnhancedViewSupportAnnotations;
+	}
+
+	public static List<Class<? extends Annotation>> validEnhancedComponentAnnotations() {
+		return validEnhancedComponentAnnotations;
+	}
+
+	public static void addEnhancedViewSupportAnnotation(Class<? extends Annotation> annotation) {
```
WonderCsabo: I do not see where these methods are called.

WonderCsabo: Okay, i understand now. I suggested a different thing: to just collect the enhanced component annotations from the handlers after we registered all the plugins. I would not allow to modify this later.

smaugho: @WonderCsabo there is an issue with that one though. We don't know to which of both lists the component should be assigned. In fact, this is determined by the "ComponentHolder" which is created with the method:

```
        public EBeanHolder createGeneratedClassHolder(AndroidAnnotationsEnvironment environment, TypeElement annotatedComponent) throws Exception {
		return new EBeanHolder(environment, annotatedComponent);
	}
```

in the `GeneratingAnnotationHandler`

So, to do it like that, will have to add another method to the GeneratingAnnotationsHandlers, something like a boolean where each handler indicates if it is a Generating Component Handler, or if it has ViewSupport (well, all which have ViewSupport are also Enhanced Component Handlers)

mmm.. or we could always add them to the Enhanced Component Handlers list, but we'll don't know if to add them to the list with ViewSupport. 

Right now I need only the first though.. but I think it would not be generic enough. At least that we create a mechanism for it.

WonderCsabo: Let's discuss this in Gitter.

AndroidAnnotations/androidannotations-core/androidannotations/src/main/java/org/androidannotations/helper/ModelConstants.java
```diff
@@ -70,4 +71,21 @@ public static String classSuffix() {
 	public static String generationSuffix() {
 		return generationSuffix;
 	}
+
+	public static List<Class<? extends Annotation>> validEnhancedViewSupportAnnotations() {
+		return validEnhancedViewSupportAnnotations;
```
dodgex: shouldn't we return a new or immutable copy of our internal list? just returning the list allows plugins to fully modify the list.

@WonderCsabo wdyt?

AndroidAnnotations/androidannotations-core/androidannotations/src/main/java/org/androidannotations/helper/ModelConstants.java
```diff
@@ -70,4 +71,21 @@ public static String classSuffix() {
...
+	}
+
+	public static List<Class<? extends Annotation>> validEnhancedComponentAnnotations() {
+		return validEnhancedComponentAnnotations;
```
dodgex: same as above

AndroidAnnotations/androidannotations-core/androidannotations/src/main/java/org/androidannotations/helper/ModelConstants.java
WonderCsabo: We should. However they still can modify it through the setters...

dodgex: There are no setters. only "adders". and ADDING stuff is ok, but with the original list you could also remove stuff.

WonderCsabo: True.

WonderCsabo: But also if you adds invalid stuff, that is not a good thing. :)

smaugho:
Ruturning Immutable copy now

AndroidAnnotations/androidannotations-core/androidannotations/src/main/java/org/androidannotations/helper/ModelConstants.java
WonderCsabo: Let's keep it this way for now.

WonderCsabo:
@dodgex this PR also breaks plugins...

dodgex:
@WonderCsabo as mentioned in #2218 If we decide to break plugins with a minor release I'm ok with this changes.

dodgex:
Closing PR due to breaking changes. See https://github.com/androidannotations/androidannotations/pull/2230#issuecomment-497313578.

PR #2220, opened 2019-02-24
Fix bug duplicate create

smaugho:
This PR fixes a bug which happens (with plugins). Basically the method "setOnCreate"  could be called more than once. A simple null-check has been added which fixes the bug.


WonderCsabo:
https://github.com/androidannotations/androidannotations/pull/2218 must be merged before this.

dodgex:
I'll review this as soon as we have decided what to do with the breaking change in #2218 and merged that PR.

It is hard for me to see what this PR changes in beside that what should happen with #2218.

dodgex:
Closing PR due to breaking changes. See https://github.com/androidannotations/androidannotations/pull/2230#issuecomment-497313578.

PR #2231, opened 2019-05-30
 Support Android Gradle Plugin 3.5.0-beta03 and 3.6.0-alpha01

dodgex:
Google updated the generated code folder for Java Builds and with this PR we support those new folders for Manifest detection. :)

Fixes https://github.com/androidannotations/androidannotations/issues/2229.

### DIFF
--- a/AndroidAnnotations/androidannotations-core/androidannotations/src/main/java/org/androidannotations/internal/helper/AndroidManifestFinder.java
+++ b/AndroidAnnotations/androidannotations-core/androidannotations/src/main/java/org/androidannotations/internal/helper/AndroidManifestFinder.java
@@ -126,7 +126,7 @@ public class AndroidManifestFinder {
 		return findManifestInKnownPathsStartingFromGenFolder(holder.sourcesGenerationFolder.getAbsolutePath());
 	}
 
-	File findManifestInKnownPathsStartingFromGenFolder(String sourcesGenerationFolder) throws FileNotFoundException {
+	File findManifestInKnownPathsStartingFromGenFolder(String sourcesGenerationFolder) {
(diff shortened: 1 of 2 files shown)


AndroidAnnotations/androidannotations-core/androidannotations/src/main/java/org/androidannotations/internal/helper/AndroidManifestFinder.java
```diff
@@ -193,22 +221,26 @@ boolean applies() {
...
-			String path = matcher.group(1);
-			String mode = matcher.group(2);
-			String gradleVariant = matcher.group(3);
-			String variantPart = gradleVariant.substring(1);
+			String path = getPath();
+			String gradleVariant = getGradleVariant();
 
 			List<String> possibleLocations = new ArrayList<>();
-			findPossibleLocationsV32(path, variantPart, possibleLocations);
+			findPossibleLocationsV32(path, gradleVariant, possibleLocations);
```
lucalves: For which reason the change in the name of the variable `variantPart`?

dodgex: Before this change we had `gradleVariant ` and `variantPart`. `variantPart` was a substring of `gradleVariant` and as this substring part is gone i decided to use `gradleVariant ` as the remaining variable.

psmorandi:
Hello there, could anyone review this?? AA is failing for the new android gradle plugin released (3.5.0).

WonderCsabo:
I'll try to review this on Monday.

WonderCsabo:
The fix is now available in the 4.7.0-SNAPSHOT build.

PR #2232, opened 2019-06-03
Reformat gradle sample

XieEDeHeiShou:
+ Add `protected` modifier to members;
+ Reformat layout files;
+ Show un-cancelable dialog after api 21 when click `Click Me!`;


WonderCsabo:
Thank you for this contribution. However i think it adds lot of unnecessary things. Respectfully i would reject this for now.

PR #2243, opened 2019-08-30
Fix Travis build

WonderCsabo:
https://blog.travis-ci.com/2019-04-15-xenial-default-build-environment

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -31,7 +31,7 @@ before_script:
 script:
   - set -e
   - cd $TRAVIS_BUILD_DIR/AndroidAnnotations
-  - ./mvnw install -P travis
+  - ./mvnw install -P travis --quiet
   - cd $TRAVIS_BUILD_DIR/examples/maven
   - ./mvnw install
   - cd $TRAVIS_BUILD_DIR/examples/gradle


WonderCsabo:
Travis support suggested the above, but it did not help unfortunately. :(

dodgex:
@WonderCsabo I added [`--quiet`](http://maven.apache.org/ref/3.6.1/maven-embedder/cli.html#quiet) to the task as the build fails with `The job exceeded the maximum log length, and has been terminated.` and quiet only shows errors.

in the "early" lines (line [1741](https://travis-ci.org/androidannotations/androidannotations/builds/578764925#L1741)) it mentiones that it could not download the cache. I hope that the quiet allows to build a new cache and then have it working again without quiet. :)

dodgex:
okay, it looks like if we run our develop once with `--quiet` on our [maven install step](https://github.com/androidannotations/androidannotations/blob/develop/.travis.yml#L34) the .m2 cache gets build for xenial and should then work for PRs too.

right now this PR has no changes remaining.

dodgex:
Yeah, build is green with no changes compared to develop.

---

I squashed all my commits and removed the `dist: trusty` part you added. now this PR only adds `--quiet`. For now I'd suggest to merge this, let develop build and remove the --quiet again by reverting this PR. Or we keep the quiet flag and only have test output and possible errors... here is a build with --quiet enabled: https://travis-ci.org/androidannotations/androidannotations/builds/579284541

